PR #2, opened 2022-07-06
Initial xr

byrcolin:
This is the result of filtering and moving gallifrey XR Gem, OpenXRVk Gem and the OpenXRTest project and their histories into o3de-extras


amzn-phist (approved):
Considering all this has been reviewed before, it looks like it's brought over correctly and has preserved history.  LGTM.

amzn-changml (approved):
Looks reasonable. The duplicate commits are a minor issue, but this shouldn't affect the aggregate history once new gems are merged here

moraaar:
Shouldn't this be merged into development branch instead?

moraaar (commented):
I took a look at the folder structure rather than the content of the files, which I assume it's the same as the source repo, and it looks good.

I'll hold the approve until verifying main is the right target branch and not development. Plus the following question: It was mentioned previously that projects won't be included, this PR includes a project, has that been reconsidered?

byrcolin:
I think main is fine as we do not have AR yet on this repo. The should we include projects question, its not a question of should we include projects, rather should we include this project... Well I still think we need something like this project in there. It does not work currently and as long as we do not add the project to the canonical list of projects then it will not be discovered by the program manager, but will still allow us to work on it... but we could still work on it in gallifrey, however gallifrey is private and we want the communities help, so I have a hard time saying it should not go in here.

moraaar (approved):

PR #16, opened 2022-08-17
OpenXRVk: Use PAL to initialize xr loader on Android

moraaar:


Gems/OpenXRVk/Code/Source/OpenXRVkInstance.cpp
```diff
@@ -88,26 +85,11 @@ namespace OpenXRVk
 
     AZ::RHI::ResultCode Instance::InitInstanceInternal(AZ::RHI::ValidationMode validationMode)
     {
-#ifdef AZ_PLATFORM_ANDROID
-        // Initialize the loader for Android platform
+#if OPENXRVK_NEEDS_INITIALIZE_XR_LOADER
```
amzn-phist, line 88: For PAL defines like this it's common to use the word `TRAIT` in the name (e.g. `OPENXRVK_TRAIT_NEEDS_INITIALIZE_XR_LOADER`).

Gems/OpenXRVk/Code/Source/Platform/Android/OpenXRInitializeLoader_Android.cpp
```diff
@@ -0,0 +1,45 @@
...
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AzCore/Android/AndroidEnv.h>
+
+#include <OpenXRVk/OpenXRVkUtils.h>
+#include <OpenXRInitializeLoader_Android.h>
+
+namespace OpenXRVk
```
amzn-phist, line 14: Consider naming this `namespace OpenXRVk::Platform`, it gives a good indication at call site that there's platform-specific implementation behind it.

amzn-phist (approved):

Gems/OpenXRVk/Code/Source/OpenXRVkInstance.cpp
moraaar: Excellent point. I'll change it.

Gems/OpenXRVk/Code/Source/OpenXRVkInstance.cpp
```diff
@@ -88,26 +85,11 @@ namespace OpenXRVk
 
     AZ::RHI::ResultCode Instance::InitInstanceInternal(AZ::RHI::ValidationMode validationMode)
     {
-#ifdef AZ_PLATFORM_ANDROID
-        // Initialize the loader for Android platform
+#if OPENXRVK_TRAIT_NEEDS_INITIALIZE_XR_LOADER
```
moudgils, line 88: Do we need this macro if we have a common implementation of OpenXRInitializeLoader that returns true. Makes the code a little messy.

moraaar: I think it's a larger change doing it that way, plus adding a cpp and a header just to do an empty function seems like a waste, this way it's just a one line macro.

moudgils (approved):

PR #17, opened 2022-08-25
Add automated testing

byrcolin:
This file instructs the AR register objects and enable gems for the AR of AutomatedTesting project


byrcolin:
need to fix this

PR #22, opened 2022-08-30
XR/OpenXrVk changes related to the new RPI sample added to AtomSampleViewer

moudgils:
Extended XR apis to support the new sample added to AtomSampleViewer. This sample supports new multi-view VR pipelines and calls into XR/OpenXrVk gem to support and test controller data + head set tracking data. 
- Extended api to get state related to following button presses - A, B, X, Y, Squeeze, Trigger
- Extended api to support Joystick control
- Cleaned up some api related to extracting Pose dat related to head set and the controllers
- Added support for reverse depth when creating stereoscopic projection matrix
- Misc cleanup
 - Please see PR description for other PRs linked to this one in order to get the full context of the work here

Linked PRs are as follows
Core engine PR - https://github.com/o3de/o3de/pull/11670
ASV PR - https://github.com/aws-lumberyard-dev/o3de-atom-sampleviewer/pull/7
image

Signed-off-by: moudgils <47460854+moudgils@users.noreply.github.com>

### DIFF
--- a/Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkDevice.h
+++ b/Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkDevice.h
@@ -30,9 +30,9 @@ namespace OpenXRVk
         // Create the xr specific native device object and populate the XRDeviceDescriptor with it.
         AZ::RHI::ResultCode InitDeviceInternal(AZ::RHI::XRDeviceDescriptor* instanceDescriptor) override;
         //! Get the Fov data  of the view specified by view index
-        AZ::RPI::FovData GetViewFov(AZ::u32 viewIndex) const override;
+        AZ::RHI::ResultCode GetViewFov(const AZ::u32 viewIndex, AZ::RPI::FovData& outFovData) const override;
         //! Get the Pose data  of the view specified by view index
-        AZ::RPI::PoseData GetViewPose(AZ::u32 viewIndex) const override;
+        AZ::RHI::ResultCode GetViewPose(const AZ::u32 viewIndex, AZ::RPI::PoseData& outPoseData) const override;
         //////////////////////////////////////////////////////////////////////////
 
         //! Returns true if rendering data is valid for the current frame.
@@ -61,7 +61,7 @@ namespace OpenXRVk
         //! Release the oldest swapchain image and inform the drivers that the frame is ending 
         void EndFrameInternal(XR::Ptr<XR::SwapChain>) override;
         //! Locate views, acquire swapchain image and synchronize gpu with cpu
-        bool AcquireSwapChainImageInternal(AZ::u32 viewIndex, XR::SwapChain* baseSwapChain) override;
+        bool AcquireSwapChainImageInternal(const AZ::u32 viewIndex, XR::SwapChain* baseSwapChain) override;
         //////////////////////////////////////////////////////////////////////////
 
         VkDevice m_xrVkDevice = VK_NULL_HANDLE;
--- a/Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkDevice.h
+++ b/Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkDevice.h
@@ -30,9 +30,9 @@ namespace OpenXRVk
         // Create the xr specific native device object and populate the XRDeviceDescriptor with it.
         AZ::RHI::ResultCode InitDeviceInternal(AZ::RHI::XRDeviceDescriptor* instanceDescriptor) override;
         //! Get the Fov data  of the view specified by view index
-        AZ::RHI::ResultCode GetViewFov(const AZ::u32 viewIndex, AZ::RPI::FovData& outFovData) const override;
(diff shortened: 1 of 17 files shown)


Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkDevice.h
```diff
@@ -30,9 +30,9 @@ namespace OpenXRVk
         // Create the xr specific native device object and populate the XRDeviceDescriptor with it.
         AZ::RHI::ResultCode InitDeviceInternal(AZ::RHI::XRDeviceDescriptor* instanceDescriptor) override;
         //! Get the Fov data  of the view specified by view index
-        AZ::RPI::FovData GetViewFov(AZ::u32 viewIndex) const override;
+        AZ::RHI::ResultCode GetViewFov(const AZ::u32 viewIndex, AZ::RPI::FovData& outFovData) const override;
```
jeremyong-az, line 33: Suggest leaving the first argument as `AZ::u32 viewIndex` since the fact that `viewIndex` is unchanged internally is essentially an _implementation detail_ of `GetViewFov` and we wouldn't want the API of `GetViewFov` to change if `viewIndex` changed internally (not that it would). Generally, I think arguments passed by value should never be marked `const` for this reason (exposes/leaks an implementation detail). This comment probably applies to the other APIs introduced in this PR

Gems/OpenXRVk/Code/Source/OpenXRVkInput.cpp
```diff
@@ -53,37 +56,88 @@ namespace OpenXRVk
...
+            aznumeric_cast<uint32_t>(m_handSubactionPath.size()), m_handSubactionPath.data());
+        CreateAction(m_bButtonAction.m_actionHandle, XR_ACTION_TYPE_FLOAT_INPUT, "b_button", "B Button Object",
+            aznumeric_cast<uint32_t>(m_handSubactionPath.size()), m_handSubactionPath.data());
+        CreateAction(m_joyStickXAction.m_actionHandle, XR_ACTION_TYPE_FLOAT_INPUT, "joystick_x", "JoyStick X Object",
+            aznumeric_cast<uint32_t>(m_handSubactionPath.size()), m_handSubactionPath.data());
+        CreateAction(m_joyStickYAction.m_actionHandle, XR_ACTION_TYPE_FLOAT_INPUT, "joystick_y", "JoyStick Y Object",
+            aznumeric_cast<uint32_t>(m_handSubactionPath.size()), m_handSubactionPath.data());
+
         AZStd::array<XrPath, static_cast<uint32_t>(XR::Side::Count)> squeezeValuePath;
+        AZStd::array<XrPath, static_cast<uint32_t>(XR::Side::Count)> triggerValuePath;
```
jeremyong-az, line 73: Nit: Maybe we should have a `constexpr static uint32_t` value for this count to reduce the noise here a little (lot of static casts in this file in general which hurts readability a little haha)

jeremyong-az (approved):

Gems/XR/Code/Source/XRSystem.cpp
```diff
@@ -186,23 +186,50 @@ namespace XR
...
+        return AZ::RHI::ResultCode::Fail;
+    }
+
+    AZ::RHI::ResultCode System::GetControllerStagePose(AZ::u32 handIndex, AZ::RPI::PoseData& outPoseData) const
+    {
+        if (m_session->IsSessionRunning())
+        {
+            return m_session->GetControllerStagePose(handIndex, outPoseData);
+        }
+        return AZ::RHI::ResultCode::Fail;
```
amzn-phist, line 214: Wondering if these similar functions should return something other than `AZ::RHI::ResultCode::Fail` when the session is not running, like `AZ::RHI::ResultCode::NotReady` or `AZ::RHI::ResultCode::InvalidOperation`.

amzn-phist (approved):

Gems/XR/Code/Source/XRUtils.cpp
```diff
@@ -32,16 +32,32 @@ namespace XR
...
+            }
+            else
+            {
+                result.SetRow(2, 0.0f, 0.0f, -1.0f * (farDist + nearDist) * invfn, -2.0f * farDist * nearDist * invfn);
+            }
             result.SetRow(3, 0.0f, 0.0f, -1.0f, 0.0f);
         }
         else
         {
+            // place the far plane at infinity
```
VickyAtAZ, line 47: Why set far plane to infinity if far dist is smaller than near dist?

moudgils: In theory you can use that to render sky, sun, moon, stars, etc. It basically takes f to infinity so -(f+n)/(f-n) essentially becomes -1. It has the effect of mapping vertex positions on the infinity distant far plane. You could also use it to render stencil shadow volumes too. But given that the non-stereoscopic projection function doesnt support this functionality, it wont be used for stereoscopic projections as well. Regardless let me update the comments in the header to make this clear.

VickyAtAZ (approved):

PR #26, opened 2022-09-09
fix device no found

byrcolin:
Signed-off-by: byrcolin <byrcolin@amazon.com>

### DIFF
--- a/Gems/XR/Code/Include/XR/XRInput.h
+++ b/Gems/XR/Code/Include/XR/XRInput.h
@@ -10,6 +10,7 @@
 
 #include <AzCore/Memory/SystemAllocator.h>
 #include <XR/XRBase.h>
+#include <XR/XRDevice.h>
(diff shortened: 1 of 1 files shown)


moudgils (approved):

Gems/XR/Code/Include/XR/XRInput.h
```diff
@@ -10,6 +10,7 @@
 
 #include <AzCore/Memory/SystemAllocator.h>
 #include <XR/XRBase.h>
+#include <XR/XRDevice.h>
```
amzn-phist, line 13: I fixed this locally too, but instead of including the header, you can just forward declare `class Device;` below in the namespace.

byrcolin: done

amzn-phist (approved):

PR #31, opened 2022-09-15
O3de extras ar

byrcolin:


Gems/XR/Code/Tests/XRTest.cpp
```diff
@@ -0,0 +1,31 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <AzTest/AzTest.h>
+#include "XRTest.h"
+
+#if !AZ_TRAIT_DISABLE_ALL_XR_TESTS
```
amzn-phist, line 12: In OpenXRVkTests.cpp the trait there begins with `O3DE`.  Here we should replace the `AZ` with `O3DE`.

Gems/OpenXRVk/Code/Tests/OpenXRVkTests.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+}
+
+#if O3DE_TRAIT_DISABLE_FAILED_OPENXRVK_TESTS
+TEST_F(OpenXRVkTestFixture, DISABLED_ExpectTrue)
+#else
+TEST_F(OpenXRVkTestFixture, ExpectTrue)
+#endif // O3DE_TRAIT_DISABLE_FAILED_OPENXRVK_TESTS
+{
+    EXPECT_TRUE(false);
```
amzn-phist, line 31: Just testing AR runs here, right?  This would be corrected before checkin?

amzn-phist (approved):

Gems/OpenXRVk/Code/Tests/OpenXRVkTests.cpp
byrcolin: Yes this is just testing

moraaar (approved):
LGTM.

Although I wonder if we are making it a bit overcomplicated to start with by PALifying the tests, having traits per platform and having macros to disable failing tests. Is this the default stuff is created by the scripts? Seems like an overkill.

Gems/OpenXRVk/Code/Tests/OpenXRVkTests.cpp
lumberyard-employee-dm: Looks like the place holder were submitted and should not have been.

PR #35, opened 2022-09-28
Small Fixes for Work With Project Manager Remote Templates

AMZN-Gene:
- Adding canonical tags.
- Fix origin field: this is now an "author" field. Now it references o3de.org as the author. 
- Fix repo.json to not use raw links (which are throttled)
- Deleting unused _origin_url_


repo.json
```diff
@@ -6,6 +6,6 @@
     "additional_info": "See the README.md at the root of this repository for more information",
     "last_updated": "2022-09-27",
     "templates": [
-        "https://raw.githubusercontent.com/o3de/o3de-extras/development/Templates/Multiplayer"
+        "https://github.com/o3de/o3de-extras/tree/development/Templates/Multiplayer"
```
byrcolin, line 9: not sure this is correct either, but for now ok.

byrcolin (approved):

repo.json
```diff
@@ -6,6 +6,6 @@
...
     ]
 }
```
lumberyard-employee-dm, line 11: nitpick: can you add a newline to this file

lumberyard-employee-dm (approved):

repo.json
AMZN-Gene: Fixed! @AMZN-Phil caught this as well. 
Updated the template link to use .git; the original returns the page you'd see in a web-browser, whereas .git causes the script to use the git API to get the raw file

repo.json
AMZN-Gene: fixed! thank you

AMZN-Phil (approved):

PR #44, opened 2022-10-12
Added setting to be able to scale the resolution of OpenXR views

moraaar:

The setting can make use of platform prefixes to only target one platform, for example:

````
{
    "O3DE": {
        "Atom": {
            "OpenXR": {
                "android_ViewResolutionScale": 0.5
            }
        }
    }
}
````

NOTE: This setting is a temporary measure that will help to control the performance when running on Quest 2 by lowering the resolution. Further optimizations to make the VR pipeline more performant overall will be done in the future.

### DIFF
--- a/Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
+++ b/Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
@@ -161,10 +161,11 @@ namespace OpenXRVk
             }
 
             double xrViewResolutionScale = 1.0;
-            if (AZ::SettingsRegistryInterface* settingsRegistry = AZ::SettingsRegistry::Get())
+            if (auto* settingsRegistry = AZ::SettingsRegistry::Get();
+                settingsRegistry != nullptr)
             {
                 AZ::SettingsRegistryMergeUtils::PlatformGet(*settingsRegistry, xrViewResolutionScale,
-                    "/O3DE/Atom", "OpenXRViewResolutionScale");
+                    "/O3DE/Atom/OpenXR", "ViewResolutionScale");
             }
 
             // Create a swapchain for each view.
--- a/Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
+++ b/Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
@@ -11,6 +11,8 @@
 #include <AzCore/Casting/numeric_cast.h>
 #include <AzCore/std/containers/set.h>
 #include <AzCore/std/containers/vector.h>
+#include <AzCore/Settings/SettingsRegistry.h>
+#include <AzCore/Settings/SettingsRegistryMergeUtils.h>
 #include <OpenXRVk/OpenXRVkDevice.h>
 #include <OpenXRVk/OpenXRVkInstance.h>
 #include <OpenXRVk/OpenXRVkSession.h>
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by moudgils — APPROVED** [review]

File: Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp [hunk 1]
```diff
@@ -158,10 +160,20 @@ namespace OpenXRVk
...
+            }
+
             // Create a swapchain for each view.
             for (uint32_t i = 0; i < m_numViews; i++)
             {
-                const XrViewConfigurationView& configView = m_configViews[i];
+                XrViewConfigurationView& configView = m_configViews[i];
+
+                configView.recommendedImageRectWidth = static_cast<uint32_t>(static_cast<double>(configView.recommendedImageRectWidth) * xrViewResolutionScale);
+                configView.recommendedImageRectHeight = static_cast<uint32_t>(static_cast<double>(configView.recommendedImageRectHeight) * xrViewResolutionScale);
```
amzn-phist, line 176: The outer `static_cast` should be `aznumeric_cast` due to precision loss.

Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
```diff
@@ -158,10 +160,20 @@ namespace OpenXRVk
                 AZ_Printf("OpenXRVk", "Swapchain Formats: %s\n", swapchainFormatsString.c_str());
             }
 
+            double xrViewResolutionScale = 1.0;
+            if (AZ::SettingsRegistryInterface* settingsRegistry = AZ::SettingsRegistry::Get())
```
amzn-phist, line 164: Can add a check on the `settingsRegistry` pointer...

`if (auto settingsRegistry = AZ::SettingsRegistry::Get(); settingsRegistry != nullptr)`

Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
```diff
@@ -158,10 +160,20 @@ namespace OpenXRVk
...
+            {
+                AZ::SettingsRegistryMergeUtils::PlatformGet(*settingsRegistry, xrViewResolutionScale,
+                    "/O3DE/Atom", "OpenXRViewResolutionScale");
```
amzn-phist, line 167: Since this is a new SR key for OpenXR, maybe we can organize it better.  For instance:
"/O3DE/Atom/OpenXR/ViewResolutionScale"

That way additional OpenXR render settings have a path they can be put under.

Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
jeremyong-az: `aznumeric_cast` actually is just synonymous to `static_cast` at the moment and shouldn't be used any more IMO

Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
moraaar: It's not necessary, I've seen both ways in the code base. But I'm ok changing it.

Gems/OpenXRVk/Code/Source/OpenXRVkSwapChain.cpp
moraaar: Good suggestion. I'll change "/O3DE/Atom/OpenXREnable" to "/O3DE/Atom/OpenXR/Enable" then so they are both consistent.

amzn-phist: I forgot about this "OpenXREnable" setting, does changing this have any effect on the `-openxr=enable` command-line arg that we've been using?

moraaar: >I forgot about this "OpenXREnable" setting, does changing this have any effect on the -openxr=enable command-line arg that we've been using?

No, you can keep using the command line option like that, but if you want to use the setting registry instead you can have a .setreg file with:

````
    "O3DE": {
        "Atom": {
            "OpenXR": {
                "Enable": true
            }
        }
    }
````

PR #46, opened 2022-10-17
Fix crash running editor/launcher on Quest 2 over link cable

moraaar:

Running OpenXR on host platforms, where the rendering in done over the PC's GPU, the vulkan context created from XR will be used for the PC version and it needs to to gather its features' functions supported by the physics devices features. It was crashing because although its physical device reported raytracing support, the context didn't load the vulkan ray tracing functions.

### DIFF
--- a/Gems/OpenXRVk/Code/Source/OpenXRVkDevice.cpp
+++ b/Gems/OpenXRVk/Code/Source/OpenXRVkDevice.cpp
@@ -91,12 +91,17 @@ namespace OpenXRVk
         }
 
         // Now that we have created the device, load the function pointers for it.
+#if AZ_TRAIT_OS_IS_HOST_OS_PLATFORM
+        bool functionsLoaded = xrVkInstance->GetFunctionLoader().LoadProcAddresses(
+            &xrVkInstance->GetContext(), xrVkInstance->GetNativeInstance(), xrVkInstance->GetActivePhysicalDevice(), m_xrVkDevice);
+#else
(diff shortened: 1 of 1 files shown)


Gems/OpenXRVk/Code/Source/OpenXRVkDevice.cpp
```diff
@@ -91,12 +91,17 @@ namespace OpenXRVk
         }
 
         // Now that we have created the device, load the function pointers for it.
+#if AZ_TRAIT_OS_IS_HOST_OS_PLATFORM
+        bool functionsLoaded = xrVkInstance->GetFunctionLoader().LoadProcAddresses(
+            &xrVkInstance->GetContext(), xrVkInstance->GetNativeInstance(), xrVkInstance->GetActivePhysicalDevice(), m_xrVkDevice);
+#else
```
amzn-phist, line 97: Instead of repeating the whole line, you could extract only the part that's different and wrap it under the OS trait.

```cpp
VkPhysicalDevice vkPhysicalDevice = VK_NULL_HANDLE;

#if AZ_TRAIT_OS_IS_HOST_OS_PLATFORM
    // ...
    vkPhysicalDevice = xrVkInstance->GetActivePhysicalDevice();
#endif

// ...
bool functionsLoaded = xrVkInstance->GetFunctionLoader().LoadProcAddresses(
    &xrVkInstance->GetContext(), xrVkInstance->GetNativeInstance(), vkPhysicalDevice, m_xrVkDevice);
```

amzn-phist (approved):

moudgils (approved):

PR #48, opened 2022-10-19
Implement Movement in XR via Controller Input

amzn-phist:
Implements a new camera movement controller component.

Simple controls to move the camera:
Left Thumb-Stick X and Y will translate along camera's Right and Forward vectors, respectively.
Buttons A and B will translate down and up along a fixed UP vector, respectively.

Debug Draw added which will display values of XR controller inputs.
This is tied to a Cvar: `xr_DebugDrawInput` (bool)

Also during application runtime, you can simply press `Menu + Left Trigger` to toggle the debug display.
(Menu button is on the left controller, and pull the left trigger fully)

Also made some improvements to PAL usage in the OpenXRVk Gem:
- Added missing platform files.
- Moved declarations to the Gem proper, and moved implementations to Common/Default or Common/Unimplemented files as needed.
- Also moved an input device `Create` function to PAL, and removed `#include`s that were using `..` to path into PAL directories.


Gems/OpenXRVk/Code/Include/OpenXRVk/XRCameraMovementInterface.h
```diff
@@ -0,0 +1,31 @@
...
+        : public AZ::ComponentBus
+    {
+    public:
+        AZ_RTTI(OpenXRVk::XRCameraMovementRequests, "{90D22228-2EAB-44FE-A7A0-693983DABEC6}");
+
+        // Put your public request methods here.
+        
+        // Put notification events here. Examples:
+        // void RegisterEvent(AZ::EventHandler<...> notifyHandler);
+        // AZ::Event<...> m_notifyEvent1;
```
moudgils, line 25:
```suggestion
        // AZ::Event<...> m_notifyEvent;
```

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
```diff
@@ -160,11 +180,21 @@ namespace AzFramework
 
         // Connect to haptic feedback request bus
         InputHapticFeedbackRequestBus::Handler::BusConnect(GetInputDeviceId());
+
+#if !defined(_RELEASE)
```
moudgils, line 184: Is this the normal pattern for using DebugDisplayEventBus where we need to protect it behind RELEASE macro_? I think it would be cleaner to move the macros within the Bus itself instead of making the calling code messy.

Gems/OpenXRVk/Code/Include/OpenXRVk/InputDeviceXRController.h
```diff
@@ -413,6 +437,13 @@ namespace AzFramework
         Implementation* GetImplementation() const;
 
     protected:
+        ////////////////////////////////////////////////////////////////////////////////////////////
+        // AzFramework::DebugDisplayEventBus interface
+#if !defined(_RELEASE)
+        void CheckDebugDrawCheat();
```
moudgils, line 443: Nice. Add a quick comment explaining the purpose of the methods.

Gems/OpenXRVk/Code/Source/OpenXRVkInput.cpp
```diff
@@ -356,7 +356,26 @@ namespace OpenXRVk
             LocateVisualizedSpace(device->GetPredictedDisplayTime(), session->GetXrSpace(spaceType),
                                   session->GetXrSpace(OpenXRVk::SpaceType::View), spaceType);
         }
+        
+        // XR to AZ vector conversion...
```
moudgils, line 360: In the comment add the fact that it is going from y-up to z-up

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include <AzCore/Component/Component.h>
+#include <AzCore/Component/TickBus.h>
+#include <AzFramework/Input/Events/InputChannelEventListener.h>
+
+#include <OpenXRVk/XRCameraMovementInterface.h>
+
+namespace OpenXRVk
+{
+    /*
```
moudgils, line 19: Commenting guidelines dictate using //!

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    void XRCameraMovementComponent::Deactivate()
+    {
+        XRCameraMovementRequestBus::Handler::BusDisconnect(GetEntityId());
+        AZ::TickBus::Handler::BusDisconnect();
+        AzFramework::InputChannelEventListener::Disconnect();
+    }
+
+    void XRCameraMovementComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint timePoint)
```
moudgils, line 96: Do you need [[maybe_unused]] here?

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include <Atom/RPI.Public/View.h>
+#include <AzFramework/Components/CameraBus.h>
+
+#include <Atom/RPI.Public/ViewportContext.h>
+#include <Atom/RPI.Public/ViewportContextBus.h>
+
+
+namespace OpenXRVk
+{
+    void XRCameraMovementComponent::Reflect(AZ::ReflectContext* context)
```
moudgils, line 28: Are we using this class within ASV samples for testing? If so, is there another PR for ASV work?

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.h
```diff
@@ -0,0 +1,116 @@
...
+        // AzFramework::InputChannelEventListener
+        bool OnInputChannelEventFiltered(const AzFramework::InputChannel& inputChannel) override;
+
+    private:
+        void OnXRControllerEvent(const AzFramework::InputChannel& inputChannel);
+        static AZ::Transform GetCameraTransformFromCurrentView();
+
+        // Transient data...
+        AZ::Vector3 m_movement{ AZ::Vector3::CreateZero() };    // Only use the X/Y components of this vector
+        float m_heightZ{ 0.f };
```
moudgils, line 109: We use following convention everywhere in XR gems so maybe good to stick with this everywhere. 
float m_heightZ = 0.f;

Gems/OpenXRVk/Code/Include/OpenXRVk/XRCameraMovementInterface.h
amzn-phist: I think it's indicating a series of (generically named) events to follow, `m_notifyEvent2` etc.

Anyways, this file was created from the `DefaultComponent` template, and it doesn't look like this interface file is needed.  I will look into removing it.

Gems/OpenXRVk/Code/Include/OpenXRVk/InputDeviceXRController.h
amzn-phist: Sure thing.

Gems/OpenXRVk/Code/Source/OpenXRVkInput.cpp
amzn-phist: Will do.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.h
amzn-phist: Ok I will change it.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.cpp
amzn-phist: No longer on `deltaTime`, that was placed before writing the body.  Will update.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.cpp
amzn-phist: I haven't modified ASV sample, but I can test things out there, see if we can integrate this component into it.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.h
amzn-phist: Sorry, it's my habit to use curly initializers nearly everywhere.  I'll update it.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
amzn-phist: I was just wrapping all the "debug" functionality with !_RELEASE because that makes sense to strip it out of release builds.  I can poke a little further and see how DebugDisplayEventBus handles RELEASE (if at all).  I'd prefer to keep this, but we can potentially hide it a bit better.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.h
amzn-phist: While I don't think there's anything wrong with the `/**/` comments, I went ahead and changed these to be `//!`.  I looked at https://github.com/o3de/sig-core/blob/main/governance/API-Ref-Guidelines-Update.md to see if it mentioned multi-line comments and it didn't show any in the examples.  However, I instantiated this file from the DefaultComponent template, so I guess either the templates should be updated or multi-line comments are okay to use.

Moving on...

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
amzn-phist: Updated this, let me know what you think.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
```diff
@@ -419,4 +450,307 @@ namespace AzFramework
...
+            cheatWasPressed = false;
+        }
+    }
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    void InputDeviceXRController::DrawGlobalDebugInfo()
+    {
+        CheckDebugDrawCheat();
+
+#if !defined(_RELEASE)
```
moraaar, line 531: Seem `CheckDebugDrawCheat` is not necessary either in release.

````
#if !defined(_RELEASE)
        CheckDebugDrawCheat();
````

File: Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp [hunk 10]
```diff
@@ -419,4 +450,307 @@ namespace AzFramework
...
+        if (OpenXRVk::xr_DebugDrawInput)
+        {
```
moraaar, line 533: [nit] Seems this can be an early return to avoid having a large block of code.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
```diff
@@ -419,4 +450,307 @@ namespace AzFramework
...
+            const auto& rawControllerData = m_impl->GetRawState();
+            DrawControllerAxes(*debugDisplay, rawControllerData.m_leftPositionState, rawControllerData.m_leftOrientationState);
+            DrawControllerAxes(*debugDisplay, rawControllerData.m_rightPositionState, rawControllerData.m_rightOrientationState);
+
+            float drawX = 20.f;     // current draw X
+            float drawY = 20.f;     // current draw Y
+            constexpr float textSize = 0.8f;
+            constexpr float lineHeight = 15.f;
+
+            const AZ::Color whiteColor{ 1.f, 1.f, 1.f, 1.f };
```
moraaar, line 555: [nit] There is already a white color defined in `AZ::Colors::White`

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.h
```diff
@@ -0,0 +1,97 @@
...
+        //! does not return an error and still activates this component.
+        static void GetDependentServices(AZ::ComponentDescriptor::DependencyArrayType& dependent);
+
+    protected:
+        //! Puts this component into an active state.
+        //! The system calls this function once during activation of each entity that owns the
+        //! component. You must override this function. The system calls a component's Activate()
+        //! function only if all services and components that the component depends on are present
+        //! and active.
+        void Activate() override;
```
moraaar, line 65: [minor] I would recommend removing all these comments that are already written in the super class. Otherwise the same large comment is going to be repeated and it's going to be difficult to maintain (if the comment in the super class changes then all the others would have to be changed too). So what I've seen normally is the comment in the class that defines the functions and no comments in the overrides.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.cpp
```diff
@@ -0,0 +1,167 @@
...
+        {   // down
+            m_movement.SetZ(-inputChannel.GetValue() * m_movementSensitivity);
+        }
+        if (channelId == AzFramework::InputDeviceXRController::Button::B)
+        {   // up
+            m_movement.SetZ(inputChannel.GetValue() * m_movementSensitivity);
+        }
+    }
+
+    // static
```
moraaar, line 148: left over comment? please remove `// static`.

moraaar (changes requested):

Gems/OpenXRVk/Code/Include/OpenXRVk/InputDeviceXRController.h
```diff
@@ -17,7 +19,15 @@
...
 #include <AzFramework/Input/Devices/InputDevice.h>
 
-////////////////////////////////////////////////////////////////////////////////////////////////////
+#include <AzFramework/Entity/EntityDebugDisplayBus.h>
+
+namespace OpenXRVk
+{
+    AZ_CVAR_EXTERNED(bool, xr_DebugDrawInput);
+
+} // namespace OpenXRVk
```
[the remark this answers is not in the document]
amzn-phist: Removed the extern cvar delcaration.

Gems/OpenXRVk/Code/Source/Platform/Mac/InputDeviceXRController_Platform.h
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <../Common/Unimplemented/InputDeviceXRController_Unimplemented.cpp>
```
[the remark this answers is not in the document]
amzn-phist: This was a mistake and has been fixed.  More fixes to the rest of the PAL stuff coming after.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
amzn-phist: I agree.  And other feedback wasn't too keen on having a ton of code wrapped in `!_RELEASE`, so I avoided doing that in the header, and in the cpp I wrapped the bodies of these two functions in `!_RELEASE`.  Hopefully that's a good compromise, at the expense that release has these empty functions to strip out.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
amzn-phist: Early return added.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
amzn-phist: Good to know, probably won't address it now.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.h
amzn-phist: Removed the comments.  Again, I instantiated this component from the `DefaultComponent` template.  It was my first time using a template for a new component that I intended to commit, so in the future I will be more cautious about leaving all the default code and comments in as-is.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.cpp
amzn-phist: Removed.  Although, I think this function could just be a static function in the cpp, doesn't need to be a class static.  I'll update it.

Gems/OpenXRVk/Code/Include/OpenXRVk/InputDeviceXRController.h
```diff
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#include <AzCore/Console/IConsole.h>
```
moraaar, line 11: I think you can move `#include <AzCore/Console/IConsole.h>` to the cpp now.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
```diff
@@ -419,4 +449,311 @@ namespace AzFramework
...
+        {
+            cheatWasPressed = false;
+        }
+#endif // !_RELEASE
+    }
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    void InputDeviceXRController::DrawGlobalDebugInfo()
+    {
+#if !defined(_RELEASE)
```
moraaar, line 530: [nit] By the way, please use `AZ_RELEASE_BUILD` instead of `_RELEASE`

Gems/OpenXRVk/Code/Include/OpenXRVk/InputDeviceXRController.h
amzn-phist: Done.

Gems/OpenXRVk/Code/Source/InputDeviceXRController.cpp
amzn-phist: Done.

Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkInstance.h
```diff
@@ -15,6 +15,26 @@
...
+    void OpenXRBeginFrameInternal();
+
+    //! Called when the device is ending a frame for processing.
+    //! @note This function is called from the thread related to the presentation queue.
+    void OpenXREndFrameInternal();
+
+    //! Called after the EndFrame has been executed.
+    //! @note This function is called from the main thread.
+    void OpenXRPostFrameInternal();
+}
```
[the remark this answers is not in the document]
amzn-phist: I've moved this block to its own file: Source/OpenXRVkCommon.h

If you don't like the name "Common" we can change that easily.  But this sets it up better and doesn't expose these functions to public includes.

Gems/OpenXRVk/Code/Source/Platform/Android/OpenXRVkInstance_Android.h
```diff
@@ -47,11 +47,11 @@ namespace OpenXRVk::Platform
 
         return true;
     }
-    
+
```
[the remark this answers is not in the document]
amzn-phist: This is in OpenXRVkCommon_Android.cpp now.

Gems/OpenXRVk/Code/Source/Platform/Common/Default/InputDeviceXRController_Default.h
```diff
@@ -0,0 +1,25 @@
...
+{
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // static
+    inline InputDeviceXRController::Implementation* InputDeviceXRController::Implementation::Create(
+        InputDeviceXRController& inputDevice)
+    {
+        // Future versions of this function may be able to select from a variety of different device
+        // types and do so based on knowledge of the hardware, but for now force the Oculus Touch controller.
+        return aznew OpenXRVk::InputDeviceOculusTouch(inputDevice);
+    }
```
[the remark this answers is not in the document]
amzn-phist: Done.

Gems/OpenXRVk/Code/Source/Platform/Common/Unimplemented/OpenXRVkInstance_Unimplemented.h
```diff
@@ -6,23 +6,23 @@
...
+#pragma once
 
 namespace OpenXRVk::Platform
 {
     bool OpenXRInitializeLoader()
     {
         return true;
     }
-    
+
```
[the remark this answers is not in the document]
amzn-phist: This is now in OpenXRVkCommon_Unimplemented.cpp.

Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkInstance.h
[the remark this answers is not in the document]
moraaar: This is better, thanks.

Gems/OpenXRVk/Code/Source/Platform/Android/OpenXRVkInstance_Android.h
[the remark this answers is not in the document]
moraaar: Great, thanks!

Gems/OpenXRVk/Code/Source/Platform/Common/Unimplemented/OpenXRVkInstance_Unimplemented.h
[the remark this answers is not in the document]
moraaar: Awesome! Thanks!

Gems/OpenXRVk/Code/Source/Platform/Common/Default/InputDeviceXRController_Default.cpp
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include <OpenXRVk/InputDeviceXRController.h>
+#include <../Common/Default/OculusTouch_Default.h>
```
moraaar, line 10: [suggestion] `#include <../Common/Default/OculusTouch_Default.h>` -> `#include "OculusTouch_Default.h"` since they are in the same folder. Normally I incline towards not using quotes, but in this case seems reasonable if it's not trivial to add `/Common/Default/` to the include path for PAL reasons.

moraaar (approved):
Nice! This is looking much better for the common PAL stuff. Thanks for going the extra mile. Approved.

Gems/OpenXRVk/Code/Source/Platform/Common/Default/InputDeviceXRController_Default.cpp
amzn-phist: Thanks for the review!  I agree with you, to prefer `<>` in most cases.  Having to `..` up a directory is not ideal, but as I mentioned before in PAL we commonly break certain conventions.  Afaik, we don't add Common/Default etc folders to CMake include paths, because you'd have to potentially change that per platform and it gets messy.  So in this case I think doing `""` here is an OK change to make.

Gems/OpenXRVk/Code/Source/XRCameraMovementComponent.cpp
amzn-phist: ASV sample already has some controls working, and they are similar functionality.  If we really want to, we can go do that, but I think for now we can get this through and address that later.

moudgils (approved):

PR #50, opened 2022-11-01
Fixes an issue with camera movement in Launcher

amzn-phist:
- Asset Processor was processing the XRCameraMovement component, but with a warning.  This caused the Camera entity to not have this component when spawned at runtime.
- The issue was that the OpenXRVk Gem's cmake didn't declar an alias for the .Builders variant.
- Also fixed some stuff on the Edit Context and re-saved the XR_Office level.

### DIFF
--- a/Gems/OpenXRVk/Code/CMakeLists.txt
+++ b/Gems/OpenXRVk/Code/CMakeLists.txt
@@ -79,9 +79,13 @@ ly_add_target(
             Gem::OpenXRVk.Static
 )
 
-# use the OpenXRVk module in clients and tools:
+# use the OpenXRVk module in all aliases:
 ly_create_alias(NAME OpenXRVk.Clients NAMESPACE Gem TARGETS Gem::OpenXRVk)
-ly_create_alias(NAME OpenXRVk.Tools NAMESPACE Gem TARGETS Gem::OpenXRVk)
+
+if(PAL_TRAIT_BUILD_HOST_TOOLS)
+    ly_create_alias(NAME OpenXRVk.Tools NAMESPACE Gem TARGETS Gem::OpenXRVk)
+    ly_create_alias(NAME OpenXRVk.Builders NAMESPACE Gem TARGETS Gem::OpenXRVk)
+endif()
 
 ################################################################################
 # Tests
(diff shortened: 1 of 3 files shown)


Gems/OpenXRVk/Code/CMakeLists.txt
```diff
@@ -79,9 +79,13 @@ ly_add_target(
...
 )
 
-# use the OpenXRVk module in clients and tools:
+# use the OpenXRVk module in all aliases:
 ly_create_alias(NAME OpenXRVk.Clients NAMESPACE Gem TARGETS Gem::OpenXRVk)
-ly_create_alias(NAME OpenXRVk.Tools NAMESPACE Gem TARGETS Gem::OpenXRVk)
+
+if(PAL_TRAIT_BUILD_HOST_TOOLS)
+    ly_create_alias(NAME OpenXRVk.Tools NAMESPACE Gem TARGETS Gem::OpenXRVk)
+    ly_create_alias(NAME OpenXRVk.Builders NAMESPACE Gem TARGETS Gem::OpenXRVk)
```
lumberyard-employee-dm, line 87: You may want to update the gem to reference the `${gem_name}` variable to make it easier to rename in the future.
https://github.com/o3de/o3de/blob/development/Templates/DefaultGem/Template/CMakeLists.txt
https://github.com/o3de/o3de/blob/development/Templates/DefaultGem/Template/Code/CMakeLists.txt

lumberyard-employee-dm (approved):

Gems/OpenXRVk/Code/CMakeLists.txt
amzn-phist: The "parent" cmake file doesn't appear to have set the `${gem_name}` variable:
https://github.com/o3de/o3de-extras/blob/development/Gems/OpenXRVk/CMakeLists.txt

Which makes me think this gem wasn't instantiated from the template.

lumberyard-employee-dm: Probably not

moraaar (approved):
Great finding!

PR #54, opened 2022-11-07
Materials to base Materials optimization. @shawstar

shawstar:
Signed-off-by: shawstar <shawstar@amazon.com>

Materials to base Materials optimization.


moudgils (approved):

Projects/OpenXRTest/project.json.bak0
```diff
@@ -0,0 +1,24 @@
...
+    "engine": "o3de",
+    "external_subdirectories": [
+        "Gem"
+    ],
+    "restricted": "OpenXRTest",
+    "gem_names": [
+        "XR",
+        "OpenXRVk"
+    ]
+}
```
amzn-phist, line 24: Don't commit this file.

amzn-phist (changes requested):

Projects/OpenXRTest/project.json.bak0
```diff
@@ -0,0 +1,24 @@
+{
+    "project_name": "OpenXRTest",
+    "project_id": "{947211d5-689a-437f-8ad7-7f558edcd40a}",
+    "origin": "The primary repo for OpenXRTest goes here: i.e. http://www.mydomain.com",
```
smurly, line 4: do we want to set something here for github/o3de/o3de-extras or whatever it is?

Projects/OpenXRTest/project.json.bak0
```diff
@@ -0,0 +1,24 @@
...
+    "license": "What license OpenXRTest uses goes here: i.e. https://opensource.org/licenses/Apache-2.0 Or https://opensource.org/licenses/MIT etc.",
```
smurly, line 5: are we Apache-2.0? I don't think we are MIT

smurly (approved):

Projects/OpenXRTest/project.json.bak0
amzn-phist: We already have a project.json: https://github.com/o3de/o3de-extras/blob/development/Projects/OpenXRTest/project.json

This file needs to be removed from this PR, but it is a good point.  We should update the real `project.json` for this project because it has default properties same as this file.

Projects/OpenXRTest/project.json.bak0
shawstar: Fixed.

PR #69, opened 2022-11-25
Move shiboken gem to extras, modify build system to use pyside find f…

sphrose:
…unctionality.

This PR adds a new gem to build Python bindings for the AzQtComponents library. It uses the pending changes to the find_pyside script in the 3P pyside, so as yet won't work, but has been tested using a hand build package.

Signed-off-by: sphrose <82213493+sphrose@users.noreply.github.com>


lawsonamzn:
I'd like to understand a little better here why this gem would be in extras.   is it not needed to run any part of O3DE or any of its dependencies?  So if someone makes a 'must have' gem for O3DE that uses AzQtComponents, you _don't_ need shiboken gem active?

alexmontAmazon:
> I'd like to understand a little better here why this gem would be in extras. is it not needed to run any part of O3DE or any of its dependencies? So if someone makes a 'must have' gem for O3DE that uses AzQtComponents, you _don't_ need shiboken gem active? You can still use the gem and that gem could still be part of O3DE (not extras) without requiring shiboken? Or does it become a link dependency?

Adding @lumberyard-employee-dm who is probably best suited to answer this...

lumberyard-employee-dm:
> > I'd like to understand a little better here why this gem would be in extras. is it not needed to run any part of O3DE or any of its dependencies? So if someone makes a 'must have' gem for O3DE that uses AzQtComponents, you _don't_ need shiboken gem active? You can still use the gem and that gem could still be part of O3DE (not extras) without requiring shiboken? Or does it become a link dependency?
> 
> Adding @lumberyard-employee-dm who is probably best suited to answer this...

To answer Nick's questions, the AzQtComponentsForPythonGem isn't a core gem of O3DE. The AzQtComponentsForPythongem would open up Python support for accessing AzQtComponents. Another gem can depend on AzQtComponents without bringing the AzQtComponentsForPython gem dependency.

lawsonamzn:
> > > I'd like to understand a little better here why this gem would be in extras. is it not needed to run any part of O3DE or any of its dependencies? So if someone makes a 'must have' gem for O3DE that uses AzQtComponents, you _don't_ need shiboken gem active? You can still use the gem and that gem could still be part of O3DE (not extras) without requiring shiboken? Or does it become a link dependency?
> > 
> > 
> > Adding @lumberyard-employee-dm who is probably best suited to answer this...
> 
> To answer Nick's questions, the AzQtComponentsForPythonGem isn't a core gem of O3DE. The AzQtComponentsForPythongem would open up Python support for accessing AzQtComponents. Another gem can depend on AzQtComponents without bringing the AzQtComponentsForPython gem dependency.

I'm fine with that, it just kind of depends on how deeply we lean into using python for our editor scripting.  I can see this coming back to the main repo in the future, if we start (re) writing core editor ui tools to be leveraging python for Tech Artist availability.  But it can always move back if it has to

byrcolin:
The sig responsible for this code should be involved in the decision on where it goes, i.e. if it is core or extra or what...
Looking at this PR, it is fine to submit it this way, however as these are all new files it will lose the git history of these files, which again is ok if that is what you want, but we could submit them from whatever git repo they currently live in with histories, https://github.com/o3de/sig-platform/wiki/Procedure-to-move-code-from-one-repo-to-another-with-history

lsemp3d (approved):

lawsonamzn (approved):
The changes look fine, assuming its okay for o3de-extras.

lumberyard-employee-dm:
> I still do not see the SIG who is going to own this sign off on it. I would also like to see the codeowners file updated in this PR. https://github.com/o3de/o3de-extras/blob/development/.github/CODEOWNERS

My vote would be for SIG-Content to own this area of the code by adding @o3de/sig-content-maintainers and @o3de/sig-content-reviewers.

Gems/AzQtComponentsForPython/Code/license_header.txt
```diff
@@ -0,0 +1,8 @@
+/*
```
lumberyard-employee-dm, line 1: This text file is unneeded since all it is an empty file with the O3DE copyright blerb.
The LICENSE.txt in the root of this repo covers the Apache 2.0 license for this gem

Gems/AzQtComponentsForPython/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
```
lumberyard-employee-dm, line 8: I advise adding this code blerb to this file to make it easier to rename the gem in the future if need be.
```
 Query the gem name from the gem.json file if possible
# otherwise fallback to using ${Name}
o3de_find_ancestor_gem_root(gempath gem_name "${CMAKE_CURRENT_SOURCE_DIR}")
if (NOT gem_name)
    set(gem_name "AzQtComponenntsForPython")
endif()
```

Gems/AzQtComponentsForPython/Code/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+    set(QT_MODULE_DIR ${QT_DOWNLOAD_LOCATION}/${QT_PACKAGE_NAME}/qt)
+                     
+    # The library we're creating bindings for.
+    set(AZQTCOMPONENTS_LIBRARY "AzQtComponents")
+        
+    #Define the shared library for which we will create bindings.
+    set_property(TARGET ${AZQTCOMPONENTS_LIBRARY} PROPERTY PREFIX "")
+    
+    add_shiboken_project(
+        NAME AzQtComponentsForPython
```
lumberyard-employee-dm, line 28: With the change to query the "gem_name" from the `gem.json` in theprevious comment, I recommend replacing AzQtComponentsForPython with `${gem_name}`.
```suggestion
        NAME ${gem_name}
```

Gems/AzQtComponentsForPython/Code/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+    ly_package_get_target_folder(${QT_PACKAGE_NAME} QT_DOWNLOAD_LOCATION)
+    set(QT_MODULE_DIR ${QT_DOWNLOAD_LOCATION}/${QT_PACKAGE_NAME}/qt)
+                     
+    # The library we're creating bindings for.
+    set(AZQTCOMPONENTS_LIBRARY "AzQtComponents")
+        
+    #Define the shared library for which we will create bindings.
+    set_property(TARGET ${AZQTCOMPONENTS_LIBRARY} PROPERTY PREFIX "")
+    
+    add_shiboken_project(
```
lumberyard-employee-dm, line 27: Where is this function defined at?
I don't see it defined anywhere, so Cmake will have a configuration error.

Gems/AzQtComponentsForPython/Code/CMakeLists.txt
sphrose: Done

Gems/AzQtComponentsForPython/Code/CMakeLists.txt
sphrose: This is part of the pyside2 package (https://github.com/o3de/3p-package-source/blob/main/package-system/pyside2/Findpyside2.cmake)

Gems/AzQtComponentsForPython/Code/license_header.txt
sphrose: This file is passed into the shiboken command line and is placed at the top of all files it generates. If this is unnecessary, will remove.

Gems/AzQtComponentsForPython/Code/CMakeLists.txt
lumberyard-employee-dm: Thanks for the link, I can see the function.
I have no other blocking comments for this PR.

AMZN-alexpete (approved):

PR #78, opened 2022-12-14
Fix Vulkan extension loading on Quest2

akioCL:
Fixed loading of Vulkan device extensions when running on the Quest2 that was introduced by https://github.com/o3de/o3de-extras/pull/43
This PR is also fixing the crash when using `adb shell ovrgpuprofiler -e` command that the other PR was trying to fix.
For some reason the DEBUG_UTILS extension is reported as available but the function pointers do not load.

Signed-off-by: Akio Gaule <10719597+akioCL@users.noreply.github.com>


Gems/OpenXRVk/Code/Source/OpenXRVkDevice.cpp
```diff
@@ -91,20 +91,18 @@ namespace OpenXRVk
...
             // Now that we have created the device, load the function pointers for it.
             const bool functionsLoaded = xrVkInstance->GetFunctionLoader().LoadProcAddresses(
                 &xrVkInstance->GetContext(), xrVkInstance->GetNativeInstance(), xrVkPhysicalDevice, m_xrVkDevice);
             m_context = xrVkInstance->GetContext();
+            // In some cases (like when running with the GPU profiler on Quest2) the extension is reported as available
+            // but the function pointers do not load. Disable the extension if that's the case.
+            if (m_context.EXT_debug_utils && !m_context.CmdBeginDebugUtilsLabelEXT)
+            {
+                m_context.EXT_debug_utils = 0;
+            }
```
moraaar, line 104: The instance's context needs to be updated as well. Suggesting:

````
            // Now that we have created the device, load the function pointers for it.
            auto& instanceContext = xrVkInstance->GetContext();
            const bool functionsLoaded = xrVkInstance->GetFunctionLoader().LoadProcAddresses(
                &instanceContext, xrVkInstance->GetNativeInstance(), xrVkPhysicalDevice, m_xrVkDevice);
            // In some cases (like when running with the GPU profiler on Quest2) the extension is reported as available
            // but the function pointers do not load. Disable the extension if that's the case.
            if (instanceContext.EXT_debug_utils && !instanceContext.CmdBeginDebugUtilsLabelEXT)
            {
                instanceContext.EXT_debug_utils = 0;
            }
            m_context = instanceContext;
````

**Review by rgba16f — APPROVED** [review]

File: Gems/OpenXRVk/Code/Source/OpenXRVkInstance.cpp [hunk 2]
```diff
@@ -412,6 +415,16 @@ namespace OpenXRVk
         return m_supportedXRDevices[m_physicalDeviceActiveIndex];
     }
 
+    void Instance::FilterAvailableExtensions(GladVulkanContext& context) const
+    {
+        // In some cases (like when running with the GPU profiler on Quest2) the extension is reported as available
```
moudgils, line 420: Can you add that this was specifically when we try to use "adb shell ovrgpuprofiler -e" so it is clear when this code applies

moudgils (approved):

Gems/OpenXRVk/Code/Source/OpenXRVkDevice.cpp
moraaar: This `LoadProcAddresses` call is using instance's context at the moment, if intance and device don't need to have their context with the same data then please use the device's context directly.

But test it because I remember that crashed.

moraaar: Great finding. Good job.

moraaar (approved):

PR #90, opened 2023-01-12
 Introduced XACRO parsing to ROS2 Gem

michalpelka:
The URDF importer is able to call "xacro" to convert xacro to urdf. It will show the page in the wizard with parameters.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "XacroParamsPage.h"
+#include <RobotImporter/RobotImporterWidget.h>
+
+namespace ROS2
+{
+
+    XacroParamsPage::XacroParamsPage(RobotImporterWidget* parent)
+        : QWizardPage(parent)
+    {
+        setTitle(tr("Xacro Parameters"));
+        m_table = new QTableWidget(this);
+
+        m_table->setColumnCount(2);
+        m_table->horizontalHeader()->setSectionResizeMode(0, QHeaderView::Stretch);
+        m_table->setShowGrid(true);
+        m_table->setSelectionMode(QAbstractItemView::SingleSelection);
+        m_table->setSelectionBehavior(QAbstractItemView::SelectRows);
+        m_table->setHorizontalHeaderLabels({ tr("Parameter"), tr("Value") });
+        m_table->horizontalHeader()->setStretchLastSection(true);
+
+        QVBoxLayout* layout = new QVBoxLayout;
+        layout->addWidget(m_table);
+        this->setLayout(layout);
+    }
+
+    bool XacroParamsPage::isComplete() const
+    {
+        return true;
+    }
+
+    void XacroParamsPage::SetParameters(const Utils::xacro::Params& params)
+    {
+        m_defaultParams = params;
+        m_table->setRowCount(0);
+        for (const auto& [name, value] : params)
+        {
+            int i = m_table->rowCount();
+            m_table->setRowCount(i + 1);
+            QTableWidgetItem* p1 = new QTableWidgetItem(QString::fromUtf8(name.data()));
+            QTableWidgetItem* p2 = new QTableWidgetItem(QString::fromUtf8(value.data()));
+            m_table->setItem(i, 0, p1);
+            m_table->setItem(i, 1, p2);
+        }
+    }
+
+    Utils::xacro::Params XacroParamsPage::GetParams() const
+    {
+        Utils::xacro::Params params;
+        const int rowCount = m_table->rowCount();
+        for (int i = 0; i < rowCount; i++)
+        {
+            const auto p1 = m_table->item(i, 0);
+            const auto p2 = m_table->item(i, 1);
+            AZ_Assert(p1, "cell should exist");
+            AZ_Assert(p2, "cell should exist");
+            AZStd::string name(p1->text().toUtf8());
+            AZStd::string value(p2->text().toUtf8());
+            auto it = m_defaultParams.find(name);
+            if (it != m_defaultParams.end())
+            {
+                AZ_Printf("XacroParamsPage", "name : %s value : %s (default %s)\n", name.c_str(), value.c_str(), it->second.c_str());
+                if (it->second != value)
+                {
+                    params[name] = value;
+                }
+            }
+        }
+        AZ_Printf("XacroParamsPage", "number of modified parameters %d\n", params.size());
+        return params;
+    }
+
+} // namespace ROS2
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
@@ -28,7 +28,7 @@ namespace ROS2
 
         QVBoxLayout* layout = new QVBoxLayout;
         layout->addWidget(m_table);
-        this->setLayout(layout);
+        setLayout(layout);
     }
 
     bool XacroParamsPage::isComplete() const
@@ -36,37 +36,36 @@ namespace ROS2
         return true;
     }
 
-    void XacroParamsPage::SetParameters(const Utils::xacro::Params& params)
+    void XacroParamsPage::SetXacroParameters(const Utils::xacro::Params& params)
     {
         m_defaultParams = params;
         m_table->setRowCount(0);
         for (const auto& [name, value] : params)
         {
-            int i = m_table->rowCount();
-            m_table->setRowCount(i + 1);
+            const int rowIndex = m_table->rowCount();
+            m_table->setRowCount(rowIndex + 1);
             QTableWidgetItem* p1 = new QTableWidgetItem(QString::fromUtf8(name.data()));
             QTableWidgetItem* p2 = new QTableWidgetItem(QString::fromUtf8(value.data()));
-            m_table->setItem(i, 0, p1);
-            m_table->setItem(i, 1, p2);
+            m_table->setItem(rowIndex, 0, p1);
+            m_table->setItem(rowIndex, 1, p2);
         }
     }
 
-    Utils::xacro::Params XacroParamsPage::GetParams() const
+    Utils::xacro::Params XacroParamsPage::GetXacroParameters() const
     {
         Utils::xacro::Params params;
         const int rowCount = m_table->rowCount();
         for (int i = 0; i < rowCount; i++)
         {
-            const auto p1 = m_table->item(i, 0);
-            const auto p2 = m_table->item(i, 1);
+            const auto* p1 = m_table->item(i, 0);
+            const auto* p2 = m_table->item(i, 1);
             AZ_Assert(p1, "cell should exist");
             AZ_Assert(p2, "cell should exist");
-            AZStd::string name(p1->text().toUtf8());
-            AZStd::string value(p2->text().toUtf8());
+            const AZStd::string name(p1->text().toUtf8().constData());
+            const AZStd::string value(p2->text().toUtf8().constData());
             auto it = m_defaultParams.find(name);
             if (it != m_defaultParams.end())
             {
-                AZ_Printf("XacroParamsPage", "name : %s value : %s (default %s)\n", name.c_str(), value.c_str(), it->second.c_str());
                 if (it->second != value)
                 {
                     params[name] = value;
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#if !defined(Q_MOC_RUN)
+#include <QFileDialog>
+#include <QLabel>
+#include <QLineEdit>
+#include <QPushButton>
+#include <QString>
+#include <QTableWidget>
+#include <QWizardPage>
+#include <RobotImporter/xacro/XacroUtils.h>
+#endif
+
+namespace ROS2
+{
+    class RobotImporterWidget;
+    class XacroParamsPage : public QWizardPage
+    {
+        Q_OBJECT
+    public:
+        explicit XacroParamsPage(RobotImporterWidget* parent);
+
+        void setSuccess(bool success);
+        bool isComplete() const override;
+
+        void SetParameters(const Utils::xacro::Params& params);
+        Utils::xacro::Params GetParams() const;
+
+    private:
+        Utils::xacro::Params m_defaultParams;
+        QTableWidget* m_table;
+    };
+} // namespace ROS2
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
@@ -22,20 +22,21 @@
 namespace ROS2
 {
     class RobotImporterWidget;
+
+    //!  Wizard page that allows user to modify Xacro parameters
     class XacroParamsPage : public QWizardPage
     {
         Q_OBJECT
     public:
         explicit XacroParamsPage(RobotImporterWidget* parent);
 
-        void setSuccess(bool success);
         bool isComplete() const override;
 
-        void SetParameters(const Utils::xacro::Params& params);
-        Utils::xacro::Params GetParams() const;
+        void SetXacroParameters(const Utils::xacro::Params& params);
+        Utils::xacro::Params GetXacroParameters() const;
 
     private:
         Utils::xacro::Params m_defaultParams;
-        QTableWidget* m_table;
+        QTableWidget* m_table {};
     };
 } // namespace ROS2
--- a/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
@@ -29,9 +29,11 @@ namespace ROS2
         m_checkUrdfPage = new CheckUrdfPage(this);
         m_assetPage = new CheckAssetPage(this);
         m_prefabMakerPage = new PrefabMakerPage(this);
+        m_xacroParamsPage = new XacroParamsPage(this);
 
         addPage(m_introPage);
         addPage(m_fileSelectPage);
+        addPage(m_xacroParamsPage);
         addPage(m_checkUrdfPage);
         addPage(m_assetPage);
         addPage(m_prefabMakerPage);
@@ -78,12 +80,35 @@ namespace ROS2
 
     void RobotImporterWidget::OpenUrdf()
     {
-        m_urdfPath = AZStd::string(m_fileSelectPage->getFileName().toUtf8().constData());
+        QString report;
         if (!m_urdfPath.empty())
         {
-            AZ_Printf("Wizard", "Loading URDF file : %s", m_urdfPath.c_str());
-            m_parsedUrdf = UrdfParser::ParseFromFile(m_urdfPath);
-            QString report;
+            if (m_urdfPath.ends_with("xacro"))
+            {
+                Utils::xacro::ExecutionOutcome outcome = Utils::xacro::ParseXacro(m_urdfPath, m_params);
+                if (outcome)
+                {
+                    m_parsedUrdf = outcome.m_urdfHandle;
+                    report += "# " + tr("XACRO execution succeed") + "\n";
+                }
+                else
+                {
+                    report += "# " + tr("XACRO parsing failed") + "\n";
+                    report += "\n\n" + tr("Command called : \n'") + QString::fromUtf8(outcome.m_called.data()) + "'";
+                    report += "\n\n" + tr("Process failed with code : ") + QString::number(outcome.m_returnCode);
+                    report += "\n\n" + tr("error output") + " :\n\n";
+                    report += QString::fromLocal8Bit(outcome.m_logErrorOutput.data(), static_cast<int>(outcome.m_logErrorOutput.size()));
+                    m_checkUrdfPage->ReportURDFResult(report, false);
+                    return;
+                }
+            }
+            else
+            {
+                // standard URDF
+                AZ_Printf("Wizard", "Loading urdf file : %s", m_urdfPath.c_str());
+                m_parsedUrdf = UrdfParser::ParseFromFile(m_urdfPath);
+            }
+
             const auto log = UrdfParser::GetUrdfParsingLog();
             if (m_parsedUrdf)
             {
@@ -191,6 +216,24 @@ namespace ROS2
     {
         if (currentPage() == m_fileSelectPage)
         {
+            m_params.clear();
+            m_urdfPath = AZStd::string(m_fileSelectPage->getFileName().toUtf8().constData());
+            if (IsFileXacro(m_urdfPath))
+            {
+                m_params = Utils::xacro::GetParameterFromXacroFile(m_urdfPath);
+                AZ_Printf("RobotImporterWidget", "Xacro has %d arguments\n", m_params.size()) m_xacroParamsPage->SetParameters(m_params);
+            }
+            // no need to wait for param page - parse urdf now, nextId will skip unnecessary pages
+            if (m_params.empty())
+            {
+                OpenUrdf();
+            }
+        }
+        if (currentPage() == m_xacroParamsPage)
+        {
+            AZ_Printf("RobotImporterWidget", "Validate m_xacroParamsPage\n");
+
+            m_params = m_xacroParamsPage->GetParams();
             OpenUrdf();
         }
         return currentPage()->validatePage();
@@ -198,7 +241,7 @@ namespace ROS2
 
     int RobotImporterWidget::nextId() const
     {
-        if (currentPage() == m_fileSelectPage)
+        if ((currentPage() == m_fileSelectPage && m_params.empty()) || currentPage() == m_xacroParamsPage)
         {
             if (m_parsedUrdf)
             {
@@ -312,4 +355,10 @@ namespace ROS2
         QMessageBox::critical(this, QObject::tr("Error"), errorMessage);
         AZ_Error("RobotImporterWidget", false, "%s", errorMessage.toUtf8().constData());
     }
+
+    bool RobotImporterWidget::IsFileXacro(const AZStd::string& filename) const
+    {
+        return filename.ends_with(".xacro");
+    }
+
 } // namespace ROS2
--- a/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
@@ -83,9 +83,9 @@ namespace ROS2
         QString report;
         if (!m_urdfPath.empty())
         {
-            if (m_urdfPath.ends_with("xacro"))
+            if (IsFileXacro(m_urdfPath))
             {
-                Utils::xacro::ExecutionOutcome outcome = Utils::xacro::ParseXacro(m_urdfPath, m_params);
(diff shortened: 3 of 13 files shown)


Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
```diff
@@ -0,0 +1,80 @@
...
+        return true;
+    }
+
+    void XacroParamsPage::SetParameters(const Utils::xacro::Params& params)
+    {
+        m_defaultParams = params;
+        m_table->setRowCount(0);
+        for (const auto& [name, value] : params)
+        {
+            int i = m_table->rowCount();
```
hultonha, line 45: nit: we could name `i` to `rowIndex` instead just to improve readability slightly

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
```diff
@@ -0,0 +1,80 @@
...
+            m_table->setRowCount(i + 1);
+            QTableWidgetItem* p1 = new QTableWidgetItem(QString::fromUtf8(name.data()));
+            QTableWidgetItem* p2 = new QTableWidgetItem(QString::fromUtf8(value.data()));
+            m_table->setItem(i, 0, p1);
+            m_table->setItem(i, 1, p2);
+        }
+    }
+
+    Utils::xacro::Params XacroParamsPage::GetParams() const
```
hultonha, line 54: nit: We have `GetParams` and `SetParameters`, it'd be good to be consistent with the abbreviation or not, so maybe they're both `GetParameters` and `SetParameters`?

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
```diff
@@ -0,0 +1,80 @@
...
+            const auto p1 = m_table->item(i, 0);
+            const auto p2 = m_table->item(i, 1);
+            AZ_Assert(p1, "cell should exist");
+            AZ_Assert(p2, "cell should exist");
+            AZStd::string name(p1->text().toUtf8());
+            AZStd::string value(p2->text().toUtf8());
+            auto it = m_defaultParams.find(name);
+            if (it != m_defaultParams.end())
+            {
+                AZ_Printf("XacroParamsPage", "name : %s value : %s (default %s)\n", name.c_str(), value.c_str(), it->second.c_str());
```
hultonha, line 69: question: Just double checking we want to leave the prints in permanently. If they're helpful to see in the log that's great but just wanted to check if they should be removed after debugging for example.

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
```diff
@@ -0,0 +1,41 @@
...
+#include <QString>
+#include <QTableWidget>
+#include <QWizardPage>
+#include <RobotImporter/xacro/XacroUtils.h>
+#endif
+
+namespace ROS2
+{
+    class RobotImporterWidget;
+    class XacroParamsPage : public QWizardPage
```
hultonha, line 25: suggestion: An API comment would be really helpful here just to state briefly what XacroParamsPage does and is responsible for

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -78,12 +80,35 @@ namespace ROS2
...
+                    report += "\n\n" + tr("error output") + " :\n\n";
+                    report += QString::fromLocal8Bit(outcome.m_logErrorOutput.data(), static_cast<int>(outcome.m_logErrorOutput.size()));
+                    m_checkUrdfPage->ReportURDFResult(report, false);
+                    return;
+                }
+            }
+            else
+            {
+                // standard URDF
+                AZ_Printf("Wizard", "Loading urdf file : %s", m_urdfPath.c_str());
```
hultonha, line 108: update: Looks like this print should be removed?

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -78,12 +80,35 @@ namespace ROS2
...
     void RobotImporterWidget::OpenUrdf()
     {
-        m_urdfPath = AZStd::string(m_fileSelectPage->getFileName().toUtf8().constData());
+        QString report;
         if (!m_urdfPath.empty())
         {
-            AZ_Printf("Wizard", "Loading URDF file : %s", m_urdfPath.c_str());
-            m_parsedUrdf = UrdfParser::ParseFromFile(m_urdfPath);
-            QString report;
+            if (m_urdfPath.ends_with("xacro"))
```
hultonha, line 86: suggestion: For now this is fine but if in future there are more urdf formats we might want to consider refactoring this so it is easy to add new cases without modifying an if/else or switch block (if there's just there's two it's definitely not worth it but just something that could help maintainability in future potentially

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -191,14 +216,32 @@ namespace ROS2
...
+            }
+            // no need to wait for param page - parse urdf now, nextId will skip unnecessary pages
+            if (m_params.empty())
+            {
+                OpenUrdf();
+            }
+        }
+        if (currentPage() == m_xacroParamsPage)
+        {
+            AZ_Printf("RobotImporterWidget", "Validate m_xacroParamsPage\n");
```
hultonha, line 234: nit: Should this printf now be removed?

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
hultonha, line 86: suggestion: We could use the helper `IsFileXacro` call here instead of having the same logic again (`if (IsFileXacro(m_urdfPath)`)

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -73,12 +74,25 @@ namespace ROS2::Utils
                     fullSourcePath.c_str(),
                     crc);
 
-                AvailableAsset t;
-                t.m_sourceAssetRelativePath = info.m_relativePath;
-                t.m_assetId = info.m_assetId;
-                t.m_sourceAssetGlobalPath = fullSourcePathStr;
-                t.m_productAssetRelativePath = info.m_relativePath;
-                availableAssets[crc] = t;
+                AvailableAsset foundAsset;
```
hultonha, line 77: question: This code looks very similar to that from another PR? Maybe once the other is merged this diff will go away? Just looks like this PR has some other unrelated changes in it potentially?

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+                        value = AZStd::string{ attributeDefault->value() };
+                    }
+                    params.emplace(AZStd::make_pair(AZStd::move(name), AZStd::move(value)));
+                }
+            }
+        }
+        return params;
+    }
+
+    AZStd::unordered_map<AZStd::string, AZStd::string> GetParameterFromXacroFile(AZStd::string filename)
```
hultonha, line 121: nit: Would be nice to pass the filename by `const` ref (or use a `string_view` potentially)

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+        AZ::IO::SizeType length = fileStream.GetLength();
+        if (length == 0)
+        {
+            return xacro::Params();
+        }
+
+        AZStd::vector<char> charBuffer;
+        charBuffer.resize_no_construct(length + 1);
+        fileStream.Read(length, charBuffer.data());
+        charBuffer.back() = 0;
```
hultonha, line 138: nit: Here it might be clearer to use `'\0'` to indicate EOF character

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace ROS2::Utils::xacro
+{
+
+    static const char* kXacroExecutable = "xacro";
+    using Params = AZStd::unordered_map<AZStd::string, AZStd::string>;
+    //! Structure that keeps all artifacts of xacro execution
+    struct ExecutionOutcome
+    {
+        //! parsed URDF from successful xacro's output
```
hultonha, line 23: uber nit: This types of comments in .h files tend to start with upper case characters (follow normal sentence rules) - so this just becomes `//! Parsed URDF from successful xacro's output.`

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.h
```diff
@@ -0,0 +1,51 @@
...
+        int m_returnCode{ 255 };
+        //! called program name
+        AZStd::string m_called;
+        //! standard output of xacro execution
+        AZStd::string m_logStandardOutput;
+        //! standard error output of xacro process
+        AZStd::string m_logErrorOutput;
+
+        //! Gets if execution was a success
+        operator bool()
```
hultonha, line 35: suggestion: I believe it's preferred to use `explicit` when using operator `bool` to help guard against any weird implicit conversions (this article has a not rundown - https://www.kdab.com/explicit-operator-bool/)

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
```diff
@@ -440,4 +447,16 @@ namespace UnitTest
         EXPECT_EQ(result, "/home/foo/ros_ws/install/foo_robot/meshes/bar.dae");
     }
 
+    TEST_F(UrdfParserTest, XacroParseArgs)
```
hultonha, line 450: Really awesome to see the tests! If there's time to check any other cases (maybe some incorrect data as input) to see how the code handles it, that would be great, but it's always a balance with how far you want to go

hultonha (commented):
Overall looks really solid, I just have a few questions/suggestions and wanted to check about some code that maybe shouldn't be in this PR. Happy to approve once you've had a look through my comments. Thanks!

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
```diff
@@ -440,4 +447,16 @@ namespace UnitTest
...
+    {
+        AZStd::string st = GetXacroParams();
```
moraaar, line 452: `st` -> `xacroParams`

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
```diff
@@ -440,4 +447,16 @@ namespace UnitTest
...
+        AZStd::vector<char> data;
+        data.resize(st.size());
+        AZStd::copy(st.begin(), st.end(), data.begin());
+        ROS2::Utils::xacro::Params p = ROS2::Utils::xacro::GetParameterFromXacroData(data);
```
moraaar, line 456: `p` -> `params`

moraaar (commented):
A few more minor suggestions, agree with Tom's comments. Great job!

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
```diff
@@ -0,0 +1,41 @@
...
+
+        void setSuccess(bool success);
+        bool isComplete() const override;
+
+        void SetParameters(const Utils::xacro::Params& params);
+        Utils::xacro::Params GetParams() const;
+
+    private:
+        Utils::xacro::Params m_defaultParams;
+        QTableWidget* m_table;
```
lumberyard-employee-dm, line 39:
```suggestion
        QTableWidget* m_table{};
```

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.h
lumberyard-employee-dm: It is definitely perfected to use explicit here and something we should add to the [C++ best practices guide](https://github.com/o3de/sig-core/blob/main/governance/C%2B%2B-Best-Practices-Guide.md).

Also this function should be marked **const** as well.
```suggestion
        explicit operator bool() const
```

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+        myProcess.start(program, arguments);
+        if (myProcess.waitForFinished() && myProcess.exitStatus() == QProcess::NormalExit && myProcess.exitCode() == 0)
+        {
+            auto output = myProcess.readAllStandardOutput().trimmed();
+            return AZStd::string{ output.data(), static_cast<AZStd::string::size_type>(output.size()) };
+        }
+        return AZStd::string{};
+    }
+
+    AZStd::unordered_map<AZStd::string, AZStd::string> GetParameterFromXacroData(AZStd::vector<char> data)
```
lumberyard-employee-dm, line 89: It would be better to accept an `AZStd::span<char>` here and that would allow the any type of contiguous structure of `char` to be supplied to this method(`AZStd::vector<char>`, `AZStd::array<char, N>`, `AZStd::string`, `AZStd::fixed_string<N>`, `AZStd::string_view`, `AZStd::fixed_vector<char, N>`, etc...)

Here is the API for https://en.cppreference.com/w/cpp/container/span

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+        {
+            AZ_Printf("ParseXacro", "xacro finished with success");
+            auto output = myProcess.readAllStandardOutput();
+            outcome.m_urdfHandle = UrdfParser::Parse(output.data());
+        }
+        else
+        {
+            AZ_Printf("ParseXacro", "xacro finished with code %d", myProcess.exitCode());
+            auto std = myProcess.readAllStandardOutput();
+            auto cerr = myProcess.readAllStandardError();
```
lumberyard-employee-dm, line 66: I recommend renaming the `cerr` in case some previous include file includes `<iostream>` and performs a `using namespace std;`, which would cause the `std::cerr` variable to be brought into scope.

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
```diff
@@ -0,0 +1,41 @@
...
+namespace ROS2
+{
+    class RobotImporterWidget;
+    class XacroParamsPage : public QWizardPage
+    {
+        Q_OBJECT
+    public:
+        explicit XacroParamsPage(RobotImporterWidget* parent);
+
+        void setSuccess(bool success);
```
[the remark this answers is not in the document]
spham-amzn: as @hultonha pointed out, just the signature is specified, but there is no body. The compiler will report this as any error, so if its not used or implemented, it should be removed.

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -78,12 +80,35 @@ namespace ROS2
...
+                    m_parsedUrdf = outcome.m_urdfHandle;
+                    report += "# " + tr("XACRO execution succeed") + "\n";
+                }
+                else
+                {
+                    report += "# " + tr("XACRO parsing failed") + "\n";
+                    report += "\n\n" + tr("Command called : \n'") + QString::fromUtf8(outcome.m_called.data()) + "'";
+                    report += "\n\n" + tr("Process failed with code : ") + QString::number(outcome.m_returnCode);
+                    report += "\n\n" + tr("error output") + " :\n\n";
+                    report += QString::fromLocal8Bit(outcome.m_logErrorOutput.data(), static_cast<int>(outcome.m_logErrorOutput.size()));
```
spham-amzn, line 100: nit: Should add an additional '\n'
```suggestion
                    report += QString::fromLocal8Bit(outcome.m_logErrorOutput.data(), static_cast<int>(outcome.m_logErrorOutput.size())) + "\n";
```

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
lumberyard-employee-dm: Or avoid the issue entirely and change the charBuffer type to an AZStd::string.
The `+1` on the length wouldn't be needed as the AZStd::string will set `charBuffer[length] = char{}` in  the [resize_no_construct](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/std/string/string.h#L1037-L1060) function

```
        AZStd::string charBuffer;
        charBuffer.resize_no_construct(length);
        fileStream.Read(length, charBuffer.data());
```

Or better yet, use the [resize_and_overwrite](https://en.cppreference.com/w/cpp/string/basic_string/resize_and_overwrite) function which will avoid the string from ever being in an invalid state

```suggestion
        AZStd::string charBuffer;
        auto ReadFile = [&fileStream](char* buffer, size_t size)
        {
            return systemFileStream.Read(size, buffer);
        };
        charBuffer.resize_and_overwrite(length, ReadFile);
```

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -312,4 +355,10 @@ namespace ROS2
         QMessageBox::critical(this, QObject::tr("Error"), errorMessage);
         AZ_Error("RobotImporterWidget", false, "%s", errorMessage.toUtf8().constData());
     }
+
+    bool RobotImporterWidget::IsFileXacro(const AZStd::string& filename) const
+    {
+        return filename.ends_with(".xacro");
```
spham-amzn, line 361: Linux is case sensitive, suggest capitalizing the filename and then checking against an all caps ".XACRO"

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
```diff
@@ -0,0 +1,80 @@
...
+        Utils::xacro::Params params;
+        const int rowCount = m_table->rowCount();
+        for (int i = 0; i < rowCount; i++)
+        {
+            const auto p1 = m_table->item(i, 0);
+            const auto p2 = m_table->item(i, 1);
+            AZ_Assert(p1, "cell should exist");
+            AZ_Assert(p2, "cell should exist");
+            AZStd::string name(p1->text().toUtf8());
+            AZStd::string value(p2->text().toUtf8());
```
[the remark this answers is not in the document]
lumberyard-employee-dm: > `name` and `value` can be const.

Also this needs to call `.constData` on the converted QByteArray returned from the `toUtf8` call.
This will cause an an implicit conversion warning on clang which will result in a compile error
```suggestion
            AZStd::string name(p1->text().toUtf8().constData());
            AZStd::string value(p2->text().toUtf8().constData());
```

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+        if (myProcess.waitForFinished() && myProcess.exitStatus() == QProcess::NormalExit && myProcess.exitCode() == 0)
+        {
+            AZ_Printf("ParseXacro", "xacro finished with success");
+            auto output = myProcess.readAllStandardOutput();
+            outcome.m_urdfHandle = UrdfParser::Parse(output.data());
+        }
+        else
+        {
+            AZ_Printf("ParseXacro", "xacro finished with code %d", myProcess.exitCode());
```
lumberyard-employee-dm, line 64: These printfs are missing newlines to make sure the next output is on a separate line.

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+        // test if xacro exists
+        AZStd::string xacroPath = WhichXacro();
+
+        AZ_Warning("ParseXacro", !xacroPath.empty(), "There is xacro executable in your path");
+        if (xacroPath.empty())
+        {
+            return ExecutionOutcome{};
+        }
+        AZ_Printf("ParseXacro", "xacro executable : %s", xacroPath.c_str());
+        AZ_Printf("ParseXacro", "Convert xacro file : %s", filename.c_str());
```
lumberyard-employee-dm, line 30: repeat: these printfs will all be on the same line.
A newline needs to be added.
```suggestion
        AZ_Printf("ParseXacro", "xacro executable : %s\n", xacroPath.c_str());
        AZ_Printf("ParseXacro", "Convert xacro file : %s\n", filename.c_str());
```

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+        QString called = program;
+
+        for (const auto& p : arguments)
+        {
+            called += " " + p;
+        }
+
+        outcome.m_called = AZStd::string(called.toUtf8().constData());
+        AZ_Printf("ParseXacro", "calling file : %s", outcome.m_called.c_str());
```
lumberyard-employee-dm, line 52: A newline is needed here.
```suggestion
        AZ_Printf("ParseXacro", "calling file : %s\n", outcome.m_called.c_str());
```

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+        const QString program{ "which" };
+        QStringList arguments;
+        arguments << kXacroExecutable;
+        QProcess myProcess;
+        myProcess.start(program, arguments);
+        if (myProcess.waitForFinished() && myProcess.exitStatus() == QProcess::NormalExit && myProcess.exitCode() == 0)
+        {
+            auto output = myProcess.readAllStandardOutput().trimmed();
+            return AZStd::string{ output.data(), static_cast<AZStd::string::size_type>(output.size()) };
+        }
```
spham-amzn, line 85: ```which``` is Linux specific. Even though this is Linux supported only, it still needs to gate this for just Linux. The suggested approach is to apply the PAL pattern for this behavior if possible, and asserting or erroring out on non-supported platforms until it becomes supported

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+        return AZStd::string{};
+    }
+
+    AZStd::unordered_map<AZStd::string, AZStd::string> GetParameterFromXacroData(AZStd::vector<char> data)
+    {
+        const char* kArgName = "xacro:arg";
+        const char* kName = "name";
+        const char* kDefault = "default";
```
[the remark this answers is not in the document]
spham-amzn: Also could be constexpr as well

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
lumberyard-employee-dm: I have a suggesstion of making the path to the `xacro` executable a setting within the Settings Registry

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "XacroUtils.h"
+#include <AzCore/IO/FileIO.h>
+#include <AzCore/XML/rapidxml.h>
+#include <QProcess>
+#include <QString>
```
lumberyard-employee-dm, line 13: There is actually no need to use Qt here.

The AzFramework library has a [ProcessWatcher](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Process/ProcessWatcher.h#L69-L70) class that can launch processes across all the support platforms

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.cpp
```diff
@@ -0,0 +1,80 @@
...
+        m_table->setSelectionBehavior(QAbstractItemView::SelectRows);
+        m_table->setHorizontalHeaderLabels({ tr("Parameter"), tr("Value") });
+        m_table->horizontalHeader()->setStretchLastSection(true);
+
+        QVBoxLayout* layout = new QVBoxLayout;
+        layout->addWidget(m_table);
+        this->setLayout(layout);
+    }
+
+    bool XacroParamsPage::isComplete() const
```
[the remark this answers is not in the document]
michalpelka: It is overridden from QWizardPage::isComplete, can't do.

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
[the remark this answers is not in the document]
michalpelka: Sorry about that!

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: Yes, it was rebased on top of PR that we already merged and squashed. I will rebase.

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
michalpelka: Unfortunately, rapid json needs a zero-terminated, mutable char array. I moved copy inside the parsing function.

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.cpp
michalpelka: Implemented @lumberyard-employee-dm suggestion. Thanks.

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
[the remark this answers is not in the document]
michalpelka: Thanks, fixed.

Gems/ROS2/Code/Source/RobotImporter/Pages/XacroParamsPage.h
michalpelka: Added line of documentation.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -33,7 +33,8 @@ namespace ROS2::Utils
 
     AZStd::unordered_map<AZ::Crc32, AvailableAsset> GetInterestingSourceAssetsCRC()
     {
-        const AZStd::unordered_set<AZStd::string> kInterestingExtensions{ ".dae", ".stl", ".obj" };
+        const AZStd::unordered_set<AZStd::string> kInterestingExtensions{ ".dae", ".stl", ".obj", ".fbx" };
+        constexpr char AzModelExtension[] {".azmodel"} ;
```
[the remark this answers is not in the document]
michalpelka: Code was removed.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
hultonha: Okay great thanks :+1:

Gems/ROS2/Code/Source/RobotImporter/xacro/XacroUtils.h
hultonha: Could you make a PR for this update to the C++ best practices/coding standards @lumberyard-employee-dm? 🙂

hultonha (approved):

moraaar (approved):

spham-amzn (approved):

lemonade-dm (approved):

spham-amzn:
@michalpelka I was testing this PR locally and noticed that I had to install an additional ROS package:
```
sudo apt install ros-$ROS_DISTRO-xacro
```
Can we add that to the README.md for this Gem as part of this PR?

michalpelka:
@spham-amzn updated the readme with the package you pointed out. Thanks!

hultonha:
Just started another build 👍

adamdbrw:
Seems that all is in order here, could this be merged?

hultonha:
> Seems that all is in order here, could this be merged?

Merged 👍

PR #105, opened 2023-02-01
Add sensor profiling to ROS2 Gem

michalpelka:
This change allows to display and benchmark update times in simulated ROS2 sensors.
image


Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponent.h
```diff
@@ -54,5 +63,13 @@ namespace ROS2
         virtual void Visualise(){};
 
         float m_timeElapsedSinceLastTick = 0.0f;
+        double m_timeElapsedSinceLastFrequencyTick = 0.0f;
```
adamdbrw, line 66: Please separate this out to a class that ROS2SensorComponent does not depend on (it is an addition).

adamdbrw (changes requested):

PR #119, opened 2023-02-06
Adjustment automatically ROS2FrameComponent to publish static or dynamic transform

michalpelka:
Adjustment automatically ROS2FrameComponent to publish static or dynamic transform.
It deduces from existing components what if the transform is static.

Signed-off-by: Michał Pełka <michal.pelka@robotec.ai>

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Frame/ROS2FrameComponent.h
+++ b/Gems/ROS2/Code/Include/ROS2/Frame/ROS2FrameComponent.h
@@ -86,6 +86,7 @@ namespace ROS2
         AZStd::string m_frameName = "sensor_frame";
 
         bool m_publishTransform = true;
+        bool m_isDynamic = false;
         AZStd::unique_ptr<ROS2Transform> m_ros2Transform;
     };
 } // namespace ROS2
--- a/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
+++ b/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
@@ -14,6 +14,7 @@
 #include <ROS2/ROS2Bus.h>
 #include <ROS2/ROS2GemUtilities.h>
 #include <ROS2/Utilities/ROS2Names.h>
+
 namespace ROS2
 {
     namespace Internal
@@ -57,6 +58,23 @@ namespace ROS2
             // Found the component!
             return component;
         }
+
+        //! Returns if entity has component of given type
+        //! @param entity pointer to entity
+        //! @param type tpye of the component
+        //! @returns true if entity has component with given type
+        bool CheckIfEntityHasComponentOfType(const AZ::Entity* entity, const AZ::Uuid type)
+        {
+            for (const auto& c : entity->GetComponents())
+            {
+                if (c->RTTI_IsTypeOf(type))
+                {
+                    return true;
+                }
+            }
+            return false;
+        }
+
     } // namespace Internal
 
     void ROS2FrameComponent::Activate()
@@ -65,9 +83,27 @@ namespace ROS2
 
         if (m_publishTransform)
         {
+            AZ_TracePrintf("ROS2FrameComponent", "Setting up %s", GetFrameID().data());
+
+            // The frame is static when is not a top entity and one of the following is true:
+            //    - if the entity has at least one Fixed Joint
+            //    - if the entity has no joints at all.
+            if (IsTopLevel())
+            {
+                m_isDynamic = true;
+            }
+            else
+            {
+                bool hasJoints = Internal::CheckIfEntityHasComponentOfType(
+                    m_entity, AZ::Uuid("{B01FD1D2-1D91-438D-874A-BF5EB7E919A8}")); // Physx::JointComponent;
+                bool hasFixedJoints = Internal::CheckIfEntityHasComponentOfType(
+                    m_entity, AZ::Uuid("{02E6C633-8F44-4CEE-AE94-DCB06DE36422}")); // Physx::FixedJointComponent
+                m_isDynamic = hasJoints && !hasFixedJoints;
+            }
+
             AZ_TracePrintf(
                 "ROS2FrameComponent",
-                "Setting up %s transfrom between parent %s and child %s to be published %s\n",
+                "Setting up %s transform between parent %s and child %s to be published %s\n",
                 IsDynamic() ? "dynamic" : "static",
                 GetParentFrameID().data(),
                 GetFrameID().data(),
@@ -114,7 +150,7 @@ namespace ROS2
 
     bool ROS2FrameComponent::IsDynamic() const
     {
-        return IsTopLevel();
+        return m_isDynamic;
     }
 
     const ROS2FrameComponent* ROS2FrameComponent::GetParentROS2FrameComponent() const
--- a/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
+++ b/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
@@ -7,6 +7,7 @@
  */
 
 #include <AzCore/Component/Entity.h>
+#include <AzCore/Component/EntityUtils.h>
 #include <AzCore/Serialization/EditContext.h>
 #include <AzCore/Serialization/EditContextConstants.inl>
 #include <AzCore/Serialization/SerializeContext.h>
@@ -59,20 +60,14 @@ namespace ROS2
             return component;
         }
 
-        //! Returns if entity has component of given type
+        //! Checks whether the entity has a component of the given type
         //! @param entity pointer to entity
-        //! @param type tpye of the component
+        //! @param typeId type of the component
         //! @returns true if entity has component with given type
-        bool CheckIfEntityHasComponentOfType(const AZ::Entity* entity, const AZ::Uuid type)
+        static bool CheckIfEntityHasComponentOfType(const AZ::Entity* entity, const AZ::Uuid typeId)
         {
-            for (const auto& c : entity->GetComponents())
-            {
-                if (c->RTTI_IsTypeOf(type))
-                {
-                    return true;
-                }
-            }
-            return false;
+            auto components = AZ::EntityUtils::FindDerivedComponents(entity, typeId);
+            return !components.empty();
         }
 
     } // namespace Internal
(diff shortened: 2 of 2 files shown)


michalpelka:
Addresses : https://github.com/RobotecAI/o3de-ros2-gem/issues/13

Gems/ROS2/Code/Include/ROS2/Frame/ROS2FrameComponent.h
```diff
@@ -50,6 +50,10 @@ namespace ROS2
         //! Set a above-mentioned frame id
         void SetFrameID(const AZStd::string& frameId);
 
+        //! Set is Frame is dynamic or static
+        //! @param isDynamic true for dynamic frame, that should be published to `/tf`
+        void SetFrameDynamic(const bool isDynamic);
```
adamdbrw, line 55: Quick look: it seems that we only want to set this if there are joints (other than Fixed joint) attached. Is there any need to expose this to the user?

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -73,6 +73,29 @@ namespace ROS2
                 GetFrameID().data(),
                 IsDynamic() ? "continuously to /tf" : "once to /tf_static");
 
+            if (!IsDynamic())
+            {
+                bool hasMovableJoints = false;
+                const auto& components = this->m_entity->GetComponents();
+                for (const auto component : components)
```
adamdbrw, line 80: This could be checked on activation and used to set m_isDynamic

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -73,6 +73,29 @@ namespace ROS2
...
+                {
+                    const auto type = component->GetUnderlyingComponentType();
+
+                    if (type == AZ::Uuid("{9C5B955C-6C80-45FA-855D-DDA449C85313}") || // PhysXHingeJoint
```
adamdbrw, line 84: Would it be easier if we considered any ROS2Frame without, specifically, a FixedJoint to be a dynamic frame?
It does look correct and easier to maintain (consider new joint types being added).

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -73,6 +73,29 @@ namespace ROS2
...
+                        type == AZ::Uuid("{CEE6A6DF-FDE1-4E30-9EE2-631C7561C1C7}") || // PhysXPrismaticJoint
+                        type == AZ::Uuid("{A5CA0031-72E4-4908-B764-EDECD3091882}") || // HingeJointComponent
+                        type == AZ::Uuid("{9B34CA1B-C063-4D42-A15B-CE6CD7C828DC}")) // PrismaticJointComponent
+                    {
+                        hasMovableJoints = true;
+                        break;
+                    }
+                }
+                AZ_Warning(
```
adamdbrw, line 93: This warning could be removed if the value m_isDynamic is simply set by the logic and not settable by the user

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -114,7 +137,7 @@ namespace ROS2
 
     bool ROS2FrameComponent::IsDynamic() const
     {
-        return IsTopLevel();
+        return m_isDynamic;
```
adamdbrw, line 140: Nice. If we consider 
Dynamic = not having a fixed joint
then the top level logic is also retained.

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -265,6 +265,11 @@ namespace ROS2
             auto* component = Utils::GetGameOrEditorComponent<ROS2FrameComponent>(entity);
             AZ_Assert(component, "ROS2 Frame Component does not exist for %s", entityId.ToString().c_str());
             component->SetFrameID(AZStd::string(link->name.c_str(), link->name.size()));
+            // non-static transforms are applied to movable link and root.
+            if (!link->parent_joint || Utils::IsMovableLink(link))
```
adamdbrw, line 269: this would not be needed if we set it in ROS2Frame Activate instead and would be encapsulated.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -46,6 +46,27 @@ namespace ROS2
         return false;
     }
 
+    bool Utils::IsMovableLink(const urdf::LinkConstSharedPtr& link)
+    {
+        // Parent joint needs to be CONTINOUS, REVOLUTE, PRISMATIC
+        if (link->parent_joint)
+        {
+            if (link->parent_joint->type == urdf::Joint::CONTINUOUS)
```
adamdbrw, line 54: !Fixed

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -33,6 +33,12 @@ namespace ROS2
         //! @return true if the link is likely a wheel link.
         bool IsWheelURDFHeuristics(const urdf::LinkConstSharedPtr& link);
 
+        //! Determine whether a given link is likely movable and should has non-static transform.
```
adamdbrw, line 36: should have a dynamic transform

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -33,6 +33,12 @@ namespace ROS2
...
+        //! This can be useful to provide a good default behavior - set
+        //! @param link the link that will be subjected to the heuristic.
```
adamdbrw, line 38: what is the heuristic? 
"The link to be checked"?

michalpelka:
Made major revision of code.

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -57,6 +58,23 @@ namespace ROS2
             // Found the component!
             return component;
         }
+
+        //! Returns if entity has component of given type
+        //! @param entity pointer to entity
+        //! @param type tpye of the component
+        //! @returns true if entity has component with given type
```
adamdbrw, line 65: Minor adjustment in wording:
First sentence: "Check whether the entity has a component of the given type"

Typo in "tpye"

adamdbrw (approved):

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -57,6 +58,23 @@ namespace ROS2
...
+        bool CheckIfEntityHasComponentOfType(const AZ::Entity* entity, const AZ::Uuid type)
```
adamdbrw, line 66: This looks like a candidate for EntityUtils.h 

There is another function to consider here:
        template<class ComponentType>
        inline AZStd::vector<ComponentType*> FindDerivedComponents(Entity* entity)

michalpelka: Thanks for tip! Used FindDerivedComponents.

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -57,6 +59,17 @@ namespace ROS2
...
         }
+
+        //! Check whether the entity has a component of the given type
+        //! @param entity pointer to entity
+        //! @param typeId type of the component
+        //! @returns true if entity has component with given type
+        bool CheckIfEntityHasComponentOfType(const AZ::Entity* entity, const AZ::Uuid typeId)
+        {
+            auto components = AZ::EntityUtils::FindDerivedComponents(entity, typeId);
+            return !components.empty();
```
hultonha, line 70: question: Just wanted to sanity check does this check the whole entity hierarchy or just the entity being passed in? Might be good to clarify just in case.

Also this is super minor but for API comments we general phrase them in such a way that it describes what it does not what it's doing (as opposed to an implementation comment). So this API comment above would simply read `Checks whether the entity has a component of the given type.` (adding one 's' 😅 I know, but in my mind it does make a world of difference - can totally do this in another PR though, I'm happy to start a build now).

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -57,6 +59,17 @@ namespace ROS2
             // Found the component!
             return component;
         }
+
+        //! Check whether the entity has a component of the given type
+        //! @param entity pointer to entity
+        //! @param typeId type of the component
+        //! @returns true if entity has component with given type
+        bool CheckIfEntityHasComponentOfType(const AZ::Entity* entity, const AZ::Uuid typeId)
```
hultonha, line 67: nit: As this function is local to the .cpp file we can mark it `static` to show it has internal linkage (this should make it possible to inline and will compile slightly faster hopefully! 😅)

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -65,9 +78,27 @@ namespace ROS2
...
+            //    - if the entity has no joints at all.
+            if (IsTopLevel())
+            {
+                m_isDynamic = true;
+            }
+            else
+            {
+                bool hasJoints = Internal::CheckIfEntityHasComponentOfType(
+                    m_entity, AZ::Uuid("{B01FD1D2-1D91-438D-874A-BF5EB7E919A8}")); // Physx::JointComponent;
+                bool hasFixedJoints = Internal::CheckIfEntityHasComponentOfType(
```
hultonha, line 94: nit: This is super minor but we general try and mark everything that doesn't change `const`, so these could have `const` in front of them

hultonha (approved):
Change looks good! I just have some minor feedback, happy to leave for now. Let me know when you're happy and I can kick off a build 👍

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -65,9 +78,27 @@ namespace ROS2
 
         if (m_publishTransform)
         {
+            AZ_TracePrintf("ROS2FrameComponent", "Setting up %s", GetFrameID().data());
+
+            // The frame is static when is not a top entity and one of the following is true:
+            //    - if the entity has at least one Fixed Joint
+            //    - if the entity has no joints at all.
```
moraaar, line 85: nit: this is super minor, but this comment is describing the inversed condition of what the code is doing, making it hard to follow. I'd recommend describing the the positive condition rather than the inversed. Example:

````
            // The frame will always be dynamic if it's a top entity.
            if (IsTopLevel())
            {
                m_isDynamic = true;
            }
            // Otherwise it'll be dynamic when it has joints and and it's not a fixed joint.
            else
            {
                bool hasJoints = Internal::CheckIfEntityHasComponentOfType(
                    m_entity, AZ::Uuid("{B01FD1D2-1D91-438D-874A-BF5EB7E919A8}")); // Physx::JointComponent;
                bool hasFixedJoints = Internal::CheckIfEntityHasComponentOfType(
                    m_entity, AZ::Uuid("{02E6C633-8F44-4CEE-AE94-DCB06DE36422}")); // Physx::FixedJointComponent
                m_isDynamic = hasJoints && !hasFixedJoints;
            }
````

moraaar (approved):

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
michalpelka: Sanity: I am checking components belonging to one entity only.
API Comment: totally agree!

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
michalpelka: Looks great now! thank you for your suggestion!

hultonha:
Just kicked off a build 👍 https://jenkins.build.o3de.org/view/Projects/job/o3de-extras/view/change-requests/job/PR-119/1/

PR #177, opened 2023-02-16
A fix for ROS2Frame component

### LINKED ISSUE #139: TF Frame translations value are invalid when the `FrameComponent` chain is broken in an object hierarchy
**Bug description**
TF Frame translation values are invalid when the chain of `FrameComponent`s is broken within the single hierarchy.

If there is a prefab inside another prefab, and both of them contain a `FrameComponent`s, the virtual container object between them breaks the translation between the components.

For example, if the robot prefab has the `base_link` frame and the sensor prefab inside the robot got `sensor` frame and is located in some (x,y,z) position, then the returned `base_link -> sensor` tf frame is (0,0,0) rather than (x,y,z).

**To Reproduce**
1. Create an empty prefab, lets call it vehicle
2. Add `FrameComponent` to that prefab - set the frame_id to `base_link`
3. Create another empty prefab, lets call it sensor
4. Add `FrameComponent`  to the sensor prefab, set frame id to `sensor`
5. Insert sensor prefab into vehicle prefab
6. Add a local translation for sensor prefab, ie (1,0,0)
7. Run the scene
8. See the TF tree

**Expected behavior**
Tf `sensor->base_link` is (1,0,0).

**Actual behavior**
Tf `sensor->base_link` is (0,0,0).

**Additional context**
It is possible to hack and fix it by adding `FrameComponent` to the virtual container object - but that should not be possible anyway.

adamdbrw:
- Previously, it did not work well with intermediate entities without ROS2Frame component.
- For example, the prefab container entity.
- This resulted in skipping prefab transform and publishing invalid /tf

Fixes #139

### DIFF
--- a/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
+++ b/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
@@ -161,8 +161,8 @@ namespace ROS2
             auto* ancestorTransformInterface = Internal::GetEntityTransformInterface(parentFrame->GetEntity());
             AZ_Assert(ancestorTransformInterface, "No transform interface for an entity with a ROS2Frame component, which requires it!");
 
-            auto ancestorWorldTm = ancestorTransformInterface->GetWorldTM();
-            auto thisWorldTm = transformInterface->GetWorldTM();
+            const auto ancestorWorldTm = ancestorTransformInterface->GetWorldTM();
+            const auto thisWorldTm = transformInterface->GetWorldTM();
             return ancestorWorldTm.GetInverse() * thisWorldTm;
         }
         return transformInterface->GetWorldTM();
--- a/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
+++ b/Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
@@ -50,6 +50,7 @@ namespace ROS2
             AZ::Entity* parentEntity = nullptr;
             AZ::ComponentApplicationBus::BroadcastResult(parentEntity, &AZ::ComponentApplicationRequests::FindEntity, parentEntityId);
             AZ_Assert(parentEntity, "No parent entity id : %s", parentEntityId.ToString().c_str());
+
             auto* component = Utils::GetGameOrEditorComponent<ROS2FrameComponent>(parentEntity);
             if (component == nullptr)
             { // Parent entity has no ROS2Frame, but there can still be a ROS2Frame in its ancestors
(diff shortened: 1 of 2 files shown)


adamdbrw:
Validated correctness with a few examples, one pictured here: 
image (28)

spham-amzn (approved):

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -152,12 +153,17 @@ namespace ROS2
...
         auto* transformInterface = Internal::GetEntityTransformInterface(GetEntity());
-        if (GetParentROS2FrameComponent() != nullptr)
+        if (const auto* parentFrame = GetParentROS2FrameComponent(); parentFrame != nullptr)
         {
-            return transformInterface->GetLocalTM();
+            auto* ancestorTransformInterface = Internal::GetEntityTransformInterface(parentFrame->GetEntity());
+            AZ_Assert(ancestorTransformInterface, "No transform interface for an entity with a ROS2Frame component, which requires it!");
+
+            auto ancestorWorldTm = ancestorTransformInterface->GetWorldTM();
+            auto thisWorldTm = transformInterface->GetWorldTM();
```
moraaar, line 165: `ancestorWorldTm` and `thisWorldTm` can be const.

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -152,12 +153,17 @@ namespace ROS2
...
+            return ancestorWorldTm.GetInverse() * thisWorldTm;
```
hultonha, line 166: suggestion: For these types of transforms I came across a wonderful article a few years ago with a great recommendation on naming variables for matrix transforms - https://www.sebastiansylvan.com/post/matrix_naming_convention/.

The general idea is to have the space encoded in the variable name, and have them match up when multiplying, so we could rename things below like so....


```
worldFromAncestor
worldFromThis

worldFromAncestor = ancestorFromWorld.GetInverse();
return ancestorFromWorld * worldFromThis; // here we can see the 'world' parts of the variables are next to each other, confirming this is the right multiplication order
```

Totally optional, feel free to ignore, but I always like mentioning it as I found it incredibly useful and tend to use that style where I can now.

moraaar (approved):

hultonha (approved):
Change looks good to me! My only suggestion would be (time permitting) this could be a useful bit of functionality to add a test for to verify the different conditions. It might be a little tricky to setup as-is but we do have some examples of how to do this in O3DE (or we could extract some of the logic and test in isolation). Not blocking but just wanted to mention it maybe as a follow-up. Thanks!

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
adamdbrw: This increases clarity, thank you for such a good suggestion, I will apply it!

adamdbrw: I agree we could use tests for this feature. I would like to think about adding more Unit tests and map out the work for the Gem in general. I have created an issue to track unit tests for the Frame component specifically: https://github.com/o3de/o3de-extras/issues/178. For new features, we will be adding unit tests when possible.

hultonha (approved):
Thanks for making the changes, looks great! I'll start another build 👍

michalpelka (approved):
Tested manually against Panda robot, works as intended.

PR #179, opened 2023-02-20
Pid refactoring

danielemorra98:
Moving the location of `PIDConfiguration.*` from `VehicleDynamics/DriveModels` to `Utilities/Controllers` so that it could be reused by other components without referring to VehicleDynamics dependency.


danielemorra98:
@michalpelka I am fighting against some proxy limits which don't allow me to push on LFS (it happens all the time I try to rebase to respect the DCO check). I opened another PR hoping that I would have solved it. It did not happen.
Moreover I cannot see why in the files changed I continue to see files that I haven't touched at all. I directed this PR against the newest commit of development branch

Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/AckermannDriveModel.cpp
```diff
@@ -21,7 +21,7 @@ namespace ROS2::VehicleDynamics
     void AckermannDriveModel::Reflect(AZ::ReflectContext* context)
     {
         AckermannModelLimits::Reflect(context);
-        PidConfiguration::Reflect(context);
+        ROS2::Controllers::PidConfiguration::Reflect(context);
```
adamdbrw, line 24: There is a decision where to reflect a configuration structure which is used in more than one other component that needs it for its reflection. Considering that the choice would be arbitrary and that some components could be refactored outside to another Gem (VehicleDynamics possibly could), it is arguably better to reflect such common configurations in ROS2SystemComponents. For example, TopicConfiguration and QoS are reflected in the system component already.

Gems/ROS2/Code/ros2_header_files.cmake
```diff
@@ -19,8 +19,8 @@ set(FILES
         Include/ROS2/Sensor/ROS2SensorComponent.h
         Include/ROS2/Sensor/SensorConfiguration.h
         Include/ROS2/Spawner/SpawnerBus.h
+        Include/ROS2/Utilities/Controllers/PidConfiguration.h
         Include/ROS2/Utilities/ROS2Conversions.h
         Include/ROS2/Utilities/ROS2Names.h
-        Include/ROS2/VehicleDynamics/DriveModels/PidConfiguration.h
```
adamdbrw, line 24: This change possibly breaks apple orchard demo with development. However, since we decided to support exact commit hashes and upgrade gradually, this is ok.

adamdbrw (approved):
The change set looks good! Please address my comment about reflection.

michalpelka (commented):
Looks good, I would suggest squashing old commits into one.
I can do it for you if you want.

adamdbrw (approved):

danielemorra98:
> Looks good, I would suggest squashing old commits into one. I can do it for you if you'd like.

Should it be done automatically by the merging, right?

PR #204, opened 2023-03-06
Change how ROS 2 packages are included.

michalpelka:
Include cmake configs from the ROS 2 package instead of building lib and include paths.
Added function `target_depends_on_ros2_package` that allows specifying version and optionality of the ROS 2 package.
Requested minimum version of control_toolbox to be 2.2.0.
Addresses issues :
#198

Issue #155, opened 2022-05-11
IMU Sensor improvements


mdrwiega:
One of the possible improvements to IMU sensor is to provide better accuracy of angular velocity and linear acceleration calculations.
To do that, there all calculations (especially on quaternions) should be done on double-precision representation (currently single-precision is used).

Steps:
- provide Quaternion class on the double with basic operations
- provide basic operations on Vector3 with double precision
- replace existing Quaternion class and operations on Vector3 in the current IMU implementation

michalpelka:
There is more work to be done at IMU.
# Gyroscope
Measures the current rate of rotation. It is a 3D vector.
Let us assume that we have poses of the body under generalized motion in two consecutive frames $t_1$ and $t_2$.
In time $t_1$ the body has pose $P_1$ and in time $t_2$ body has pose $P_2$. 
Assuming the scale is constant we can write that  $P_1$ and $P_2$ are orthonormal matrices.
To find rotation vector $\omega$ we can take Log to obtain rate:
$\omega = \frac{Log(P_1^{-1}P_2)}{t_2-t_1}$.
Note that Log is a logarithm that takes the matrix here and produce a vector, more on that operator, which is called the tangential map of Lie group can be found in this excellent [paper](https://arxiv.org/pdf/1812.01537v4.pdf).

# Accelerometers
Accelerometer is a device that has a proof mass that is displaced by force that acts upon this mass.
This mass is affected by the sum of accelerations:
 - change of linear speed  $a_v$
 - centripetal acceleration $a_c$
 - gravity acceleration $a_g$
 
 ## Change of linear speed
Let us assume that we have poses of body under generalized motion in two consecutive frames $t_1$ and $t_2$.
In time $t_1$ the body has pose $P_1$ and in time $t_2$ body has pose $P_2$.
 $a_v = \frac{ translation(P_2) - translation(P_1)}{t_2-t_1}$
 
 ## Gravity
 It is a constant vector pointing in a negative Z direction.
![CodeCogsEqn](https://user-images.githubusercontent.com/3209244/219658861-22f414d3-0e37-4cdf-abf9-44b06fb0aa50.gif)

 ## centripetal acceleration $a_c$ 
centripetal acceleration is a result of a change in direction of movement. 
It is a vector cross product of rotation speed $\omega$ and radius vector $r$. 
The radius vector has a direction pointing from a rotation center toward the moving body.
[...]

PR #206, opened 2023-03-06
High frequency implementation of odometry sensor.

michalpelka:
- [ ] Change the ROS2SensorComponent interface, and remove its ticking loop (e.g. add another base class in between). A ticking loop is not needed, wait for the solution proposed in #155.
- [ ] Add a mechanism to limit output the frequency
- [ ] Profile `m_onSceneActiveSimulatedBodiesEvent` - it performs expensive and redundant 'GetFrameTransform', it can be delegated to TickBus::Handler
- [ ] Rebase and clean unblocking code (covered by #200 and #204)
- [ ] Remove draft status


michalpelka:
FYI @wojtek-robotecai

Issue #118, opened 2023-02-06
O3DE editor crashes on URDF load.


michalpelka:
The o3de editor is terminated on the second page of the URDF import wizard.

**Version:**
 - o3de :  5eec62c40bcae2d440c9e58ceb61c3747301f0fc
 - o3de-extras : 930806fb3b1b22b504d9b8f4816881e05c8cd656
 
 **Asset**
 https://github.com/husarion/rosbot_xl_ros/blob/master/rosbot_xl_description/urdf/rosbot_xl.urdf.xacro
 
**Call stack**
The process crashes during conversion xacro to urdf in `WaitForReadyOutputs`.
```
__pthread_kill_implementation 0x00007ffff6d62a7c
__pthread_kill_internal 0x00007ffff6d62a7c
__GI___pthread_kill 0x00007ffff6d62a7c
__GI_raise 0x00007ffff6d0e476
__GI_abort 0x00007ffff6cf47f3
__libc_message 0x00007ffff6d556f6
__GI___fortify_fail 0x00007ffff6e0276a
__GI___chk_fail 0x00007ffff6e010c6
__fdelt_chk 0x00007ffff6e026ab
AzFramework::StdInOutProcessCommunicator::WaitForReadyOutputs ProcessCommunicator_Linux.cpp:215
AzFramework::ProcessCommunicator::ReadIntoProcessOutput ProcessCommunicator.cpp:53
AzFramework::ProcessWatcher::LaunchProcessAndRetrieveOutput ProcessWatcher.cpp:38
ROS2::Utils::xacro::ParseXacro XacroUtils.cpp:58
ROS2::RobotImporterWidget::OpenUrdf RobotImporterWidget.cpp:88
ROS2::RobotImporterWidget::validateCurrentPage RobotImporterWidget.cpp:240
QWizard::next 0x00007fffe7f6bf8a
doActivate<false> 0x00007ffff7c82b50
QMetaObject::activate 0x00007ffff7c7bee7
QAbstractButton::clicked 0x00007fffe7e137e6
QAbstractButtonPrivate::emitClicked 0x00007fffe7e13a0e
QAbstractButtonPrivate::click 0x00007fffe7e14ee3
QAbstractButton::mouseReleaseEvent 0x00007fffe7e150a5
QWidget::event 0x00007fffe7d5dfd6
QApplicationPrivate::notify_helper 0x00007fffe7d1ae03
QApplication::notify 0x00007fffe7d23f17
[...]

PR #222, opened 2023-03-17
Change way of communication with ROS2 Xacro.

michalpelka:
It is PR that is used to debug #118. 
It seems that the code here is going around some bug in ProcessWatcher.


michalpelka:
Close since this code is not fixing anything. The problem was in o3de :
https://github.com/aws-lumberyard-dev/o3de/commit/9f07d23eedd58bc2f6bb1a6c1d94d735bac4daaa

Issue #225, opened 2023-03-22
The ROS2 Urdf importer resolves packages incorrectly


michalpelka:
URDF/XACRO can contain such syntax:
```
        <geometry>
          <mesh filename="package://husky_description/meshes/base_link.dae" />
        </geometry>
```
It is resolved incorrectly :
```
/home/michal/ros2_workspace/husky/husky_description/husky_description/base_link.dae
```
That causes these meshes cannot be loaded.
![image](https://user-images.githubusercontent.com/3209244/226902623-e489346a-29cf-473f-aaca-ce27861b9838.png)

michalpelka:
Beinging fixed #226

michalpelka:
Fixed in #226

PR #226, opened 2023-03-22
ROS2 URDF importer will parse AMENT_PREFIX_PATH

michalpelka:
AMENT_PREFIX_PATH contains installed packages. URDF can refer to it as :
`package://example_package`. This change would provide a way to parse it correctly.

As alternative this API can be used: [https://github.com/ament/ament_index/blob/d9f7f2783dedc5687a5bdaf3ce5252dce5c073c5[…]dex_cpp/include/ament_index_cpp/get_package_share_directory.hpp](https://github.com/ament/ament_index/blob/d9f7f2783dedc5687a5bdaf3ce5252dce5c073c5/ament_index_cpp/include/ament_index_cpp/get_package_share_directory.hpp#L25-L33)
I decided to parse AMENT_PREFIX_PATH directly - ament_index_cpp uses exceptions.
Solves #225 
Tested using attached unit tests.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
@@ -148,21 +148,55 @@ namespace ROS2
 
     /// Finds global path from URDF path
     AZStd::string Utils::ResolveURDFPath(
-        AZStd::string unresolvedPath, const AZStd::string& urdfFilePath, const AZStd::function<bool(const AZStd::string&)>& fileExists)
+        AZStd::string unresolvedPath,
+        const AZStd::string& urdfFilePath,
+        const AZStd::string& amentPrefixPath,
+        const AZStd::function<bool(const AZStd::string&)>& fileExists)
     {
+        AZ_Printf("ResolveURDFPath", "ResolveURDFPath with %s\n", unresolvedPath.c_str());
         if (unresolvedPath.starts_with("package://"))
         {
             AZ::StringFunc::Replace(unresolvedPath, "package://", "", true, true);
+            AZ::IO::Path unresolvedProperPath(unresolvedPath);
+
+            // check existing path in Ament
+            if (!unresolvedProperPath.empty())
+            {
+                const AZStd::string packageNameCandidate = unresolvedProperPath.begin()->String();
+                AZStd::vector<AZStd::string> packages;
+                AZ::StringFunc::Tokenize(amentPrefixPath, packages, ':');
+                for (const auto& package : packages)
+                {
+                    if (package.ends_with(packageNameCandidate))
+                    {
+                        const AZ::IO::Path packagePath = AZ::IO::Path{ package } / "share";
+                        AZ::IO::Path packageXmlCandite = packagePath / AZ::IO::Path{ packageNameCandidate } / "package.xml";
+                        // check if it is Ament package
+                        if (fileExists(packageXmlCandite.String()))
+                        {
+                            const AZ::IO::Path resolvedPath = packagePath / unresolvedPath;
+                            AZ_Printf("ResolveURDFPath", "Resolved to using Ament to : %s\n", resolvedPath.String().data());
+                            // package.xml has been found
+                            return resolvedPath.String();
+                        }
+                        break;
+                    }
+                }
+            }
+
             AZ::IO::Path urdfProperPath(urdfFilePath);
             AZ::IO::Path packagePath;
             for (auto it = urdfProperPath.begin(); it != urdfProperPath.end(); it++)
             {
                 packagePath /= *it;
                 AZStd::string packageXmlCandite = (packagePath / "package.xml").String();
+                // check if it is Ament package
                 if (fileExists(packageXmlCandite))
                 {
+                    const AZ::IO::Path resolvedPath = packagePath / unresolvedPath;
+                    AZ_Printf("ResolveURDFPath", "ResolveURDFPath with relative path to : %s\n", unresolvedPath.c_str());
                     // package.xml has been found
-                    return (packagePath / unresolvedPath).String();
+                    return resolvedPath.String();
                 }
             }
             // No path available
@@ -172,13 +206,16 @@ namespace ROS2
         {
             // Paths that start with 'file:///' are absolute paths
             AZ::StringFunc::Replace(unresolvedPath, "file://", "", true, true);
+            AZ_Printf("ResolveURDFPath", "ResolveURDFPath with absolute path to : %s\n", unresolvedPath.c_str());
             return unresolvedPath;
         }
         // seems to be relative path
         AZ::IO::Path relativePath(unresolvedPath);
         AZ::IO::Path urdfProperPath(urdfFilePath);
         AZ::IO::Path urdfParentPath = urdfProperPath.ParentPath();
-        return (urdfParentPath / relativePath).String();
+        const AZ::IO::Path resolvedPath = urdfParentPath / relativePath;
+        AZ_Printf("ResolveURDFPath", "ResolveURDFPath with relative path to : %s\n", unresolvedPath.c_str());
+        return resolvedPath.String();
     }
 
 } // namespace ROS2
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
@@ -146,6 +146,36 @@ namespace ROS2
         return filenames;
     }
 
+    AZStd::optional<AZ::IO::Path> GetResolvedPath(const AZ::IO::Path &packagePath,
+                                                 const AZ::IO::Path &unresolvedPath,
+                                                 const AZStd::function<bool(const AZStd::string&)>& fileExists)
+    {
+        AZ::IO::Path packageXmlCandite = packagePath / "package.xml";
+        if (fileExists(packageXmlCandite.String()))
+        {
+            AZ::IO::Path resolvedPath = packagePath / unresolvedPath;
+            if (fileExists(resolvedPath.String()))
+            {
+                return AZStd::optional<AZ::IO::Path>{resolvedPath};
+            }
+        }
+        return AZStd::optional<AZ::IO::Path>{};
+    }
+
+    AZ::IO::Path GetPathFromSubPath(const AZ::IO::Path::const_iterator& begin, const AZ::IO::Path::const_iterator& end)
+    {
+        AZ::IO::Path subpath;
+        if (begin == end)
+        {
+            return subpath;
+        }
+        for (AZ::IO::Path::iterator pathIt = begin;pathIt!=end;pathIt++)
+        {
+            subpath /= *pathIt;
+        }
+        return subpath;
+    }
+
     /// Finds global path from URDF path
     AZStd::string Utils::ResolveURDFPath(
         AZStd::string unresolvedPath,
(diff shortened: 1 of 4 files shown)


Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -172,13 +206,16 @@ namespace ROS2
...
             AZ::StringFunc::Replace(unresolvedPath, "file://", "", true, true);
+            AZ_Printf("ResolveURDFPath", "ResolveURDFPath with absolute path to : %s\n", unresolvedPath.c_str());
             return unresolvedPath;
         }
         // seems to be relative path
         AZ::IO::Path relativePath(unresolvedPath);
         AZ::IO::Path urdfProperPath(urdfFilePath);
         AZ::IO::Path urdfParentPath = urdfProperPath.ParentPath();
-        return (urdfParentPath / relativePath).String();
+        const AZ::IO::Path resolvedPath = (urdfParentPath / relativePath);
```
hultonha, line 216: uber nit: It looks like the parens here are now redundant and could be removed?

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -66,6 +66,7 @@ namespace ROS2
         AZStd::string ResolveURDFPath(
             AZStd::string unresolvedPath,
             const AZStd::string& urdfFilePath,
+            const AZStd::string& amentPrefixPath,
```
hultonha, line 69: update: It would be great to update the API comment above to include the new `amentPrefixPath` parameter

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -103,12 +103,16 @@ namespace ROS2::Utils
 
     UrdfAssetMap FindAssetsForUrdf(const AZStd::unordered_set<AZStd::string>& meshesFilenames, const AZStd::string& urdFilename)
     {
+        auto enviromentalVariable = std::getenv("AMENT_PREFIX_PATH");
```
hultonha, line 106: question: Do we have an `AZStd` equivalent of this? 🤔 May well be fine but it might be useful to add an alias in future.

e.g.
```
namespace AZStd
{
    using std::env;
}
```

In a more central location (we might be able to follow-up with @o3de/sig-core on this)

hultonha (approved):
Changes look good to me and great work adding a new test too! 🥳 I only had very minor feedback, I can kick off build once you've responded. Thanks!

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: I am leaving here std::env for now.

SergeyAMZN (approved):

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -148,21 +148,55 @@ namespace ROS2
...
+                        AZ::IO::Path packageXmlCandite = packagePath / AZ::IO::Path{ packageNameCandidate } / "package.xml";
+                        // check if it is Ament package
+                        if (fileExists(packageXmlCandite.String()))
+                        {
+                            const AZ::IO::Path resolvedPath = packagePath / unresolvedPath;
+                            AZ_Printf("ResolveURDFPath", "Resolved to using Ament to : %s\n", resolvedPath.String().data());
+                            // package.xml has been found
+                            return resolvedPath.String();
+                        }
+                        break;
```
adamdbrw, line 182: Is there a reason to break here? The current logic: find a path in path list, if it ends up with name we are looking for, we will terminate this or other way. Is there a reason not to look further? Another path in the list might be a proper match and an ament package.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -148,21 +148,55 @@ namespace ROS2
...
+            }
+
             AZ::IO::Path urdfProperPath(urdfFilePath);
             AZ::IO::Path packagePath;
             for (auto it = urdfProperPath.begin(); it != urdfProperPath.end(); it++)
             {
                 packagePath /= *it;
                 AZStd::string packageXmlCandite = (packagePath / "package.xml").String();
+                // check if it is Ament package
                 if (fileExists(packageXmlCandite))
```
adamdbrw, line 194: Please refactor out the section commented with `//check if it is Ament package`, it is reused.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -148,21 +148,55 @@ namespace ROS2
...
+                            return resolvedPath.String();
+                        }
+                        break;
+                    }
+                }
+            }
+
             AZ::IO::Path urdfProperPath(urdfFilePath);
             AZ::IO::Path packagePath;
             for (auto it = urdfProperPath.begin(); it != urdfProperPath.end(); it++)
```
adamdbrw, line 189: If we have urdf under /home/user/workspace/install/some_package/file.urdf, we will be looking at
`/home/package.xml`, `/home/user/package.xml` and so on, correct?
This does not look correct. What is the goal here?
In some cases we might also find something there which is not actually what we are looking for.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -148,21 +148,55 @@ namespace ROS2
...
             {
                 packagePath /= *it;
                 AZStd::string packageXmlCandite = (packagePath / "package.xml").String();
+                // check if it is Ament package
                 if (fileExists(packageXmlCandite))
                 {
+                    const AZ::IO::Path resolvedPath = packagePath / unresolvedPath;
```
adamdbrw, line 196: The file pointed by this path might not exist

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -103,12 +103,16 @@ namespace ROS2::Utils
...
+        AZ_Warning("UrdfAssetMap", enviromentalVariable, "AMENT_PREFIX_PATH is not found.");
```
adamdbrw, line 107: This is quite a fatal error for the Gem, since it means the ROS distro was not sourced and nothing will work. Could we promote it to an error?

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
michalpelka: It covers cases where the mesh is referenced as `package://meshes/file.urdf`.
I parse back path of requested urdf to find first package.xml. If somehow package.xml was deleted, it will try to poke '/home/package.xml' and '/package.xml'

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
michalpelka: Ahh, right, with current code the overlays are not covered. 
Maybe go search in opposite (end->begin) direction with break?

lawsonamzn:
Interesting.   I tried this on the raw husky model and got this during URDF import
```
(ResolveURDFPath) - ResolveURDFPath with package://husky_description/meshes/top_plate.stl
(ResolveURDFPath) - ResolveURDFPath with relative path to : husky_description/meshes/top_plate.stl
```


```
<21:29:31> [Error] (CollidersMaker) - Error loading collider. Invalid scene: , URDF path: package://husky_description/meshes/top_plate.stl
```
which comes from this in the raw urdf
```
<link name="top_plate_link">
    <visual>
      <origin rpy="0 0 0" xyz="0 0 0"/>
      <geometry>
        <mesh filename="package://husky_description/meshes/top_plate.dae"/>
      </geometry>
    </visual>
    <collision>
      <origin rpy="0 0 0" xyz="0 0 0"/>
      <geometry>
        <mesh filename="package://husky_description/meshes/top_plate.stl"/>
      </geometry>
    </collision>
  </link>
```

Note that these 2 source files have the same name, but I did fix the top_plate.stl to export to a different final name.
Is it maybe making an assumption about source file name -> atommesh name it shouldn't?

Also, it appears to be dumping a ton of spam in this format:
(RobotImporterWidget) - m_relativePath ... (every asset in the entire game, but only the daes?)

The top plate phsyxmesh (which is all it needs, I assume) does not appear in that list of all assets.

But  it is in asset browser:
image

I'd imagine that the algo that finds engine assets given the source assets would ask AP / the AP database for the source assets, then exercise the API to get the products from those source assets (rather than directly trying to search for the products).  

As in, when you see "blah.stl" in the URDF, I'd imagine the best way to find the physx mesh component would be to basically execute from the EditorAssetSystemAPI.h bus:

```
virtual bool GetSourceInfoBySourcePath(const char* sourcePath, AZ::Data::AssetInfo& assetInfo, AZStd::string& watchFolder)
Given the string "husky_description/meshes/top_plate.stl" it will output the asset info (including the UUID of the source file) as well as the folder it was found in.

given that UUID you can now execute:
virtual bool GetAssetsProducedBySourceUUID(const AZ::Uuid& sourceUuid, AZStd::vector<AZ::Data::AssetInfo>& productsAssetInfo)

to get a complete list of all the output files given that source (this includes their UUID, assedit, name, type, extension, etc) and you could see if outputs physx mesh and use that one.

**Comment by michalpelka** [comment]
@lawsonamzn , I would merge it. The issue you had with husky, is known an that is a case I am testing during development here : #224.

File: Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp [hunk 9]
```diff
@@ -146,23 +146,91 @@ namespace ROS2
         return filenames;
     }
 
+    AZStd::optional<AZ::IO::Path> GetResolvedPath(const AZ::IO::Path &packagePath,
+                                                 const AZ::IO::Path &unresolvedPath,
+                                                 const AZStd::function<bool(const AZStd::string&)>& fileExists)
+    {
+        AZ::IO::Path packageXmlCandite = packagePath / "package.xml";
+        if (fileExists(packageXmlCandite.String()))
```
lawsonamzn, line 154: minor, I assume in these var names you mean "Candidate" ?

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -146,23 +146,91 @@ namespace ROS2
...
+        {
+            AZ::IO::Path resolvedPath = packagePath / unresolvedPath;
+            if (fileExists(resolvedPath.String()))
+            {
+                return AZStd::optional<AZ::IO::Path>{resolvedPath};
+            }
+        }
+        return AZStd::optional<AZ::IO::Path>{};
```
lawsonamzn, line 162: minor, isn't it more correct to return
```
 return AZStd::nullopt;
```
to clearly indicate intent to return an optional that does not have a value.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -103,12 +103,16 @@ namespace ROS2::Utils
 
     UrdfAssetMap FindAssetsForUrdf(const AZStd::unordered_set<AZStd::string>& meshesFilenames, const AZStd::string& urdFilename)
     {
+        auto enviromentalVariable = std::getenv("AMENT_PREFIX_PATH");
+        AZ_Error("UrdfAssetMap", enviromentalVariable, "AMENT_PREFIX_PATH is not found.");
```
lawsonamzn, line 107: I'd recommend instead of saying AMENT_PREFIX_PATH is not found, here, indicate:
```
Environment variable AMENT_PREFIX_PATH is empty, (CONCEQUENCE), (HOW TO FIX)
```

like
```
Env var AMENT_PREFIX_PATH is not defined or is empty, will not be able to find package resources.  Try setting this variable to point at the folder containing xxxyyyyzzzz
```

Saves you on having to answer support questions :)

lawsonamzn (approved):
The changes requested are minor and shouldnt' block shipping.

hultonha:
Let me know when you'd like me to kick off an AR run @michalpelka 👍

michalpelka:
We can kick off build please @hultonha

byrcolin (approved):

adamdbrw:
Some comments are unresolved, but since they are minor, I believe we could go ahead and merge this. Especially since the AR wait times are quite high. I have prepared an issue to capture the remaining comments and submit as an improvement afterwards (https://github.com/o3de/o3de-extras/issues/270).

PR #227, opened 2023-03-22
As part of fixing https://github.com/o3de/o3de-extras/issues/118

lawsonamzn:
It was discovered that the error log from xacro is cut off on the first angle bracket, making it impossible to see the actual error problem fully if it occurs.

This change surrounds the error text with a code macro (three backticks) so that it shows up in markdown as the actual literal code, which causes Qt to ignore any angle brackets and other markdown inside the text.

Note that this was discovered during fixing 
https://github.com/o3de/o3de-extras/issues/118
but does not fix 118, only makes it better display when errors occur.

How it looks with these changes

image


lumberyard-employee-dm (approved):

cgalvan (approved):

michalpelka (approved):
Looks good.

PR #234, opened 2023-03-27
Change way of source asset handling in URDF importer

michalpelka:
Work in progress URDF Wizard page that is affected:
image
To do:
- [x] Deal with failed product asset (e.g. icon + button to bring AssetProcessor)
- [x] Copy packages to dedicated folder during import

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
@@ -7,8 +7,11 @@
  */
 
 #include "CheckAssetPage.h"
+#include "RobotImporter/Utils/SourceAssetsStorage.h"
 #include <AzCore/Math/MathStringConversions.h>
+#include <AzFramework/Asset/AssetSystemBus.h>
 #include <QHeaderView>
+#include <QPushButton>
 #include <QVBoxLayout>
 namespace ROS2
 {
@@ -17,19 +20,53 @@ namespace ROS2
         : QWizardPage(parent)
         , m_success(true)
         , m_missingCount(0)
+        , m_failureIcon(QStringLiteral(":/stylesheet/img/logging/failure.svg"))
+        , m_okIcon(QStringLiteral(":/stylesheet/img/logging/valid.svg"))
     {
         m_table = new QTableWidget(parent);
         SetTitle();
         QVBoxLayout* layout = new QVBoxLayout;
         layout->addWidget(m_table);
-        m_table->setColumnCount(4);
-        m_table->horizontalHeader()->setSectionResizeMode(0, QHeaderView::Stretch);
-        m_table->setShowGrid(true);
-        m_table->setSelectionMode(QAbstractItemView::SingleSelection);
-        m_table->setSelectionBehavior(QAbstractItemView::SelectRows);
-        m_table->setHorizontalHeaderLabels({ tr("URDF mesh path"), tr("CRC"), tr("Type"), tr("Asset source") });
+        m_table->setEnabled(true);
+        m_table->setAlternatingRowColors(true);
+        m_table->setMinimumHeight(500);
+        m_table->setMinimumWidth(1000);
         m_table->horizontalHeader()->setStretchLastSection(true);
+        m_table->setCornerButtonEnabled(false);
+        m_table->setSortingEnabled(false);
+        m_table->setColumnCount(5);
+        m_table->setShowGrid(true);
+        m_table->setMouseTracking(true);
+        m_table->setSelectionBehavior(QAbstractItemView::SelectRows);
+        m_table->setSelectionMode(QAbstractItemView::SingleSelection);
+        // Set the header items.
+        QTableWidgetItem* headerItem = new QTableWidgetItem(tr("URDF mesh path"));
+        headerItem->setTextAlignment(Qt::AlignVCenter | Qt::AlignLeft);
+        m_table->setHorizontalHeaderItem(0, headerItem);
+        headerItem = new QTableWidgetItem(tr("Resolved mesh from URDF"));
+        headerItem->setTextAlignment(Qt::AlignVCenter | Qt::AlignLeft);
+        m_table->setHorizontalHeaderItem(1, headerItem);
+        headerItem = new QTableWidgetItem(tr("Type"));
+        headerItem->setTextAlignment(Qt::AlignVCenter | Qt::AlignLeft);
+        m_table->setHorizontalHeaderItem(2, headerItem);
+        headerItem = new QTableWidgetItem(tr("Source asset"));
+        headerItem->setTextAlignment(Qt::AlignVCenter | Qt::AlignLeft);
+        m_table->setHorizontalHeaderItem(3, headerItem);
+        headerItem = new QTableWidgetItem(tr("Product asset"));
+        headerItem->setTextAlignment(Qt::AlignVCenter | Qt::AlignLeft);
+        m_table->setHorizontalHeaderItem(4, headerItem);
+        m_table->horizontalHeader()->resizeSection(0, 200);
+        m_table->horizontalHeader()->resizeSection(1, 350);
+        m_table->horizontalHeader()->resizeSection(2, 50);
+        m_table->horizontalHeader()->resizeSection(3, 400);
+        m_table->horizontalHeader()->resizeSection(4, 400);
+        m_table->verticalHeader()->hide();
+        connect(m_table, &QTableWidget::cellDoubleClicked, this, &CheckAssetPage::DoubleClickRow);
         this->setLayout(layout);
+        m_refreshTimer = new QTimer(this);
+        m_refreshTimer->setInterval(250);
+        m_refreshTimer->setSingleShot(false);
+        connect(m_refreshTimer, &QTimer::timeout, this, &CheckAssetPage::RefreshTimerElapsed);
     }
 
     void CheckAssetPage::SetTitle()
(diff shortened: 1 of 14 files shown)


adamdbrw:
These are great quality changes in progress, let me know when it is ready for review!

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -31,76 +38,133 @@ namespace ROS2::Utils
...
+        if (ok)
+        {
+            for (auto& product : productsAssetInfo)
+            {
+                if (product.m_assetType == typeId)
+                {
+                    AZStd::string assetPath;
+                    AZ::Data::AssetCatalogRequestBus::BroadcastResult(
+                        assetPath, &AZ::Data::AssetCatalogRequestBus::Events::GetAssetPathById, product.m_assetId);
+                    return assetPath;
```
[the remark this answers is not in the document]
michalpelka: I can see that multiple assets are created (including azmeshes) for larger files.  
I thought the first one was guaranteed to be the main one (I've tested this code on such cases).
Could you provide me details on "prefab" generation?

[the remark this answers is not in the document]
lawsonamzn: I'll try to give a brief detail on prefab generation.  Its a similar concept to URDF import.  In URDF, there is a URDF file that describes the transform heirarchy and the relation of the pieces to each other, and by convention, it references the meshes as individual DAE or STL files.   So the URDF itself is the "scene description" file.

In many other pipelines, however (common in games and other industry) the entire scene is assembled in the DCC tool like blender or maya, and the rig lives entirely in one file (minus optional attachments).  So for example imagine a blender or maya scene where the entire robot, all the joints, all the settings, and all the parts are inside the one DAE or FBX file (as seperate objects in a heirarchy of transforms).

the Prefab auto generation rule is on by default for all meshes (you'll see it show up as a prefab group rule in the assetinfo files or in the prefab tab in the GUI).  When active what it does is
* Create mesh export rules for each individual object in the scene heirarchy inside the FBX/Dae/STL/etc file, so that each seperate object from the DCC tool gets seperated into its own mesh file.  These meshes are always exported in mesh local space - that is, always with their pivot at the origin 0,0,0, with their transform in the heirarchy inversed out of their vertex data.
* Exports a .procprefab file as part of the export.  A procprefab file is an identical file format as normal prefab files, that is, a heirarchy of transforms of entities with components.  It assigns the transforms found in the DCC file, similar to how the URDF converter assigns the transforms found in the URDF file.  It also assigns the meshes it extracted from the DCC file into mesh components.  Again, very similar to robot importer.  The result is a .procprefab file which the user can then drag into the scene to spawn the scene heirarchy in the editor.    They can also create a prefab file that references the procprefab file and add overrides.

Both systems (Procedural prefab generation and URDF import) serve similar functions - make a prefab based on whats in the scene file.  however, in general, in URDF land, the URDF file itself is the scene file and the meshes it needs are already split into separate mesh files.  Wheras in procprefab land, the single FBX contains the entire scene (there is no URDF) and it needs to split them.

So in this case, I would recommend that if you are making a tool which imports a URDF, you can assume that any source DAE/FBX/STL referenced by it are already single mesh pieces and do not need a procprefab generaetd for each of them, in which case its probably best to create a scenesettings file that turns off the procprefab rule for them, so that they only generate the mesh files you expect and don't generate .procprefab files in addition.

An example here to illustrate.
Imagine you have a scene that has two cubes standing on top of each other in one dcc file, lets call it "shape_tower.fbx"
```
base_transform <0,0,0> <-- just a parent transform that is empty
     cube1 <--unit cube mesh, <0, 0.5, 0> so its sitting on top of origin 
        sphere1 <-- unit sphere mesh <0, 0.5, 0> relative to parent
```
If you allow the default processing to happen without changing anything, this 'scene' will be exported as
```
   shape_tower.atommodel <-- contains all the vertices from cube1, sphere1, all welded together in their world space poses, basically the entire scene as 1 mesh
   cube1_procprefab_(WART).atommodel <-- the cube only, centered around the origin, with no offset
   sphere1_procprefab_(WART).atommodel <-- the sphere only, centered around the origin, with no offset.
   shape_tower.procprefab <-- a prefab file, contianing 3 transforms and 2 mesh components with offsets set, meshes assigned to cube1_procprefab, sphere1.procprefab
```

If you delete the prefab group in the prefab tab, it will delte the procprefab mesh rules and you'll end up with it only making shape_tower.atommodel,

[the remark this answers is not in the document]
michalpelka: Totally agree, I did not know that such workflow exists, and it is default behavior.
Please find a new commit in PR, that disables procedural assets at all during URDF import.
It generate no procedural prefab (it also stops to create visual meshes if there is collider only mesh):
image
I still think it create a lot of product assets in local coordinate system (as you explained):
image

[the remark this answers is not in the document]
lawsonamzn: I'm surprised it does that still, actually, it would indicate that somehow the proc prefab rule is still making mesh groups for the finger even tho its been turned off.
all of the "default_xxxxxxx_(some guid)" files are procprefab stuff as far as I know

Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
```diff
@@ -139,4 +149,22 @@ namespace ROS2
...
+            if (processedAssets.contains(assetId))
+            {
+                continue;
+            }
+            const AZStd::string productRelPathVisual = Utils::GetModelProductAsset(assetId);
+            const AZStd::string productRelPathCollider = Utils::GetPhysXMeshProductAsset(assetId);
+            QString text = QString::fromUtf8(productRelPathVisual.data(), productRelPathVisual.size())+ " " +
+                QString::fromUtf8(productRelPathCollider.data(), productRelPathCollider.size());
+            bool isOk = productRelPathVisual.empty() || productRelPathCollider.empty();
+            m_table->setItem(i, 4, createCell(isOk, text));
```
lawsonamzn, line 165: The timer will continue hitting over and over, is it okay for this code to run over and over?  It depends on whether processedAssets.contains(that assetid), but does not add it to the processedAssets array, so its going to be constantly setting the item?

Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
```diff
@@ -139,4 +149,22 @@ namespace ROS2
         }
     }
 
+    void CheckAssetPage::RefreshTimerElapsed(){
+        for (int i =0; i < assetsId.size(); i++)
+        {
+            const auto &assetId  = assetsId[i];
```
lawsonamzn, line 155: style, you could use c++17 style
```
for (const auto& assetId : assetsId)
{ ... }

```
if assetsId is a member, it should probably be m_assetsId?

Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.h
```diff
@@ -40,12 +44,18 @@ namespace ROS2
...
     private:
         bool m_success;
+        QTimer* refreshTimer;
         QTableWidget* m_table {};
         QTableWidgetItem* createCell(bool isOk, const QString& text);
         QPushButton* m_reload {};
         unsigned int m_missingCount;
         void SetTitle();
+        AZStd::vector<AZ::Data::AssetId> assetsId;
+        AZStd::unordered_set<AZ::Data::AssetId> processedAssets;
```
lawsonamzn, line 54: style, `m_`

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -123,7 +122,55 @@ namespace ROS2::Utils
...
+        AZStd::string watchFolder;
+        AssetSysReqBus::BroadcastResult(
+            sourceAssetFound, &AssetSysReqBus::Events::GetSourceInfoBySourcePath, relativeSourcePath.c_str(), assetInfo, watchFolder);
+        if (!sourceAssetFound){
+            return foundAsset;
+        }
 
+        foundAsset.m_sourceGuid = assetInfo.m_assetId.m_guid;
+
+        foundAsset.m_assetId = assetInfo.m_assetId;
```
lawsonamzn, line 153: Beware of using this as an actual assetid.  The API is a bit confusing.

GetSourceInfoBySourcePath fills in the assetinfo structure, which sets the relative path, as well as the assetInfo.m_assetId, but it only sets the m_guid field of the assetId and always sets the subid to 0.  A real assetid is made of a tuple
```
  m_guid;  // the UUID of the source file (ie, the DAE or FBX or TIF or PNG or whatever)
  m_subId; // a unique number for each product generated by that source, since its 1-to-many
```

So when asking about a source file, using GetSourceInfoBySourcePath, the subid will always be 0 - this is not actually a valid assetid, its possible for a source file (like a TIF or DAE or something) to never output a product with subid 0 (its not an index, its a hash).

to get the subids (and thus the full assetid) you would need to query the output product(s) given that input source uuid, which is a different API which takes the source UUID (which can be gotten here, and also doesn't require use of GenerateRelativeSourcePath and also does not require you to use the database, its just another bus call, given a UUID of a source you can get all the output products from that source)

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -197,35 +241,137 @@ namespace ROS2::Utils
...
+                                   const AZStd::unordered_set<AZStd::string>& tip)
     {
         auto enviromentalVariable = std::getenv("AMENT_PREFIX_PATH");
         AZ_Error("UrdfAssetMap", enviromentalVariable, "AMENT_PREFIX_PATH is not found.");
         AZStd::string amentPrefixPath {enviromentalVariable};
 
         UrdfAssetMap urdfToAsset;
-        for (const auto& t : meshesFilenames)
+        // compute checksum for files from Tip
+        AZStd::unordered_map<AZ::Crc32,AZStd::string> tipFilesWithCrc;
```
lawsonamzn, line 320: It might not be necessary to use the DB or CRCs for this.

When a source file appears on disk, GetSourceInfoBySourcePath should be available to ask for its UUID (subid zero)
once you have its UUID, GetAssetsProducedBySourceUUID can be used to enumerate all the product files generated by that particular source path, which gives you their relative path as well as their full AssetID (subid non zero).

It all happens asynchronously though, so you'd have to wait until GetAssetJobsInfo (for that source path) indicates that all jobs are done.   GetAssetJobsInfo is good to call anyway since it can optioanlly escalate the jobs it returns to the top of hte list in case there is a job queue.

GetAssetJobsInfo will return a vector of jobs that were created for that asset and each of those will have a status JobStatus m_status, so you can poll for when its complete.

This would allow you to get a definite list of all the output files for any given input file (DAE, STL, etc) as well as escalate those jobs to the top of the queue in case there is a queue as well as know if 1 failed.  The same API also lets you retrieve the build logs for the files which could be parsed to show any errors or warnings.

Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
michalpelka: Reabsaed all members.
I need index for Qt row down in the code, so I cannot use a range-base loop.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: Removed m_assetId from my helper structure here to not cause confusion. It is used only to track source files, and m_subId is never used. The previous code was caused by my lack of understanding. Thank you for the clarification! :)

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: You are right, it is no need, keep legacy mechanism for now. We can remove next PR.

michalpelka:
Changes in action:
[Screencast from 04-20-2023 08:47:28 PM.webm](https://user-images.githubusercontent.com/3209244/233460449-6bb6345f-786c-4ddf-b57f-1cf833dfb6fc.webm)

nick-l-o3de (approved):
huge improvement on the previous system.

Gems/ROS2/Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.cpp
```diff
@@ -58,7 +59,16 @@ namespace ROS2
         options.isDisabledInSimMode = true;
         options.isDeletable = true;
 
-        options.toolbarIcon = ":/ROS2/ROS_import_icon.svg";
+        options.toolbarIcon = ":/ros/import.svg";
```
adamdbrw, line 62: remove this code.

adamdbrw (approved):

michalpelka:
Thanks for reviewing. PR was rebased, applying clang-format, and ready for AR.

PR #244, opened 2023-03-30
Camera sensor post processing abstraction

adamdbrw:
Small refactoring of camera sensor code needed to add post processing, which is very useful for simulation.
Includes exposing the interface to add custom post processing function to the pipeline before publishing data, both for depth and color.

### DIFF
--- a/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
@@ -6,37 +6,32 @@
  *
  */
 #include "CameraSensor.h"
-
-#include <AzCore/Math/MatrixUtils.h>
+#include <ROS2/Camera/CameraPostProcessingRequestBus.h>
 
 #include <Atom/RPI.Public/Base.h>
+#include <Atom/RPI.Public/FeatureProcessorFactory.h>
+#include <Atom/RPI.Public/Pass/PassFactory.h>
+#include <Atom/RPI.Public/Pass/PassSystemInterface.h>
 #include <Atom/RPI.Public/Pass/Specific/RenderToTexturePass.h>
 #include <Atom/RPI.Public/RPISystemInterface.h>
 #include <Atom/RPI.Public/RenderPipeline.h>
 #include <Atom/RPI.Public/Scene.h>
+#include <AzCore/Math/MatrixUtils.h>
 #include <AzFramework/Components/TransformComponent.h>
 #include <AzFramework/Scene/SceneSystemInterface.h>
-
-#include <Atom/RPI.Public/FeatureProcessorFactory.h>
-#include <Atom/RPI.Public/Pass/PassSystemInterface.h>
 #include <PostProcess/PostProcessFeatureProcessor.h>
-
-#include <Atom/RPI.Public/Pass/PassFactory.h>
 
 namespace ROS2
 {
     namespace Internal
     {
-
         /// @FormatMappings - contains the mapping from RHI to ROS image encodings. List of supported
         /// ROS image encodings lives in `sensor_msgs/image_encodings.hpp`
         /// We are not including `image_encodings.hpp` since it uses exceptions.
         AZStd::unordered_map<AZ::RHI::Format, const char*> FormatMappings{
-            { AZ::RHI::Format::R8G8B8A8_UNORM, "rgba8" },     
-            { AZ::RHI::Format::R16G16B16A16_UNORM, "rgba16" },
-            { AZ::RHI::Format::R32G32B32A32_FLOAT, "32FC4" }, // Unsuported by RVIZ2
-            { AZ::RHI::Format::R8_UNORM, "mono8" },           
-            { AZ::RHI::Format::R16_UNORM, "mono16" },
+            { AZ::RHI::Format::R8G8B8A8_UNORM, "rgba8" },     { AZ::RHI::Format::R16G16B16A16_UNORM, "rgba16" },
+            { AZ::RHI::Format::R32G32B32A32_FLOAT, "32FC4" }, // Unsupported by RVIZ2
+            { AZ::RHI::Format::R8_UNORM, "mono8" },           { AZ::RHI::Format::R16_UNORM, "mono16" },
             { AZ::RHI::Format::R32_FLOAT, "32FC1" },
         };
 
@@ -45,22 +40,19 @@ namespace ROS2
         AZStd::unordered_map<AZ::RHI::Format, int> BitDepth{
             { AZ::RHI::Format::R8G8B8A8_UNORM, 4 * sizeof(uint8_t) },
             { AZ::RHI::Format::R16G16B16A16_UNORM, 4 * sizeof(uint16_t) },
-            { AZ::RHI::Format::R32G32B32A32_FLOAT, 4 * sizeof(float) }, // Unsuported by RVIZ2
+            { AZ::RHI::Format::R32G32B32A32_FLOAT, 4 * sizeof(float) }, // Unsupported by RVIZ2
             { AZ::RHI::Format::R8_UNORM, sizeof(uint8_t) },
             { AZ::RHI::Format::R16_UNORM, sizeof(uint16_t) },
             { AZ::RHI::Format::R32_FLOAT, sizeof(float) },
         };
-
     } // namespace Internal
-    CameraSensorDescription::CameraSensorDescription(const AZStd::string& cameraName, float verticalFov, int width, int height, AZ::EntityId entityId)
-        : m_verticalFieldOfViewDeg(verticalFov)
-        , m_width(width)
-        , m_height(height)
+
+    CameraSensorDescription::CameraSensorDescription(const AZStd::string& cameraName, const CameraConfiguration& configuration)
+        : m_cameraConfiguration(configuration)
         , m_cameraName(cameraName)
-        , m_aspectRatio(static_cast<float>(width) / static_cast<float>(height))
+        , m_aspectRatio(static_cast<float>(configuration.m_width) / static_cast<float>(configuration.m_height))
         , m_viewToClipMatrix(MakeViewToClipMatrix())
         , m_cameraIntrinsics(MakeCameraIntrinsics())
-        , m_entityId(entityId)
     {
         ValidateParameters();
     }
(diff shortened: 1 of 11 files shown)


Gems/ROS2/Code/Include/ROS2/Camera/CameraSensorEditorRequestBus.h
```diff
@@ -0,0 +1,43 @@
...
+        //! @param postProcessingFunction function which will be called synchronously after rendering pipeline but before publishing.
+        virtual void RegisterDepthPostProcessing(const Camera::ImagePostProcessingFn& postProcessingFunction) = 0;
+
+        //! Register a color post-processing function.
+        //! Only one post processing can be registered, the new one overrides the former.
+        //! @param postProcessingFunction function which will be called synchronously after rendering pipeline but before publishing.
+        virtual void RegisterColorPostProcessing(const Camera::ImagePostProcessingFn& postProcessingFunction) = 0;
+
+    protected:
+        virtual ~CameraSensorEditorRequests() = default;
```
hultonha, line 39: nit: You can drop the `virtual` here as we never intend to delete the object through this request interface (and having the destructor in the protected section guards against improper use just in case 🙂)

Gems/ROS2/Code/Source/Camera/CameraSensor.h
```diff
@@ -16,31 +17,32 @@
 
 namespace ROS2
 {
+    struct CameraConfiguration
+    {
+        float verticalFieldOfViewDeg = 90.0f;  //!< vertical field of view of camera sensor.
+        int width = 640; //!< camera image width in pixels.
+        int height = 480; //!< camera image height in pixels.
+        Camera::ImagePostProcessingFn imagePostProcessingFunction; //!< post-processing function to apply to data.
```
hultonha, line 25: uber nit (repeated): Prefer sentence casing (starting with capital letter) for API comments.

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
```diff
@@ -0,0 +1,39 @@
...
+    struct CameraSensorConfiguration
+    {
+        AZ_TYPE_INFO(CameraSensorConfiguration, "{386A2640-442B-473D-BC2A-665D049D7EF5}");
+        static void Reflect(AZ::ReflectContext* context);
+
+        float verticalFieldOfViewDeg = 90.0f; //!< vertical field of view of camera sensor.
+        int width = 640; //!< Camera image width in pixels.
+        int height = 480; //!< Camera image height in pixels.
+        bool colorCamera = true; //!< Use color camera?
+        bool depthCamera = true; //!< Use depth camera?
```
hultonha, line 27: standards: Missing `m_` prefix (this is usually expected for all class members, even if public)

(ignore me if the standard O3DE coding standards don't apply here)

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -79,19 +72,20 @@ namespace ROS2
             return { publisher, AZStd::make_shared<CameraType>(description) };
         }
 
+        //! Helper that adds an image source.
+        //! @param imagePostProceessingFn a function to be called for post processing.
+        //! @param depth if true, depth image source is added, else it is a color image source.
+        void AddImageSource(const Camera::ImagePostProcessingFn& imagePostProceessingFn, bool depth = false);
```
hultonha, line 78: comment: I have a pet peeve against `bool` parameters 😅 Feel free to ignore me but to make the callsite clearer it might be worth making this an `enum` instead we can pass in (or make two functions called `AddImageSourceColor` and `AddImageSourceDepth` respectively which delegate to a shared internal/private function that can use a `bool`.

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorEditorComponent.cpp
```diff
@@ -113,8 +116,8 @@ namespace ROS2
 
         transform.SetUniformScale(frustumScale);
         debugDisplay.DepthTestOff();
-        const float ver = 0.5f * AZStd::tan((AZ::DegToRad(m_VerticalFieldOfViewDeg * 0.5f)));
-        const float hor = m_height * ver / m_width;
+        const float ver = 0.5f * AZStd::tan((AZ::DegToRad(m_cameraSensorConfiguration.verticalFieldOfViewDeg * 0.5f)));
+        const float hor = m_cameraSensorConfiguration.height * ver / m_cameraSensorConfiguration.width;
```
hultonha, line 120: nit/standards: Prefer to avoid abbreviations (so this would be `vertical` and `horizontal` (I realize this was like this before so feel free to ignore me, just a super small thing).

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorEditorComponent.h
```diff
@@ -43,16 +46,16 @@ namespace ROS2
         void BuildGameEntity(AZ::Entity* gameEntity) override;
 
     private:
-        // EntityDebugDisplayEventBus::Handler overrides
+        //! CameraSensorRequestBus::Handler overrides
```
hultonha, line 49: uber nit: Missing ellipsis (`overrides ...`)

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorEditorComponent.h
```diff
@@ -43,16 +46,16 @@ namespace ROS2
...
             const AZStd::string& topic, const AZStd::string& messageType, const AZStd::string& configName) const;
         SensorConfiguration m_sensorConfiguration;
-        float m_VerticalFieldOfViewDeg = 90.0f;
-        int m_width = 640;
-        int m_height = 480;
-        bool m_colorCamera = true;
-        bool m_depthCamera = true;
+        CameraSensorConfiguration m_cameraSensorConfiguration;
     };
 } // namespace ROS2
```
hultonha, line 61: uber nit: Could add missing EOL character here at EOF while we're here 🙂

hultonha (approved):
Changes look great! I just have a few minor suggestions, nothing blocking.

Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
```diff
@@ -196,7 +195,8 @@ namespace ROS2
     {
         RequestFrame(
             cameraPose,
-            [header, publisher](const AZ::RPI::AttachmentReadback::ReadbackResult& result)
+            [header, publisher, &postProcess = std::as_const(m_cameraSensorDescription.m_cameraConfiguration.imagePostProcessingFunction)](
```
MateuszWasilewski, line 198: Using &postProcess results in segfault while ending simulation. 
It seems that CameraSensor ends lifetime before lambda in certain scenarios 
Solution:
postProcess = m_cameraSensorDescription.m_cameraConfiguration.imagePostProcessingFunction

Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
```diff
@@ -52,12 +50,10 @@ namespace ROS2
...
     } // namespace Internal
-    CameraSensorDescription::CameraSensorDescription(const AZStd::string& cameraName, float verticalFov, int width, int height)
-        : m_verticalFieldOfViewDeg(verticalFov)
-        , m_width(width)
-        , m_height(height)
+    CameraSensorDescription::CameraSensorDescription(const AZStd::string& cameraName, const CameraConfiguration& configuration)
+        : m_cameraConfiguration(configuration)
         , m_cameraName(cameraName)
-        , m_aspectRatio(static_cast<float>(width) / static_cast<float>(height))
+        , m_aspectRatio(static_cast<float>(m_cameraConfiguration.width) / static_cast<float>(m_cameraConfiguration.height))
```
michalpelka, line 56: Nice!

Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
michalpelka: Agree, great you catch it in the debugger @MateuszWasilewski.  I would consider using a shared pointer to function to omit copying. 
The last RequestFrame occurs when a lot of assets are destroyed (e.g. it is common to see in ROS the last frame with funny textures). Please mechanism that won't give undefined behavior here.

Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
```diff
@@ -208,6 +208,7 @@ namespace ROS2
                 message.step = message.width * Internal::BitDepth.at(format);
                 message.data = std::vector<uint8_t>(result.m_dataBuffer->data(), result.m_dataBuffer->data() + result.m_dataBuffer->size());
                 message.header = header;
+                postProcess(message);
```
michalpelka, line 211: I would see the `postProcess` function that takes reference `result` and fill the message here.
Justification: when I use OpenCV I can create cvMat around buffer `result.m_dataBuffer` and I potentially have one copy less.

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.cpp
```diff
@@ -42,47 +33,54 @@ namespace ROS2
...
-                ->Field("Height", &ROS2CameraSensorComponent::m_height)
-                ->Field("Depth", &ROS2CameraSensorComponent::m_depthCamera)
-                ->Field("Color", &ROS2CameraSensorComponent::m_colorCamera);
+            serialize->Class<ROS2CameraSensorComponent, ROS2SensorComponent>()->Version(4)->Field(
+                "CameraSensorConfig", &ROS2CameraSensorComponent::m_cameraConfiguration);
         }
     }
 
-    void ROS2CameraSensorComponent::Activate()
+    void ROS2CameraSensorComponent::AddImageSource(const Camera::ImagePostProcessingFn& imagePostProceessingFn, bool depth)
```
michalpelka, line 41: I do not like this refactor, it makes code overconstrained to use case. We can have in future other sensors (e.g. optical flow). Maybe use here template argument?
```template typename T void ROS2CameraSensorComponent::AddImageSource<T>(const Camera::ImagePostProcessingFn& imagePostProceessingFn)```
This pattern is used are already with the helper `CreatePair<T>`. Code will follow DRY, won't be constrained and without ifs.

File: Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.h (hunk 4 above)
Comment (michalpelka) [reply Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.h:78]: Agree with Tom, I propose using a template argument.

**Review by michalpelka — COMMENTED** [review]
An important change, some minor issues.

File: Gems/ROS2/Code/Source/Camera/CameraSensor.cpp (hunk 10 above)
Comment (adamdbrw) [reply Gems/ROS2/Code/Source/Camera/CameraSensor.cpp]: The design is changed, please review the new one

File: Gems/ROS2/Code/Source/Camera/CameraSensor.cpp (hunk 8 above)
Comment (adamdbrw) [reply Gems/ROS2/Code/Source/Camera/CameraSensor.cpp]: The design is changed, please review the new one

**Comment by adamdbrw** [comment]
@michalpelka @hultonha I changed the design to a better one that supports multi-threaded nature and also works in builds. I also applied review comments. Please feel welcome to re-review.

File: Gems/ROS2/Code/Source/Camera/ROS2CameraSensorEditorComponent.h [hunk 12]
```diff
@@ -35,24 +35,19 @@ namespace ROS2
...
-        static void GetPr(AZ::ComponentDescriptor::DependencyArrayType& required);
         void Activate() override;
         void Deactivate() override;
 
         //! AzToolsFramework::Components::EditorComponentBase override
         void BuildGameEntity(AZ::Entity* gameEntity) override;
 
     private:
-        // EntityDebugDisplayEventBus::Handler overrides
+        //! EntityDebugDisplayEventBus::Handler overrides
```
michalpelka, line 45: It should not end up in documentation

michalpelka (approved):
Looks good.
Did you check if this change needs prefab updates?

adamdbrw:
Merge #271 first

michalpelka:
Rebased, backup is https://github.com/o3de/o3de-extras/tree/camera_sensor_post_processing_abstraction--20230417_1330

PR #267, opened 2023-04-13
Ros2ProjectTemplate performance fix

pawelbudziszewski:
Updated:
- In WarehouseSampe gem: turned on cached shadows and changed shadow filter 
- In Ros2ProjectTemplate: removed artifacts (unused lights), updated env lightning and camera settings

Fps changed from 33 to 88 in a configuration similar to #261

This addresses #261

PR #274, opened 2023-04-17
Remove additional lights from DemoLevel.prefab

macstepien:
They are redundant (placed outside of warehouse) and cause drop in fps (in my case from 25fps to 4fps).

demolevel_redundant_lights


adamdbrw:
This seems to be a duplicate of https://github.com/o3de/o3de-extras/pull/267

adamdbrw:
@macstepien Please confirm this was fixed by #267

macstepien:
I checked and can confirm that it was fixed by #267

PR #285, opened 2023-04-20
Create undo batch for prefab maker

### LINKED ISSUE #264: URDF Importer, prevent partial redo on imported URDF
There is an annoying bug that causes that gives unexpected results when importing URDF.
To recreate the issue:
 - import URDF
 - click a few times Cntrl+Z
The result is a partial deconstruction of the robot.
There is a mechanism to do that called `AzToolsFramework::ToolsApplicationRequests::Bus::Events::BeginUndoBatch`.

arturkamieniecki:
I've added a undo batch for URDF prefab creation, allowing complete undo and redo of imported URDF. This fixes #264. Unfortunately while undoing or redoing there are multiple errors: "[Error] (Prefab) - Failed to find an owning instance for entity with id <id here>". This seems to be related to issues: https://github.com/o3de/o3de/issues/15087, https://github.com/o3de/o3de/issues/15606. These errors also appear without the implemented undo batch, but only while redoing.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
@@ -10,6 +10,7 @@
 #include "CollidersMaker.h"
 #include "PrefabMakerUtils.h"
 #include <API/EditorAssetSystemAPI.h>
+#include <AzCore/Debug/Trace.h>
 #include <AzCore/IO/FileIO.h>
 #include <AzToolsFramework/Entity/EditorEntityHelpers.h>
 #include <AzToolsFramework/Prefab/PrefabLoaderInterface.h>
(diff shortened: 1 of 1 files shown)


Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -67,6 +67,11 @@ namespace ROS2
             AZStd::lock_guard<AZStd::mutex> lck(m_statusLock);
             m_status.clear();
         }
+
+         // Begin an undo batch for prefab creation process
+        AzToolsFramework::UndoSystem::URSequencePoint* currentUndoBatch;
+        AzToolsFramework::ToolsApplicationRequests::Bus::BroadcastResult(currentUndoBatch, &AzToolsFramework::ToolsApplicationRequests::BeginUndoBatch, "Robot Importer prefab creation");
```
adamdbrw, line 73: There is an early return between this call and the End call, such as lline 76/81. Would that persist and lead to a weird behavior?

michalpelka: Definitely it should finish batch on early return.

arturkamieniecki: It's an oversight on my end. I'll fix it.

arturkamieniecki:
I've compiled this pr with the engine from the pr mentioned in the attached issues. The prefab errors disappear.

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -67,12 +67,20 @@ namespace ROS2
             AZStd::lock_guard<AZStd::mutex> lck(m_statusLock);
             m_status.clear();
         }
+
+         // Begin an undo batch for prefab creation process
```
lumberyard-employee-dm, line 71: This variable needs to be initialized to `nullptr` as an EBus call might not be listening and therefore the result variable isn't updated.

forkercat (approved):

adamdbrw (approved):

PR #288, opened 2023-04-20
Fix not showing icon

### LINKED ISSUE #100: ROS2 Gem - "Robot Importer" icon is missing
There used to be an icon for "Robot Importer" toolbar and menu items. It must have been lost during the migration.
Screenshot from 2023-01-18 12-04-10

arturkamieniecki:
I've added the initialization of Qt resources enabling the ROS_import_icon.svg to be used by Qt. I've also renamed Resources.qrc to ROS2.qrc. This rename may not be needed, but leaving it as it was, wouldn't allow compiling the engine.
This fixes #100.


pawelbudziszewski (approved):

Gems/ROS2/Code/Source/ROS2EditorModule.cpp
```diff
@@ -11,6 +11,12 @@
 #include <Camera/ROS2CameraSensorEditorComponent.h>
 #include <RobotImporter/ROS2RobotImporterEditorSystemComponent.h>
 
+void InitROS2Resources()
+{
+    // Registration of Qt (ROS2.qrc) resources
+    Q_INIT_RESOURCE(ROS2);
```
adamdbrw, line 17: Where is this macro defined? I am wondering whether we can depend on its visibility based on included headers.

adamdbrw (approved):

michalpelka (approved):

Gems/ROS2/Code/Source/ROS2EditorModule.cpp
arturkamieniecki: This macro is defined in QtCore/qglobal.h. I can include it in this file to make sure it's always defined.

PR #297, opened 2023-04-24
Do not reprocess meshes on reimport

### LINKED ISSUE #289: URDF importer: importing duplicate assets
This issue concerns URDF importer in version #234 

When importing URDF, the importer creates a folder with imported assets (meshes). Such folders should be unique for different URDFs, but re-importing the same URDF uses the same assets folder. What's more, its content is removed during the import procedure. This may cause a lot of problems, including crash, in some cases when assets from this folder are being used by any entities on the scene.

**Steps to reproduce the crash:**
1. Import panda from [this urdf](https://github.com/o3de/o3de-physics-test-scene/tree/main/Assets/URDF/panda_description/panda_description) - this will create folder with assets and a prefab on the scene.
2. Modify the `panda.urdf` file by changing `hand.dae` to `finger.dae` in line 416
3. Import panda again

Crash in the moment of creating assets

michalpelka:
It is proposition to fix #289.


pawelbudziszewski (approved):

adamdbrw (approved):
Looks good. Some comments, take a look.

michalpelka:
Product asset generation is done only once (Users need to remove source assets from their Asset Browser to retrigger.)
[Uploading Screencast from 04-25-2023 12:32:50 PM.webm…]()

Issue #278, opened 2023-04-18
Extend  ROS2CameraSensor should provide ROS2CameraSensor service


adamkrawczyk:
Lack of the public interface for ROS2EditorCameraComponent to pull the configuration. Only available in editor component, have to be in game editor component also.

michalpelka:
It was fixed in https://github.com/o3de/o3de-extras/pull/342.

PR #300, opened 2023-04-26
Add ROS2CameraSensor service and horizontal field of view param

adamkrawczyk:
Added:
- macro `AZ_TYPE_INFO` to the CameraDepthSensor and CameraRGBDSensor
- services to `ROS2CameraSensorComponent`
- move `CameraSensor` header to shared directory 

Solves https://github.com/o3de/o3de-extras/issues/278
Based on https://github.com/RobotecAI/o3de-extras/pull/11

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Camera/CameraSensor.h
+++ b/Gems/ROS2/Code/Include/ROS2/Camera/CameraSensor.h
@@ -10,7 +10,7 @@
 #include <Atom/Feature/Utils/FrameCaptureBus.h>
 #include <AzCore/std/containers/span.h>
 
-#include "CameraPublishers.h"
+#include "../Source/Camera/CameraPublishers.h"
 #include <ROS2/ROS2GemUtilities.h>
 
 #include <chrono>
@@ -74,6 +74,7 @@ namespace ROS2
     class CameraDepthSensor : public CameraSensor
     {
     public:
+        AZ_TYPE_INFO(CameraDepthSensor, "{ADD4F0E4-F0ED-4770-9FF3-12AA4050953B}");
         CameraDepthSensor(const CameraSensorDescription& cameraSensorDescription, const AZ::EntityId& entityId);
 
     private:
@@ -85,6 +86,7 @@ namespace ROS2
     class CameraColorSensor : public CameraSensor
     {
     public:
+        AZ_TYPE_INFO(CameraColorSensor, "{8AA28B00-635D-4801-B1D8-37B66B535BA2}");
         CameraColorSensor(const CameraSensorDescription& cameraSensorDescription, const AZ::EntityId& entityId);
 
     private:
@@ -96,6 +98,7 @@ namespace ROS2
     class CameraRGBDSensor : public CameraColorSensor
     {
     public:
+        AZ_TYPE_INFO(CameraRGBDSensor, "{724D6C9F-EC61-4644-9EC9-C162AE5F26AF}");
         CameraRGBDSensor(const CameraSensorDescription& cameraSensorDescription, const AZ::EntityId& entityId);
 
         // CameraSensor overrides
--- a/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
@@ -5,7 +5,7 @@
  * SPDX-License-Identifier: Apache-2.0 OR MIT
  *
  */
-#include "CameraSensor.h"
+#include "ROS2/Camera/CameraSensor.h"
 #include <ROS2/Camera/CameraPostProcessingRequestBus.h>
 
 #include <Atom/RPI.Public/Base.h>
--- a/Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.cpp
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.cpp
@@ -32,6 +32,8 @@ namespace ROS2
                         &CameraSensorConfiguration::m_verticalFieldOfViewDeg,
                         "Vertical field of view",
                         "Camera's vertical (y axis) field of view in degrees.")
+                    ->Attribute(AZ::Edit::Attributes::Min, 0.0f)
+                    ->Attribute(AZ::Edit::Attributes::Max, 179.99f) // asserted to be less than 180 ([see the link](https://github.com/o3de/o3de-extras/blob/stabilization/2305/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp#L72))
                     ->DataElement(AZ::Edit::UIHandlers::Default, &CameraSensorConfiguration::m_width, "Image width", "Image width")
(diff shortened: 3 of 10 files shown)


adamdbrw (approved):
One things to correct, otherwise all good!

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.cpp
```diff
@@ -32,6 +33,15 @@ namespace ROS2
                         &CameraSensorConfiguration::m_verticalFieldOfViewDeg,
                         "Vertical field of view",
                         "Camera's vertical (y axis) field of view in degrees.")
+                    ->Attribute(AZ::Edit::Attributes::Min, 0.0f)
```
michalpelka, line 36: What is a point of exposing both HFOV and VFOV for the User?
The horizontal field of view do not affect anything.
Please rectify it. I would recommend having an HFOV getter.

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
```diff
@@ -19,7 +19,8 @@ namespace ROS2
         AZ_TYPE_INFO(CameraSensorConfiguration, "{386A2640-442B-473D-BC2A-665D049D7EF5}");
         static void Reflect(AZ::ReflectContext* context);
 
-        float m_verticalFieldOfViewDeg = 90.0f; //!< Vertical field of view of camera sensor.
+        float m_horizontalFieldOfViewDeg = 86.0f; //!< Horizontal field of view of camera sensor.
```
michalpelka, line 22: Why have default values changed here?

Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
```diff
@@ -68,9 +68,6 @@ namespace ROS2
 
     void CameraSensorDescription::ValidateParameters() const
     {
-        AZ_Assert(
```
michalpelka, line 71: Why this assert is removed?

michalpelka (changes requested):
It causes confusion and introduces users to unreflected component behavior. It will make the experience worse.
The introduced ROS2 Camera Component should have getters for HFOV and VFOV only.
Users should not be bombarded with undocumented, ineffective parameters.
Please retarget it to development.

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
adamkrawczyk: These values are the defaults for the intel realsense

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.cpp
adamkrawczyk: Horizontal fov is now affecting the image. Please consider using this parameter as an increase in the flexibility of the camera module through a changeable viewing parameter.

michalpelka: I understand that image resulution is now artifact of setting HFOV and VFOV?

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
michalpelka: I do not like that, it will change existing default cameras, it should have version converter if you want that. Please revert.

adamkrawczyk: Reverted

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.cpp
adamkrawczyk: Changed to setting only vertical FOV

michalpelka (approved):

Gems/ROS2/Code/Include/ROS2/Camera/CameraSensor.h
```diff
@@ -19,6 +19,7 @@ namespace ROS2
 {
     struct CameraConfiguration
     {
+        float m_horizontalFieldOfViewDeg = 90.0f; //!< Horizontal field of view of camera sensor.
```
adamdbrw, line 22: This is a computed field. If this is needed as a read info, instead of a member field, please use a getter function that computes it from the 3 other parameters.

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.cpp
```diff
@@ -32,6 +32,8 @@ namespace ROS2
...
+                    ->Attribute(AZ::Edit::Attributes::Max, 179.99f)
```
adamdbrw, line 36: Could you leave a comment why this value is used and not 180?

adamkrawczyk: It is because we assert this to be less than 180 ([see the link](https://github.com/o3de/o3de-extras/blob/stabilization/2305/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp#L72)) why then allow the user to set that?

Gems/ROS2/Code/Include/ROS2/Camera/CameraSensor.h
adamkrawczyk: Right! - changed

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.cpp
adamdbrw: The key is why it is being limited to under 180, the underlying, fundamental reason

adamdbrw:
@adamkrawczyk needs a rebase

Gems/ROS2/Code/Include/ROS2/Camera/CameraSensor.h
```diff
@@ -10,7 +10,7 @@
 #include <Atom/Feature/Utils/FrameCaptureBus.h>
 #include <AzCore/std/containers/span.h>
 
-#include "CameraPublishers.h"
+#include "../Source/Camera/CameraPublishers.h"
```
adamdbrw, line 13: This is not good - an API header should only include other API headers. I am not sure there is a reason to make the CameraSensor.h a part of the interface.

adamdbrw: Lets discuss this

adamdbrw (changes requested):

Gems/ROS2/Code/Include/ROS2/Camera/CameraSensor.h
adamdbrw, line 13: We will change this dependeny

adamdbrw (changes requested):

michalpelka:
Closing in favour https://github.com/o3de/o3de-extras/pull/342

PR #302, opened 2023-04-26
Update autoexec files

ZuzaOsa:
Spawn certain level after building for release


Templates/Ros2FleetRobotTemplate/Template/autoexec.cfg
```diff
@@ -0,0 +1 @@
+LoadLevel levels/warehousescene/warehouse.spawnable
```
adamdbrw, line 1:
```suggestion
LoadLevel levels/warehousescene/warehouse.spawnable

```

adamdbrw (approved):

zakmat:
Great job! That will definitely improve the onboarding experience. 
One thing that comes to my mind is that some users conventionally pass only the level name instead of the full asset path. How `SpawnableLevelSystem::LoadLevel` works is, it expects the spawnable with the level name to be located in the matching folder. How about we modify the level location in our templates to adhere?

pawelbudziszewski (approved):

PR #313, opened 2023-05-02
Updated gem.json to reflect versioning RFC

adamdbrw:
This change updates the Gem metadata to support versioning: https://github.com/o3de/sig-core/issues/44

Additionaly, it updates the json to include supported platform (Linux).

### DIFF
--- a/Gems/ROS2/gem.json
+++ b/Gems/ROS2/gem.json
@@ -18,8 +18,8 @@
         "ROS2"
     ],
     "compatible_engines": [
-        “o3de-sdk==1.2.0”,
-        “o3de>=2.0.0”
+        "o3de-sdk==1.2.0",
+        "o3de>=2.0.0"
     ],
     "icon_path": "preview.png",
     "requirements": "Requires ros2 installation (supported distributions: humble). Source your workspace before building the Gem",
--- a/Gems/ROS2/gem.json
+++ b/Gems/ROS2/gem.json
@@ -1,5 +1,9 @@
 {
     "gem_name": "ROS2",
+    "version": "1.0.0",
+    "platform": [
+        "Linux"
+    ],
     "display_name": "ROS2",
     "license": "Apache-2.0",
     "license_url": "https://opensource.org/licenses/Apache-2.0",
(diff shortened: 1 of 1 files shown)


Gems/ROS2/gem.json
```diff
@@ -1,5 +1,9 @@
 {
     "gem_name": "ROS2",
+    "version": "1.0.0",
+    "platform": [
```
AMZN-alexpete, line 4: The field name is actually plural, and I just noticed it is missing from the docs - will update my doc PR
```suggestion
    "platforms": [
```

Gems/ROS2/gem.json
```diff
@@ -13,8 +17,12 @@
     "user_tags": [
         "ROS2"
     ],
+    "compatible_engines": [
+        “o3de-sdk==1.2.0”,
+        “o3de>=2.0.0”
```
lumberyard-employee-dm, line 22: This isn't correct double quotes.
This is using the smart quotes which isn't plain ascii
```suggestion
        "o3de-sdk==1.2.0",
        "o3de>=2.0.0"
```

lumberyard-employee-dm (changes requested):

Gems/ROS2/gem.json
AMZN-alexpete, line 22: It looks like this text uses fancy quotes instead of plain ones
```suggestion
        "o3de-sdk==1.2.0",
        "o3de>=2.0.0"
```

adamdbrw:
@AMZN-alexpete @lumberyard-employee-dm I applied your suggestions, thank you! Not sure how the weird quotes got there, nice catch @lumberyard-employee-dm!

Gems/ROS2/gem.json
```diff
@@ -13,9 +17,20 @@
     "user_tags": [
         "ROS2"
     ],
+    "compatible_engines": [
+        "o3de-sdk==1.2.0",
+        "o3de>=2.0.0"
```
AMZN-alexpete, line 22: It looks like somehow the 23.05.0 o3de SDK engine name was not `o3de-sdk` in the installers, so I recommend lowering the `o3de` engine version requirement to 1.2.0 - I would leave the `o3de-sdk==1.2.0` in there in case we release a point fix that corrects this.

```suggestion
        "o3de>=1.2.0"
```

AMZN-alexpete (approved):

lumberyard-employee-dm (approved):

PR #329, opened 2023-05-15
Update Rosbot XL model

macstepien:
* changes Rosbot XL body model to a more accurate one
* adds antenna, as it was removed from the main body model
* fixes position of the front left and rear right wheel model

Screenshot from 2023-05-15 14-48-46


Gems/RosRobotSample/Assets/robot/rosbot_xl_description/meshes/components/antenna.dae
pawelbudziszewski, line 1: Please clean the mesh a bit (Blender->Decimate->...)

Gems/RosRobotSample/Assets/ROSbot.prefab
pawelbudziszewski, line 1: Please save materials as separate files for `body_link_visual` and both antenna-related objects. This way we will be able to tune material properties and add features like emissive.

Antennas can share the same material.

In the case of `body_link_visual` it seems that something is wrong with the current material component, so I suggest to:

1. Remove the `Material` component for `body_link_visual`
2. Create a new one
3. Click `Generate/Manage source materials` -> `Ok`

This should create all materials

pawelbudziszewski, line 1: Entities: `imu_link` and `cover_link` have a joint without a rigid body (O3DE complains about this). Please either remove the link or add a rigid body

pawelbudziszewski (commented):
Thank you for your work. The model looks great.

pawelbudziszewski (approved):

Issue #129, opened 2022-12-07
Sensor component: validate configuration fields


adamdbrw:
Validate fields in sensor configuration (such as frequency) - determine range and apply it

PR #344, opened 2023-05-22
Validate configuration fields for sensors

arturkamieniecki:
I've added limits for suitable configuration fields inside sensors. Most of them are minimum limits to eliminate values such as negative frequency. This addresses issue #129.

### DIFF
--- a/Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
@@ -19,13 +19,14 @@ namespace ROS2
         AZ_TYPE_INFO(CameraSensorConfiguration, "{386A2640-442B-473D-BC2A-665D049D7EF5}");
         static void Reflect(AZ::ReflectContext* context);
 
+        static constexpr int m_minWidth = 1;
+        static constexpr int m_minHeight = 1;
+
         float m_verticalFieldOfViewDeg = 90.0f; //!< Vertical field of view of camera sensor.
         float m_minVerticalFieldOfViewDeg = 0.0f;
         float m_maxVerticalFieldOfViewDeg = 360.0f;
         int m_width = 640; //!< Camera image width in pixels.
-        int m_minWidth = 1;
         int m_height = 480; //!< Camera image height in pixels.
-        int m_minHeight = 1;
         bool m_colorCamera = true; //!< Use color camera?
         bool m_depthCamera = true; //!< Use depth camera?
     };
--- a/Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
@@ -20,8 +20,12 @@ namespace ROS2
         static void Reflect(AZ::ReflectContext* context);
 
         float m_verticalFieldOfViewDeg = 90.0f; //!< Vertical field of view of camera sensor.
+        float m_minVerticalFieldOfViewDeg = 0.0f;
+        float m_maxVerticalFieldOfViewDeg = 360.0f;
         int m_width = 640; //!< Camera image width in pixels.
+        int m_minWidth = 1;
         int m_height = 480; //!< Camera image height in pixels.
+        int m_minHeight = 1;
         bool m_colorCamera = true; //!< Use color camera?
         bool m_depthCamera = true; //!< Use depth camera?
     };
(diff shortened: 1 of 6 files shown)


Gems/ROS2/Code/Source/Sensor/SensorConfiguration.cpp
```diff
@@ -37,6 +38,7 @@ namespace ROS2
                         "Toggle publishing for topic")
                     ->DataElement(
                         AZ::Edit::UIHandlers::Default, &SensorConfiguration::m_frequency, "Frequency", "Frequency of publishing [Hz]")
+                    ->Attribute(AZ::Edit::Attributes::Min, AZStd::numeric_limits<float>::epsilon())
```
adamdbrw, line 41: I guess "1" makes plenty of sense here, but perhaps we have sensors that wish to publish something less often. Could we set a rational limit such as once per day? @michalpelka what is your opinion?

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
```diff
@@ -20,8 +20,12 @@ namespace ROS2
         static void Reflect(AZ::ReflectContext* context);
 
         float m_verticalFieldOfViewDeg = 90.0f; //!< Vertical field of view of camera sensor.
+        float m_minVerticalFieldOfViewDeg = 0.0f;
```
adamdbrw, line 23: These limits are const values and not reflected. Could we group / separate them into private internal structure called Limits?

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.h
```diff
@@ -40,6 +40,7 @@ namespace ROS2
     private:
         //! Length of filter that removes numerical noise
         int m_filterSize{ 10 };
+        int m_minFilterSize{ 0 };
```
adamdbrw, line 43: Also a constant

Gems/ROS2/Code/Source/Sensor/SensorConfiguration.cpp
arturkamieniecki: I wanted to only disallow the user to input 0 as a frequency, so I figured that the smallest possible float would suffice. It seems that the O3DE Engine allows 0.0000001 as the smallest possible number to be input. This would allow sensors to update every 115 days. For now, I'll change the minimum to once per day.

Gems/ROS2/Code/Source/Camera/CameraSensorConfiguration.h
```diff
@@ -18,8 +18,18 @@ namespace ROS2
...
-
+        struct Limits
+        {
+            int m_minWidth;
+            int m_minHeight;
+        };
+        static constexpr Limits m_limits{
+            1,
+            1,
+        };
```
spham-amzn, line 29: Declaring a struct `Limits` here seems like overkill. Can't `m_minWidth` and `m_minHeight` be constexpr values for the `CameraSensorConfiguration` instead?

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.h
```diff
@@ -38,6 +38,11 @@ namespace ROS2
         //////////////////////////////////////////////////////////////////////////
 
     private:
+        struct Limits
+        {
+            int m_minFilterSize;
+        };
+        static constexpr Limits m_limits{ 1 };
```
spham-amzn, line 45: Same comment here as above regarding the need to declare a struct for this

Gems/ROS2/Code/Source/Sensor/SensorConfiguration.cpp
michalpelka: I agree with @arturkamieniecki, we should be able to publish with a frequency smaller than 1 Hz. Eps is good choice.

Gems/ROS2/Code/Source/Sensor/SensorConfiguration.cpp
```diff
@@ -9,6 +9,7 @@
 #include <AzCore/Serialization/EditContext.h>
 #include <AzCore/Serialization/EditContextConstants.inl>
 #include <ROS2/Sensor/SensorConfiguration.h>
+#include <limits>
```
michalpelka, line 12: I think it is unintentional.

adamdbrw (approved):

adamdbrw:
@arturkamieniecki if you could please rebase this PR, we can proceed to merge it

arturkamieniecki:
I've rebased onto the development branch. The merge conflict was due to some limits implemented in the imu sensor code. I've solved this by moving the constants into the header to better match other changes done in this PR.

nick-l-o3de (approved):

adamdbrw (approved):

Issue #135, opened 2022-12-06
Handle ROS 2 distro change between builds


adamdbrw:
**Is your feature request related to a problem? Please describe.**
Currently, no targets would be built in the following scenario:
1) The user sourced ROS 2 Galactic.
2) The project was built with the Gem 
3) The user sourced ROS 2 Humble
4) Built command is ran.

The ENV variable ROS_DISTRO is not considered a dependency, but it should.

hultonha:
Tagging @lumberyard-employee-dm and @lawsonamzn who may be able to advice on this.

The key problem is when switching the ROS_DISTRO, the generated build files won't be able to detect this so a full rebuild will be required.

One solution may be to create a target the always needs to be built, but this hasn't been implemented. We'd be very grateful for any suggestions/advice. Thanks!

lawsonamzn:
I think there's actually 2 problems hidden here.
1. CMake itself assumes that if the user messes with the environment, the user is smart enough to re-run cmake configure.  It doesn't consider env vars as things to watch for change to cause a re-configure, which also causes no rebuild.  There does not appear to be a way to make CMake aware of env vars at all during its pre-check phase, and if it passes the pre-check phase, it does not even attempt to reconfigure at all.  
2. Even if you do reconfigure, CMake itself also caches things like paths to programs found, if using find_program() inside the cmake cache.

Maybe a custom build rule can be added that causes the build to fail if it detects this (you can read the value of env vars in cmake) but also writes a file to the build tree that contains the current set of env vars (ie, a file that contains "ROS_DISTRO=galactic") which is used to detect the new distro, during this custom build rule.  The file can be written/created during configure, and made a dependency of configure, so if it changes, it forces a reconfigure.  Then if the user switches the distro and runs the build, this custom pre-build step would detect the change, write the new file (ROS_DISTRO=humble) and error out saying "Please re-run the build, distro has changed".  If you do that, then the next time they try to run cmake build, it will rerun configure, since it would depend on that file.

Maybe something like that?

See Also:
Cmake property CONFIGURE_DEPENDS - used to add a file to the list of files that trigger auto-reconfigure before build.
Cmake ENV var query - you can get the value of an env var.
CMake file parsing - via json or plaintext, cmake can read a file during cmake script execution
CMake custom pre-build commands : add_custom_command(TARGET ... PRE_BUILD .. Command to run)
CMake script mode commands (you can tell Cmake to execute a cmake script as the custom build command)

PR #345, opened 2023-05-22
Add check for ROS2 distribution in CMakeLists.txt

arturkamieniecki:
I've modified CMakeLists.txt to check if the initial configuration was run under the same ROS 2 distribution as the one sourced during project setup and build. I've changed CMakeLists.txt such that the ROS 2 distribution will be cached into CMakeCache.txt during the project's configuration. A check during the configuration process prevents users from building when their sourced distribution is different. I've also added the same check during the build process.

Removing the CMakeCache.txt, as proposed by me (```Remove CMakeCache.txt and reconfigure project.```), does not cause rebuilding of the O3DE Engine.

This addresses #135.

### DIFF
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -25,12 +25,18 @@ get_property(ROS_DISTRO_TYPE CACHE ROS_DISTRO PROPERTY TYPE)
 # Perform checks with cached ROS 2 distribution and sourced distribution. Save the distribution into cache if it was not cached before or if they do not match end with error.
 if (NOT ROS_DISTRO_TYPE)
     set(ROS_DISTRO $ENV{ROS_DISTRO} CACHE STRING "ROS 2 distribution of current configuration" FORCE)
-elseif(NOT "${ROS_DISTRO}" STREQUAL "$ENV{ROS_DISTRO}")
-    message(FATAL_ERROR "ROS 2 distribution does not match. Configuration created for: ${ROS_DISTRO}, sourced distribution: $ENV{ROS_DISTRO}. Remove CMakeCache.txt and reconfigure project.")
+elseif(NOT "$ENV{ROS_DISTRO}" STREQUAL "${ROS_DISTRO}")
+    get_cmake_property(CACHED_VARIABLES CACHE_VARIABLES)
+    set(ROS_DISTRO_COPY ${ROS_DISTRO})
+    # Iterate over cached variables and unset them
+    foreach(CACHED_VARIABLE ${CACHED_VARIABLES})
+        unset(${CACHED_VARIABLE} CACHE)
+    endforeach()
+    message(FATAL_ERROR "ROS 2 distribution does not match. Configuration created for: ${ROS_DISTRO_COPY}, sourced distribution: $ENV{ROS_DISTRO}.  Removed invalid configuration, please reconfigure project.")
 endif()
 
 # Add a custom target to always check during build if sourced and cached distributions match.
-ADD_CUSTOM_TARGET(
+add_custom_target(
     ROS2DistributionCheck ALL 
     COMMENT "Checking if sourced ROS 2 distribution matches with the configuration"
     COMMAND ${CMAKE_COMMAND} -DROS_DISTRO=${ROS_DISTRO} -P ${CMAKE_CURRENT_SOURCE_DIR}/checkROS2Distribution.cmake
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -19,6 +19,23 @@ if (NOT ROS2_FOUND)
 endif()
 message(DEBUG "Building ${gem_name} Gem with ros2 $ENV{ROS_DISTRO}")
 
+# Check if ROS 2 distribution is cached
+get_property(ROS_DISTRO_TYPE CACHE ROS_DISTRO PROPERTY TYPE)
+
+# Perform checks with cached ROS 2 distribution and sourced distribution. Save the distribution into cache if it was not cached before or if they do not match end with error.
+if (NOT ROS_DISTRO_TYPE)
+    set(ROS_DISTRO $ENV{ROS_DISTRO} CACHE STRING "ROS 2 distribution of current configuration" FORCE)
+elseif(NOT "${ROS_DISTRO}" STREQUAL "$ENV{ROS_DISTRO}")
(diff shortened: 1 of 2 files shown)


lawsonamzn (commented):
looks like it will work - is it necessary to remove the cache, though?
Does it cache stuff that it doesn't recompute even if you do a reconfigure? 

To be more specific, do we have to force a cache deletion here, or can we force an auto-reconfigure?

lumberyard-employee-dm:
> looks like it will work - is it necessary to remove the cache, though? Does it cache stuff that it doesn't recompute even if you do a reconfigure?
> 
> To be more specific, do we have to force a cache deletion here, or can we force an auto-reconfigure, since a lot of stuff is checked in the configuration step but a problem is that env vars changing doesn't cause a reconfigure...

An alternative is using the `-UROS_DISTRO_TYPE` option to undefine the `ROS_DISTRO_TYPE` cache variable to force a new check of the ROS2 distro

lawsonamzn:
> 

I'm thinking more along the lines of what inside the cache is never checked again even on reconfigure.  An example is things like when find_program or find_library is used, unless those vars are cleared, or FORCE is used, cmake tends to cache them.  I ran into this quite a bit when dealing with the 3p system.

Gems/ROS2/Code/checkROS2Distribution.cmake
```diff
@@ -0,0 +1,3 @@
+if (NOT "$ENV{ROS_DISTRO}" STREQUAL "${ROS_DISTRO}")
```
lumberyard-employee-dm, line 1: This file is missing a copyright and a newline at the end as well.

lumberyard-employee-dm:
> > 
> 
> I'm thinking more along the lines of what inside the cache is never checked again even on reconfigure. An example is things like when find_program or find_library is used, unless those vars are cleared, or FORCE is used, cmake tends to cache them. I ran into this quite a bit when dealing with the 3p system. To actually force those you'd have to also have to undef (in the cache) the vars that are caching those program locations.

How about checking if the value in the Environment variable is different than the one in the cache if the cache variable is set?
If so force update the cache with the new value.

lawsonamzn:
> > > 
> > 
> > 
> > I'm thinking more along the lines of what inside the cache is never checked again even on reconfigure. An example is things like when find_program or find_library is used, unless those vars are cleared, or FORCE is used, cmake tends to cache them. I ran into this quite a bit when dealing with the 3p system. To actually force those you'd have to also have to undef (in the cache) the vars that are caching those program locations.
> 
> How about checking if the value in the Environment variable is different than the one in the cache if the cache variable is set? If so force update the cache with the new value.
> 
> That can be an option instead of a `message(FATAL_ERROR)`

I think the one where it fails the build due to an error in the **build step** is valid, since CMake doesn't actually check the value of Env Vars by default during build if nothing has changed in its dependency tree.    

So maybe the flow would be

If you configure, it wipes the vars and configures with the new sources / env vars.
If you build _without configuring_ and it skips the confiugure step, the build step detects the mismatch, and forces a failure but also clears the cache variables or touches a file or something that wil cause the next attempt to auto-reconfigure?

lumberyard-employee-dm:
> > > > 
> > > 
> > > 
> > > I'm thinking more along the lines of what inside the cache is never checked again even on reconfigure. An example is things like when find_program or find_library is used, unless those vars are cleared, or FORCE is used, cmake tends to cache them. I ran into this quite a bit when dealing with the 3p system. To actually force those you'd have to also have to undef (in the cache) the vars that are caching those program locations.
> > 
> > 
> > How about checking if the value in the Environment variable is different than the one in the cache if the cache variable is set? If so force update the cache with the new value.
> > That can be an option instead of a `message(FATAL_ERROR)`
> 
> I think the one where it fails the build due to an error in the **build step** is valid, since CMake doesn't actually check the value of Env Vars by default during build if nothing has changed in its dependency tree.
> 
> So maybe the flow would be
> 
> If you configure, it wipes the vars and configures with the new sources / env vars. If you build _without configuring_ and it skips the configure step, the build step detects the mismatch, and forces a failure but also clears the cache variables or touches a file or something that will cause the next attempt to auto-reconfigure?

Ok, that can work 95% of the way. The one thing that can't be done is to clear the cache from the build step since [checkROS2Distribution.cmake](https://github.com/o3de/o3de-extras/pull/345/files#diff-6b7b5fdec66b3029927dc7ddfeaa4d79a15ff3494a6a59258954820f18c9aa50) custom target doesn't have access to the cmake binary directory cache values since it is an external command.

Gems/ROS2/Code/checkROS2Distribution.cmake
michalpelka, line 1: Do we need the same code in two places?

Gems/ROS2/Code/CMakeLists.txt
```diff
@@ -19,6 +19,23 @@ if (NOT ROS2_FOUND)
...
 message(DEBUG "Building ${gem_name} Gem with ros2 $ENV{ROS_DISTRO}")
 
+# Check if ROS 2 distribution is cached
+get_property(ROS_DISTRO_TYPE CACHE ROS_DISTRO PROPERTY TYPE)
+
+# Perform checks with cached ROS 2 distribution and sourced distribution. Save the distribution into cache if it was not cached before or if they do not match end with error.
+if (NOT ROS_DISTRO_TYPE)
+    set(ROS_DISTRO $ENV{ROS_DISTRO} CACHE STRING "ROS 2 distribution of current configuration" FORCE)
+elseif(NOT "${ROS_DISTRO}" STREQUAL "$ENV{ROS_DISTRO}")
+    message(FATAL_ERROR "ROS 2 distribution does not match. Configuration created for: ${ROS_DISTRO}, sourced distribution: $ENV{ROS_DISTRO}. Remove CMakeCache.txt and reconfigure project.")
```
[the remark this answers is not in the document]
lawsonamzn: Its the nuclear option but if it works and solves the problem, then its a lot easier to say "Configuration has been automatically reset, please try again" than to tell them to dig in their file system. :)

lawsonamzn (approved):
Nice.  Its a brute force method but from the user's point of view they will get the error, and then if they try again, cmake will reconfigure and retry automatically...

spham-amzn (approved):

michalpelka (approved):
:boom:

adamdbrw (approved):
Nice work!

lawsonamzn:
I did a clean build, it looks like it passed.

PR #348, opened 2023-05-24
constrain IMU filter length

zakmat:
When the IMU's filter length is set to 0 sensor fails to publish data. This PR is adding a constraint on the IMU component's `Filter Length` property as only positive values make sense

### DIFF
--- a/Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
+++ b/Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
@@ -50,6 +50,7 @@ namespace ROS2
                         &ROS2ImuSensorComponent::m_filterSize,
                         "Filter Length",
                         "Filter Length, large value allows to reduce numeric noise")
+                    ->Attribute(AZ::Edit::Attributes::Min, 1)
                     ->DataElement(
                         AZ::Edit::UIHandlers::Default,
                         &ROS2ImuSensorComponent::m_includeGravity,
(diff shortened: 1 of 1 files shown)


Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
```diff
@@ -50,6 +50,7 @@ namespace ROS2
                         &ROS2ImuSensorComponent::m_filterSize,
                         "Filter Length",
                         "Filter Length, large value allows to reduce numeric noise")
+                    ->Attribute(AZ::Edit::Attributes::Min, 1)
```
spham-amzn, line 53: Is there also a max value we can attribute here as well?

adamdbrw: I guess it does not make much sense to have a very large moving window. Maybe 200 could be a good maximum?

michalpelka: Agree, add 200 maximum samples it will be 2 seconds for 100 Hz.

zakmat: I have added a max value attribute at 200 and changed UIHandler to Slider for a better user experience

adamdbrw (approved):

michalpelka (approved):

Issue #140, opened 2022-10-12
Old static TF's are being published even if the scene is stopped or the name has changed


pijaro:
**Bug description**
All old static TF's are being published even if the scene is stopped or the name has changed

**To Reproduce**
1. Add FrameComponent with some namespace (`namespace#1`)
2. Hit play scene
3. Exit play mode
4. Change FrameComponent Namespace  (`namespace#2`)
5. Hit play mode
6. Exit play mode

**Expected behavior**
No static tf's in ROS 2 environment

**Current behavior**
Both tf's with `namespace#1` and 'namespace#2` are visible. Editor restart is required to get rid of them.

arturkamieniecki:
I have recreated this issue in ROSConDemo (with https://github.com/o3de/ROSConDemo/pull/240 cherry-picked into development) with o3de-extras (last commit https://github.com/o3de/o3de-extras/commit/e9f49e6cfc1df9766bf3fde6bb0bc7f37a1692c4).
Steps to recreate:
1. launch ROSConDemo and create an apple_kraken_v1 prefab
2. in the kraken prefab under base_link set FrameComponents namespace to something (`namespace#1`)
3. launch a terminal with `ros2 topic echo tf_static`
4. start the simulation
5. wait for `ros2 topic echo tf_static` to print something
6. stop the simulation
7. change the namespace to something else (`namespace#2`)
8. start the simulation
9. wait for `ros2 topic echo tf_static`
10. stop the simulation
11. the previous namespace tf_static messages should be visible

arturkamieniecki:
After some research about removing tf_static messages, I've come to the conclusion that TF2 does not support this kind of directly. In paragraph 4.8 at http://wiki.ros.org/tf2 there is a similar example to this issue. 

The StaticTransformBroadcaster is not reset after changes in the ROS 2 Frames configuration. Unfortunately just resetting the broadcaster does not fix the issue. Only resetting the whole node resolves this problem. I've created a proof of concept PR that fixes this issue, but it has its own problem. The PR is #354.

arturkamieniecki:
Closed with #354.

PR #354, opened 2023-06-09
Reset ROS 2 Node after simulation exit

arturkamieniecki:
This PR is WIP.

I've added logic to reset the ROS 2 node after every simulation exit, to delete all tf_static messages. This PR addresses #140 and as of now acts as a proof of concept.

There is a warning created on the simulation exit. I have not yet figured out how to prevent this warning. The warning in question is:
```[WARN] [1686322333.431306351] [rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.```

### DIFF
--- a/Gems/ROS2/Code/Source/Clock/SimulationClock.cpp
+++ b/Gems/ROS2/Code/Source/Clock/SimulationClock.cpp
@@ -73,10 +73,4 @@ namespace ROS2
         AZStd::sort(frameTimeSorted.begin(), frameTimeSorted.end());
         m_currentMedian = frameTimeSorted[frameTimeSorted.size() / 2];
     }
-
-    void SimulationClock::ResetPublisher()
-    {
-        m_clockPublisher.reset();
-    }
-
 } // namespace ROS2
--- a/Gems/ROS2/Code/Source/Clock/SimulationClock.cpp
+++ b/Gems/ROS2/Code/Source/Clock/SimulationClock.cpp
@@ -73,4 +73,10 @@ namespace ROS2
         AZStd::sort(frameTimeSorted.begin(), frameTimeSorted.end());
         m_currentMedian = frameTimeSorted[frameTimeSorted.size() / 2];
     }
+
+    void SimulationClock::ResetPublisher()
+    {
+        m_clockPublisher.reset();
+    }
+
 } // namespace ROS2
--- a/Gems/ROS2/Code/Source/ROS2SystemComponent.cpp
+++ b/Gems/ROS2/Code/Source/ROS2SystemComponent.cpp
@@ -21,7 +21,6 @@
 #include <AzCore/Time/ITime.h>
 #include <AzCore/std/smart_ptr/make_shared.h>
 #include <AzCore/std/string/string_view.h>
-#include <tf2_ros/static_transform_broadcaster.h>
 
 namespace ROS2
 {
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: Gems/ROS2/Code/Source/ROS2SystemComponent.cpp [hunk 1]
```diff
@@ -167,6 +170,32 @@ namespace ROS2
         }
     }
 
+    void ROS2SystemComponent::ResetNode()
+    {
+        if (!m_wasNodeReset)
```
adamdbrw, line 175: This is not great as there is no obvious place to call it. Frame Components are many. 

Instead, try this:
1) Determine what exactly we want and need to reset between simulation runs.
2) Have an object with lifetime controlled by Start / Stop simulation events. You might hold it in this System Component but use these events to reset. Make sure lifetime is clear by design and cannot be broken.

arturkamieniecki:
I've updated the code to be more elegant and reset the ROS 2 node when the simulation starts. Unfortunately, the same error persists. It is due to some publishers not being reset.

arturkamieniecki:
The errors mentioned above are due to https://github.com/o3de/ROSConDemo/issues/253.

arturkamieniecki:
This PR is now ready for review.
I've modified the ROS2EditorSystemComponet so that the ros2_node will only be active during the simulation. This fixes the bug with old static_tfs being published.

Gems/ROS2/Code/Source/ROS2SystemComponent.cpp
```diff
@@ -138,6 +139,10 @@ namespace ROS2
         m_loadTemplatesHandler.Disconnect();
         m_dynamicTFBroadcaster.reset();
         m_staticTFBroadcaster.reset();
+        m_executor->remove_node(m_ros2Node);
+        m_executor.reset();
+        m_simulationClock->ResetPublisher();
+        m_ros2Node.reset();
```
adamdbrw, line 145: I can see reasons why the node should be up, such as ROS parameters. However, the change can be easily introduced with such feature.

Gems/ROS2/Code/Source/Clock/SimulationClock.cpp
```diff
@@ -73,4 +73,10 @@ namespace ROS2
         AZStd::sort(frameTimeSorted.begin(), frameTimeSorted.end());
         m_currentMedian = frameTimeSorted[frameTimeSorted.size() / 2];
     }
+
+    void SimulationClock::ResetPublisher()
```
adamdbrw, line 77: Could you please describe (document) why only reseting the publisher is preferred (keeping the state of m_currentMedian and m_frameTimes? Why not just reset the clock object pointer?

adamdbrw:
Looks very nice - just a couple of clarifications needed!

Gems/ROS2/Code/Source/Clock/SimulationClock.cpp
arturkamieniecki: Resetting the clock object pointer is better. There isn't a reason to keep these states between simulation runs. I'll update the code.

Gems/ROS2/Code/Source/ROS2SystemComponent.cpp
arturkamieniecki: I'm not quite sure what you are suggesting. Are you implying that the node should be active outside of the simulation and that it should be activated immediately after resetting?

adamdbrw: Yes, a node which is active outside of simulation mode could support setting of ROS [parameters](https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Parameters/Understanding-ROS2-Parameters.html#ros2-param-set) on actual Editor Components, which is quite a valid use-case.

However, this could be changed once we implement parameters.

Gems/ROS2/Code/Source/Clock/SimulationClock.cpp
```diff
@@ -73,4 +73,5 @@ namespace ROS2
         AZStd::sort(frameTimeSorted.begin(), frameTimeSorted.end());
         m_currentMedian = frameTimeSorted[frameTimeSorted.size() / 2];
     }
+
```
adamdbrw, line 76: nitpick: this newline looks unintentional

adamdbrw (approved):

pawelbudziszewski (approved):

PR #377, opened 2023-07-03
More resilient async ros2 image post processing

MateuszWasilewski:
It was found that using LocklessDispatch allows for processing multiple frames independently but provides no guarantees while disconnecting. 
Current Implementation sometimes results in asserts on simulation stop if post processing takes significant amount of time.

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
@@ -7,28 +7,26 @@
  */
 #pragma once
 
+#include "AzCore/EBus/EBusSharedDispatchTraits.h"
 #include <AzCore/Component/EntityId.h>
-#include <AzCore/EBus/EBus.h>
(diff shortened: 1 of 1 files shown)


Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
```diff
@@ -7,28 +7,26 @@
...
-
 #include <sensor_msgs/msg/image.hpp>
 
 namespace ROS2
 {
     //! Interface class that allows to add post-processing to the pipeline
-    class CameraPostProcessingRequests : public AZ::EBusTraits
+    //!
+    //! Each function call can be processed without blocking Bus for other dispatches.
+    //! Bus connects / disconnects cannot happen within event dispatch.
```
adamdbrw, line 21:
```suggestion
    //! Do not use connects / disconnects to this bus during event dispatch, as they are not allowed for this concurrency model.
```

adamdbrw (approved):

michalpelka (approved):

PR #366, opened 2023-06-21
Manipulation Components redesign & refactor

adamdbrw:
Improvements to components for manipulating chains of joints and articulation links, such as in manipulator arms.
This is a substantial rework that changes how the Components depend on each other and what they do. 
It also adds features such as reporting feedback to action user.
Adds error handling.
Fixes some errors or undefined situations.
Exposes interfaces allowing users to create develop their own components and controllers more easily.

The code has been tested to run by @michalpelka .

PR #379, opened 2023-07-04
Fix crash after moveit action stop

arturkamieniecki:
I've implemented a fix for the crash when the moveit action stop was called. The fix requires a separate thread that sets the state of the action server after the cancel request was accepted. This is needed because the state of the action server can be changed to ```canceled``` only after the state is ```canceling```.

There was also another error that would stop moving the joints after a cancelation. This resulted in the joints not keeping their positions and the robot arm to slowly falling with gravity. This was also fixed.

This PR addresses #366


adamdbrw:
@arturkamieniecki note that #366 has been merged, please retarget this PR to development

arturkamieniecki:
Rebased and ready for review.

adamdbrw (approved):

Issue #375, opened 2023-06-29
For the Joint Trajectory action server, standardize returned error results.


adamdbrw:
Currently, some ways of returning errors are not fully informative for the action client.
- Review the specification of action servers and `control_msgs::action::FollowJointTrajectory`.
- Correct the flow of results such as from cancellation or invalid goal.
For example, if the goal cannot be reached, first accept it, and then return a failure result. Currently it is rejected with `rclcpp_action::GoalResponse::REJECT,` which does not leave an information as to why.

michalpelka:
Fixed with 386 by @lewandowskikacper .

PR #386, opened 2023-07-06
Error messages in Joint Trajectory

lewandowskikacper:
Invalid goals are now accepted, and later aborted with error message explaining the reason.  Addresses #375.

### DIFF
--- a/Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
@@ -137,7 +137,7 @@ namespace ROS2
         {
             AZ_Trace(
                 "FollowJointTrajectoryActionServer",
-                "Execution not be accepted: %s",
+                "Execution was not accepted: %s",
                 executionOrderOutcome.GetError().error_string.c_str());
 
             auto result = std::make_shared<FollowJointTrajectory::Result>(executionOrderOutcome.GetError());
--- a/Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
@@ -80,7 +80,7 @@ namespace ROS2
     }
 
     rclcpp_action::GoalResponse FollowJointTrajectoryActionServer::GoalReceivedCallback(
-        [[maybe_unused]] const rclcpp_action::GoalUUID& uuid, std::shared_ptr<const FollowJointTrajectory::Goal> goal)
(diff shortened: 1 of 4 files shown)


Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
```diff
@@ -132,6 +108,42 @@ namespace ROS2
...
+        JointsTrajectoryRequestBus::EventResult(
+            executionOrderOutcome, m_entityId, &JointsTrajectoryRequests::StartTrajectoryGoal, goalHandle->get_goal());
+
+        if (!executionOrderOutcome)
+        {
+            AZ_Trace("FollowJointTrajectoryActionServer", "Execution not be accepted: %s", executionOrderOutcome.GetError().c_str());
+
+            auto result = std::make_shared<FollowJointTrajectory::Result>();
+            result->error_string = "Execution not accepted: " + std::string(executionOrderOutcome.GetError().c_str());
+            result->error_code = FollowJointTrajectory::Result::INVALID_GOAL;
```
adamdbrw, line 142: It is better to propagate result from the outcome -> here, the correct result would actually be INVALID_JOINTS (since the call earlier only validates that), but it should not depend on magic knowledge of what is happening.

adamdbrw (commented):
Looks good, one comment about result passed in error_code, take a look!

Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
lewandowskikacper: I changed StartTrajectoryGoal and TrajectoryGoal to return FollowJointTrajectory::Response in case of failure, as it includes error_code, not only string.

Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
```diff
@@ -132,6 +108,44 @@ namespace ROS2
...
+
+        AZ::Outcome<void, FollowJointTrajectory::Result> executionOrderOutcome;
+        JointsTrajectoryRequestBus::EventResult(
+            executionOrderOutcome, m_entityId, &JointsTrajectoryRequests::StartTrajectoryGoal, goalHandle->get_goal());
+
+        if (!executionOrderOutcome)
+        {
+            AZ_Trace(
+                "FollowJointTrajectoryActionServer",
+                "Execution not be accepted: %s",
```
adamdbrw, line 140:
```suggestion
                "Execution was be accepted: %s",
```

adamdbrw (approved):

pawelbudziszewski (approved):

PR #390, opened 2023-07-07
ROS2: Allow for Lidar-side PointCloud publication and update Noise handling. 

alek-kam-robotec-ai:
This pull request includes the following changes related to this [update of the RGL gem.](https://github.com/RobotecAI/o3de-rgl-gem/pull/5):
 1. Support for lidar side publication
    - configuration of the lidar-side publisher
    - timestamp provision
 2. Updated Noise handling
    - noise disabling
    - 2D lidar noise

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
@@ -11,6 +11,7 @@
 #include <AzCore/EBus/EBus.h>
 #include <AzCore/Math/Transform.h>
 #include <AzCore/Math/Vector3.h>
+#include <ROS2/Communication/QoS.h>
 
 namespace ROS2
 {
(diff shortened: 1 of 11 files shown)


Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
```diff
@@ -173,6 +174,37 @@ namespace ROS2
             AZ_Assert(false, "This Lidar Implementation does not support Max range point addition configuration!");
         }
 
+        //! Enables and configures Point Cloud Publisher.
+        //! If not called no publishing is performed.
+        //! @param topicName Name of the ROS 2 topic the pointcloud is published on.
+        //! @param frameId Id of the ROS 2 frame of the sensor.
+        //! @param qoSPolicy QoS policy of published pointcloud messages.
+        virtual void ConfigurePointCloudPublisher(
```
adamdbrw, line 182: This API & documentation can be confusing since all sensors publish their data.
If this API is not called, contrary to what the second sentence claims, the publishing is still going to happen, just not in the raycaster (but in the component).

Keeping the API independent is nice but we still need to make it clear since most commonly raycaster will be used together with the component.

I think that adding justification (why?) would be good. 

"For some implementations, it might be beneficial to publish internally in the raycaster to avoid data copying. This API enables Raycaster implementations that also handle publishing and provides them with necessary publisher configuration"

Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
```diff
@@ -173,6 +174,37 @@ namespace ROS2
...
+        //! (Time.msg: sec = timestampNanoseconds / 10^9; nanosec = timestampNanoseconds mod 10^9).
+        virtual void UpdatePublisherTimestamp([[maybe_unused]] AZ::u64 timestampNanoseconds)
+        {
+            AZ_Assert(false, "This Lidar Implementation does not support PointCloud publishing!");
+        }
+
+        //! Can the raycaster handle publishing?
+        //! This function should be called after the raycaster has been configured.
+        //! The raycaster may not be able to handle point-cloud publishing in certain configurations (e.g. when the maxPointAddition
+        //! is selected) in which case publishing must be handled somewhere else.
```
adamdbrw, line 201: (e.g. in the Lidar Component)

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -278,36 +321,39 @@ namespace ROS2
             m_visualisationPoints = m_lastScanResults.m_points;
         }
 
-        const auto inverseLidarTM = entityTransform->GetWorldTM().GetInverse();
-        for (auto& point : m_lastScanResults.m_points)
+        if (!m_canRaycasterPublish)
```
adamdbrw, line 324: you can return early instead 
if (m_canRaycasterPublish)
{ // skip publishing since the data will be published by the raycaster
  return;
}

Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
alek-kam-robotec-ai: Thanks for pointing that out! Fixed.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -130,18 +130,21 @@ namespace ROS2
     AZStd::vector<AZStd::string> ROS2LidarSensorComponent::FetchLidarSystemList()
     {
         FetchLidarImplementationFeatures();
+        m_lidarParameters.m_showNoiseConfig = m_lidarSystemFeatures & LidarSystemFeatures::Noise;
```
adamdbrw, line 133: Minor: we are doing this three times, perhaps move it to UpdateShowNoise()

adamdbrw (approved):

adamdbrw:
@pijaro @michalpelka please review

pijaro (approved):
Lgtm

Issue #395, opened 2023-07-10
URDF Importer: Meshes not importing


arturkamieniecki:
When importing a Universal Robot, I've encountered a problem with the URDF Importer on O3DE hash: ```c56e2b987caca1200ed55cdf35c6a6be3391859f```. The meshes do not load properly and the asset processor show an error:
![Screenshot from 2023-07-10 13-27-06](https://github.com/o3de/o3de-extras/assets/130671280/360367c9-d064-44b5-ab03-085b31ebf062)
This file exists on the system, but it seems that the asset processor uses a cache to check the state of a file. Due to it being copied moments before and into a folder ignored by the asset processor the file will not be added to the cache. This creates a problem when the ```AZ::SceneAPI::Events::SceneSerialization::LoadScene``` is called due to it asking the asset processor for the file which does not exist in the cache. This prevents further importing of the mesh file.
![Screenshot from 2023-07-10 15-36-06](https://github.com/o3de/o3de-extras/assets/130671280/0f14f0b5-7114-4eb9-9efa-fdd0139b9ad6)

PR #327, opened 2023-05-12
Add RobotImporterAPI and behavioral bus

michalpelka:
~Please, note that is WIP (some debug code in [FileSelectionPage.cpp](https://github.com/o3de/o3de-extras/compare/development...RobotecAI:o3de-extras:add_urdf_import_bus?expand=1#diff-a5757ec673be2e9b6aefdf10c46f109a694b2b06e37298288a4a11cd3f7afcde)~
The code is now ready for review.

PR #400, opened 2023-07-11
URDFImporter: Wait for meshes to process

arturkamieniecki:
This PR addresses the https://github.com/o3de/o3de-extras/issues/395 issue.

~~This PR is WIP.~~

~~It waits 10 seconds for the assets to be processed.~~

~~The PR in the current state can be used if the above bug is encountered.~~

~~The URDF importer was modified to disable watching assets (this breaks the visuals on resolved meshes page).~~
~~To properly import the URDF after the resolved meshes page is shown for the first time we need to use the back button and then the next button. After that, the meshes should be resolved properrly.~~

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
@@ -11,12 +11,83 @@
 #include <AzCore/Asset/AssetManager.h>
 #include <AzCore/Asset/AssetManagerBus.h>
 #include <AzCore/IO/Path/Path.h>
+#include <AzCore/Settings/SettingsRegistry.h>
 #include <AzCore/StringFunc/StringFunc.h>
 #include <AzCore/std/string/regex.h>
 #include <AzToolsFramework/API/EditorAssetSystemAPI.h>
 
 namespace ROS2
 {
+
+    bool Utils::WaitForAssetsToProcess(const AZStd::unordered_map<AZStd::string, AZ::IO::Path>& sourceAssetsPaths)
+    {
+        bool allAssetProcessed = false;
+        bool assetProcessorFailed = false;
+        auto loopStartTime = AZStd::chrono::system_clock::now();
+
+        AZStd::chrono::seconds assetJobTimeout = AZStd::chrono::seconds(30);
+        auto settingsRegistry = AZ::SettingsRegistry::Get();
+        AZ::s64 loopTimeoutValue;
+        if (settingsRegistry->Get(loopTimeoutValue, "/O3DE/ROS2/RobotImporter/AssetProcessorTimeoutInSeconds"))
+        {
+            assetJobTimeout = AZStd::chrono::seconds(loopTimeoutValue);
+        }
+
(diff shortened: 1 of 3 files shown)


adamdbrw:
@arturkamieniecki could this be worked into an acceptable change?

arturkamieniecki:
> @arturkamieniecki could this be worked into an acceptable change?

Yes. I'm working on it right now. I just wanted to create a PR so that if it was needed something could be used.

arturkamieniecki:
This PR is now ready for review. 
I've added a function that waits for the assets to be processed after being copied. This allows the asset processor to put all relevant files into the cache before further processing is done. This enables to create .assetinfo files without errors.
This was needed because the ```AZ::SceneAPI::Events::SceneSerialization::LoadScene``` event call (used to create .assetinfo files) queried the asset processor for the UUID of the file. This was not possible due to the file being stored in the directory hidden from the asset processor.

The wait for assets function is very similar to the function implemented in the https://github.com/o3de/o3de-extras/pull/327 PR.

This change removes the temporary directory where assets were stored as it is not necessary in this implementation.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -11,13 +11,84 @@
...
+                if (AssetFilePath.empty())
+                {
+                    AZ_Warning("RobotImporterUtils", false, "WaitForAssetsToProcess got an empty filepath ");
+                    continue;
+                }
+                using namespace AzToolsFramework;
+                using namespace AzToolsFramework::AssetSystem;
+                AZ::Outcome<AssetSystem::JobInfoContainer> result = AZ::Failure();
+                AssetSystemJobRequestBus::BroadcastResult(
+                    result, &AssetSystemJobRequestBus::Events::GetAssetJobsInfo, AssetFilePath.String(), true);
```
lumberyard-employee-dm, line 63: The Asset Processor actually has a [CompileAssetsSync](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Asset/AssetSystemBus.h#L191-L199) API that can force any known assets based on the relative path to process synchronously

arturkamieniecki: I'm not sure if this can be applied here as we need to wait for assets that have not been detected yet by the Asset Processor.

adamdbrw (approved):

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
michalpelka: @lumberyard-employee-dm  is right, we should avoid blocking calls here. Since this code is working fix, I created GHI https://github.com/o3de/o3de-extras/issues/420 to test approach with `[CompileAssetsSync](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/Asset/AssetSystemBus.h#L191-L199)`

michalpelka (approved):
Created GH issue to test alternative solution.

PR #403, opened 2023-07-11
Stub asset builder for sdf/urdf/world/xacro files

mbalfour-amzn:
This introduces a stub asset builder for robotics files. With this PR, the asset builder generates a placeholder procedural prefab for each input source file found. The placeholder contains a single empty "Test" entity in it. 

Future PRs should introduce actual export functionality into the asset builder.

This video demonstrates the following:
* A urdf file showing up in the Asset Browser
* Dragging the urdf file into the Editor causes the placeholder procedural prefab to get instantiated and detached, so that loose entities exist in the level. These entities will not change if the source asset changes.
* Expanding the urdf shows the .procprefab output asset.
* Dragging the .procprefab asset into the Editor causes the placeholder procedural prefab to get instantiated but left as a procedural prefab, so that it will refresh itself if the source asset changes.

https://github.com/o3de/o3de-extras/assets/82224783/6f11db96-37dc-45f2-8cc5-af67298d6aee

### DIFF
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -66,8 +66,6 @@ ly_add_target(
             Gem::Atom_Component_DebugCamera.Static
             Gem::StartingPointInput
             Gem::PhysX.Static
-        PRIVATE
-            AZ::AssetBuilderSDK
 )
 
 target_depends_on_ros2_packages(${gem_name}.Static rclcpp builtin_interfaces std_msgs sensor_msgs nav_msgs tf2_ros ackermann_msgs gazebo_msgs)
@@ -135,7 +133,9 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
                 Gem::LmbrCentral.API
                 Gem::PhysX.Editor.Static
                 Gem::${gem_name}.Static
-    )
+            PRIVATE
+                AZ::AssetBuilderSDK
+        )
 
     find_package(urdfdom)
     target_link_libraries(${gem_name}.Editor.Static PUBLIC urdfdom::urdfdom_model)
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -59,7 +59,6 @@ ly_add_target(
             Source/VehicleDynamics
     BUILD_DEPENDENCIES
         PUBLIC
-            AZ::AssetBuilderSDK
             AZ::AzCore
             AZ::AzFramework
             Gem::Atom_RPI.Public
@@ -67,6 +66,8 @@ ly_add_target(
             Gem::Atom_Component_DebugCamera.Static
             Gem::StartingPointInput
             Gem::PhysX.Static
+        PRIVATE
+            AZ::AssetBuilderSDK
 )
 
 target_depends_on_ros2_packages(${gem_name}.Static rclcpp builtin_interfaces std_msgs sensor_msgs nav_msgs tf2_ros ackermann_msgs gazebo_msgs)
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -59,6 +59,7 @@ ly_add_target(
             Source/VehicleDynamics
     BUILD_DEPENDENCIES
         PUBLIC
+            AZ::AssetBuilderSDK
             AZ::AzCore
             AZ::AzFramework
             Gem::Atom_RPI.Public
--- a/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
+++ b/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
@@ -11,6 +11,7 @@
 #include <AzCore/IO/FileIO.h>
 #include <AzCore/IO/IOUtils.h>
 #include <AzCore/Serialization/Json/JsonUtils.h>
+#include <AzCore/Settings/SettingsRegistryVisitorUtils.h>
 #include <AzFramework/StringFunc/StringFunc.h>
 #include <AzToolsFramework/Entity/EntityUtilityComponent.h>
 #include <AzToolsFramework/Prefab/PrefabLoaderScriptingBus.h>
@@ -27,13 +28,7 @@ namespace ROS2
         {
             [[maybe_unused]] constexpr const char* SdfAssetBuilderName = "SdfAssetBuilder";
             constexpr const char* SdfAssetBuilderJobKey = "Sdf Asset Builder";
-            constexpr AZStd::array<const char*,4> SdfAssetSourceExtensions =
-            {
-                "*.sdf",
-                "*.urdf",
-                "*.world",
-                "*.xacro"
-            };
+            constexpr const char* SdfAssetBuilderSupportedFileExtensionsRegistryKey = "/ROS/SdfAssetBuilder/SupportedFileTypeExtensions";
         }
 
     SdfAssetBuilder::SdfAssetBuilder()
@@ -43,11 +38,7 @@ namespace ROS2
         sdfAssetBuilderDescriptor.m_name = SdfAssetBuilderJobKey;
         sdfAssetBuilderDescriptor.m_version = 1; // bump this to rebuild all sdf files
         sdfAssetBuilderDescriptor.m_busId = azrtti_typeid<SdfAssetBuilder>();
-
-        for(const auto& extension : SdfAssetSourceExtensions)
-        {
-            sdfAssetBuilderDescriptor.m_patterns.push_back(AssetBuilderSDK::AssetBuilderPattern(extension, AssetBuilderSDK::AssetBuilderPattern::PatternType::Wildcard));
-        }
+        sdfAssetBuilderDescriptor.m_patterns = GetSupportedBuilderPatterns();
 
         sdfAssetBuilderDescriptor.m_createJobFunction = [this](auto && request, auto && response) 
             { 
@@ -73,6 +64,41 @@ namespace ROS2
         BusDisconnect();
 
         // The AssetBuilderSDK doesn't support deregistration, so there's nothing more to do here.
+    }
+
+    AZStd::vector<AssetBuilderSDK::AssetBuilderPattern> SdfAssetBuilder::GetSupportedBuilderPatterns()
+    {
+        AZStd::vector<AssetBuilderSDK::AssetBuilderPattern> patterns;
+
+        auto settingsRegistry = AZ::SettingsRegistry::Get();
+        if (settingsRegistry == nullptr)
+        {
+            AZ_Error(SdfAssetBuilderName, false, "Settings Registry not found, no sdf file type extensions enabled.");
+            return {};
+        }
+
+        // Visit each supported file type extension and create an asset builder wildcard pattern for it.
+        auto VisitFileTypeExtensions = [&settingsRegistry, &patterns]
+            (const AZ::SettingsRegistryInterface::VisitArgs& visitArgs)
+            {
+                if (AZ::SettingsRegistryInterface::FixedValueString value;
+                    settingsRegistry->Get(value, visitArgs.m_jsonKeyPath))
+                {
+                    // Support both '.sdf' and 'sdf' style entries in the setreg file for robustness.
+                    AZStd::string wildcardPattern = (value[0] == '.')
+                        ? AZStd::string::format("*%s", value.c_str())
+                        : AZStd::string::format("*.%s", value.c_str());
+
+                    patterns.push_back(
+                            AssetBuilderSDK::AssetBuilderPattern(
+                                wildcardPattern, AssetBuilderSDK::AssetBuilderPattern::PatternType::Wildcard));
+                }
+                return AZ::SettingsRegistryInterface::VisitResponse::Continue;
+            };
+        AZ::SettingsRegistryVisitorUtils::VisitArray(*settingsRegistry, VisitFileTypeExtensions, SdfAssetBuilderSupportedFileExtensionsRegistryKey);
+
+        AZ_Warning(SdfAssetBuilderName, !patterns.empty(), "SdfAssetBuilder disabled, no supported file type extensions found.");
+        return patterns;
     }
 
     void SdfAssetBuilder::CreateJobs(
@@ -176,7 +202,7 @@ namespace ROS2
 
         if (prefabTemplateId == AzToolsFramework::Prefab::InvalidTemplateId)
         {
-            AZ_Error("prefab", false, "Could not create a prefab template for entities.");
+            AZ_Error(SdfAssetBuilderName, false, "Could not create a prefab template for entities.");
             return {};
         }
 
@@ -189,7 +215,7 @@ namespace ROS2
 
         if (outcome.IsSuccess() == false)
         {
-            AZ_Error("prefab", false, "Could not create JSON string for template; maybe NaN values in the template?");
+            AZ_Error(SdfAssetBuilderName, false, "Could not create JSON string for template; maybe NaN values in the template?");
             return {};
         }
 
--- a/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
+++ b/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
@@ -127,11 +127,8 @@ namespace ROS2
         // Right now, it just generates an empty default procedural prefab. It still needs to be modified
         // to create a procedural prefab that contains the correct information in it.
 
-        AZStd::string outputFilename = request.m_sourceFile;
-        AzFramework::StringFunc::Path::ReplaceExtension(outputFilename, "procprefab");
-
-        AZStd::string tempAssetOutputPath;
-        AzFramework::StringFunc::Path::ConstructFull(request.m_tempDirPath.c_str(), outputFilename.c_str(), tempAssetOutputPath, true);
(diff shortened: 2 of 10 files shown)


Gems/ROS2/Code/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ ly_add_target(
             Source/VehicleDynamics
     BUILD_DEPENDENCIES
         PUBLIC
+            AZ::AssetBuilderSDK
```
cgalvan, line 62: Nit: Could/should this be a private build dependency?

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSystemComponent.h
```diff
@@ -0,0 +1,39 @@
...
+        static void GetProvidedServices(AZ::ComponentDescriptor::DependencyArrayType& provided);
+        static void GetIncompatibleServices(AZ::ComponentDescriptor::DependencyArrayType& incompatible);
+        static void GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required);
+        static void GetDependentServices(AZ::ComponentDescriptor::DependencyArrayType& dependent);
+
+        //////////////////////////////////////////////////////////////////////////
+        // Component overrides
+        void Activate() override;
+        void Deactivate() override;
+        //////////////////////////////////////////////////////////////////////////
```
cgalvan, line 34: Nit:
```suggestion
        // Component overrides ...
        void Activate() override;
        void Deactivate() override;
```

cgalvan (approved):

Gems/ROS2/Code/CMakeLists.txt
lumberyard-employee-dm: This shouldn't be a dependency on the `ROS2.Static` target, as this should only be part of the Tool applications.

This should be added to the `ROS2.Editor.Static` target

Also this will result in a CMake configure error when host tools aren't available to be built, such as when using the Monolithic Build Permutation

lumberyard-employee-dm (changes requested):

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,186 @@
...
+        {
+            [[maybe_unused]] constexpr const char* SdfAssetBuilderName = "SdfAssetBuilder";
+            constexpr const char* SdfAssetBuilderJobKey = "Sdf Asset Builder";
+            constexpr AZStd::array<const char*,4> SdfAssetSourceExtensions =
+            {
+                "*.sdf",
+                "*.urdf",
+                "*.world",
+                "*.xacro"
+            };
```
lumberyard-employee-dm, line 36: The Scene API has the list of available extensions data driven via the Settings Registry.

There is a [sceneassetimporter.setreg ](https://github.com/o3de/o3de/blob/development/Registry/sceneassetimporter.setreg#L4-L14) file which specifies the valid extensions to use.
That could be useful here to follow that approach.

All that would need to be added is a call to the SettingsRegistryVistUtils VisitArray function that could just read that array for specific settings path.
Here is an example of using that api to read the "external_subdirectories" array (https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Settings/SettingsRegistryMergeUtils.cpp#L1688-L1705)

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+        for(const auto& extension : SdfAssetSourceExtensions)
+        {
+            sdfAssetBuilderDescriptor.m_patterns.push_back(AssetBuilderSDK::AssetBuilderPattern(extension, AssetBuilderSDK::AssetBuilderPattern::PatternType::Wildcard));
+        }
+
+        sdfAssetBuilderDescriptor.m_createJobFunction = AZStd::bind(&SdfAssetBuilder::CreateJobs, this,
+            AZStd::placeholders::_1, AZStd::placeholders::_2);
+        sdfAssetBuilderDescriptor.m_processJobFunction = AZStd::bind(&SdfAssetBuilder::ProcessJob, this,
+            AZStd::placeholders::_1, AZStd::placeholders::_2);
```
lumberyard-employee-dm, line 55: Lambdas should be preferred over AZStd::bind, as it is quicker to compile and call and safer to use as it doens't allow for ignoring placeholder parameters: https://clang.llvm.org/extra/clang-tidy/checks/modernize/avoid-bind.html

Gems/ROS2/Code/CMakeLists.txt
mbalfour-amzn: fixed! Private AssetBuilderSDK dependency and code and dependency are moved to Editor.Static now.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,199 @@
...
+    {
+        // The following code is just a stub while building the SdfAssetBuilder. 
+        // Right now, it just generates an empty default procedural prefab. It still needs to be modified
+        // to create a procedural prefab that contains the correct information in it.
+
+        AZStd::string outputFilename = request.m_sourceFile;
+        AzFramework::StringFunc::Path::ReplaceExtension(outputFilename, "procprefab");
+
+        AZStd::string tempAssetOutputPath;
+        AzFramework::StringFunc::Path::ConstructFull(request.m_tempDirPath.c_str(), outputFilename.c_str(), tempAssetOutputPath, true);
```
lumberyard-employee-dm, line 134: Using the AZ::IO::Path API is preferred over using the AZ::StringFunc::Path API.
Also the `AzFramework::StringFunc` was replaced with the `AZ::StringFunc` [namespace](https://github.com/o3de/o3de/blob/development/Code/Framework/AzFramework/AzFramework/StringFunc/StringFunc.h#L15-L17) and that should be used instead.
However the StringFunc include isn't needed in this file at all
```suggestion
        auto tempAssetOutputPath = AZ::IO::Path(request.m_tempDirPath) / request.m_sourceFile;
        tempAssetOutputPath.ReplaceExtension("procprefab");
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,199 @@
...
+    {
+        // The following code is similar to DefaultProceduralPrefabGroup::CreatePrefabGroupManifestUpdates().
+        // It might be possible to refactor the two to have shared logic.
+        // It's not worth investigating until the SdfAssetBuilder is fully functional though, because the code
+        // is likely to change pretty drastically still.
+
+        // Create a prefab name based on the source file name.
+        AZStd::string prefabTemplateName{ request.m_sourceFile };
+        AzFramework::StringFunc::Path::ReplaceExtension(prefabTemplateName, "procprefab");
+        AZ::StringFunc::Replace(prefabTemplateName, "\\", "/"); // the source folder uses forward slash
```
lumberyard-employee-dm, line 177: The AZ::IO:::Path API can be used for this as well with the `FixedMaxPathStringAsPosix` function combined with the ReplaceExtension function..
```suggestion
        // Create a prefab name based on the source file name.
        AZ::IO::FixedMaxPath prefabTemplateName{ AZ::IO::PathView(request.m_sourceFile).FixedMaxPathStringAsPosix() };
        prefabTemplateName.ReplaceExtension("procprefab");
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,199 @@
...
+        // Convert the prefab to a JSON string
+        AZ::Outcome<AZStd::string, void> outcome;
+        AzToolsFramework::Prefab::PrefabLoaderScriptingBus::BroadcastResult(
+            outcome,
+            &AzToolsFramework::Prefab::PrefabLoaderScriptingBus::Events::SaveTemplateToString,
+            prefabTemplateId);
+
+        if (outcome.IsSuccess() == false)
+        {
+            AZ_Error("prefab", false, "Could not create JSON string for template; maybe NaN values in the template?");
```
lumberyard-employee-dm: Nan values?

I believe this comment should be updated to account for SDFormat builder
```suggestion
            AZ_Error("SDFormat Builder", false, "Could not serialize prefab template as a JSON string");
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSystemComponent.h
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include <AzCore/Serialization/SerializeContext.h>
```
lumberyard-employee-dm, line 10: The Serialize Context include is not needed here.
Only an include of Component.h

The `AZ::ReflectContext` class can be forward declared.
```suggestion
#include <AzCore/Component/Component.h>

namespace AZ
{
    class ReflectContext;
}
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,199 @@
...
+        if (prefabJson.empty())
+        {
+            response.m_resultCode = AssetBuilderSDK::ProcessJobResult_Failed;
+            return;
+        }
+
+        // Save the prefab to our output directory.
+        rapidjson::Document doc;
+        doc.Parse(prefabJson.c_str());
+        AZ::JsonSerializationUtils::WriteJsonFile(doc, tempAssetOutputPath.c_str());
```
lumberyard-employee-dm, line 148: Shouldn't the `WriteJsonFile` outcome be checked for errors and the error message logged from it?

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
mbalfour-amzn: Fixed!

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
mbalfour-amzn: Fixed! Moved to an sdfassetbuilder_settings.setreg file and verified that it was getting applied correctly.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,199 @@
...
+            prefabTemplateName);
+
+        if (prefabTemplateId == AzToolsFramework::Prefab::InvalidTemplateId)
+        {
+            AZ_Error("prefab", false, "Could not create a prefab template for entities.");
+            return {};
+        }
+
+        // Convert the prefab to a JSON string
+        AZ::Outcome<AZStd::string, void> outcome;
```
lumberyard-employee-dm, line 210: Not for PR author: Well this is one of the most useless outcomes that exist.
The `PrefabLoaderScriptingBus::SaveTemplateToString` function doesn't return a good value.
Using `void` for the error type negates the entire purpose using an AZ::Outcome.
It would be better if it returned an Error string for the error type and barring that returning an `AZStd::optional<AZStd::string>` to be better express the intent.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include <AssetBuilderSDK/AssetBuilderSDK.h>
+#include <AssetBuilderSDK/SerializationDependencies.h>
+
+namespace ROS2
+{
+        namespace
+        {
+            [[maybe_unused]] constexpr const char* SdfAssetBuilderName = "SdfAssetBuilder";
+            constexpr const char* SdfAssetBuilderJobKey = "Sdf Asset Builder";
+            constexpr const char* SdfAssetBuilderSupportedFileExtensionsRegistryKey = "/ROS/SdfAssetBuilder/SupportedFileTypeExtensions";
```
lumberyard-employee-dm, line 31: Should we place this key under `/O3DE/ROS2` instead?

Any insight @adamdbrw on how ROS2 settings are catagorized?
```suggestion
            constexpr const char* SdfAssetBuilderSupportedFileExtensionsRegistryKey = "/O3DE/ROS2/SdfAssetBuilder/SupportedFileTypeExtensions";
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -0,0 +1,225 @@
...
+        auto VisitFileTypeExtensions = [&settingsRegistry, &patterns]
+            (const AZ::SettingsRegistryInterface::VisitArgs& visitArgs)
+            {
+                if (AZ::SettingsRegistryInterface::FixedValueString value;
+                    settingsRegistry->Get(value, visitArgs.m_jsonKeyPath))
+                {
+                    // Support both '.sdf' and 'sdf' style entries in the setreg file for robustness.
+                    AZStd::string wildcardPattern = (value[0] == '.')
+                        ? AZStd::string::format("*%s", value.c_str())
+                        : AZStd::string::format("*.%s", value.c_str());
```
lumberyard-employee-dm, line 90: To protect against an invalid access when the string is empty, the [AZStd::string::starts_with](https://en.cppreference.com/w/cpp/string/basic_string/starts_with) function should be used instead which protects the string being empty.
```suggestion
                    // Support both '.sdf' and 'sdf' style entries in the setreg file for robustness.
                    AZStd::string wildcardPattern = value.starts_with('.')
                        ? value
                        : AZStd::string::format("*.%s", value.c_str());
```

adamdbrw (commented):
Well done, this also improves the UX.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
mbalfour-amzn: Makes sense, though the proposed fix isn't correct because if the value starts with `.` we still need to prepend a `*`. And to your point, if the string is empty or only contains a `.`, we probably shouldn't create a wildcard pattern at all. I'll make this a bit more robust.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
mbalfour-amzn: If ROS2 or Robotics both sound good, I'll go with the proposed /O3DE/ROS2 since that matches up with the Gem name.

Gems/ROS2/Registry/sdfassetbuilder_settings.setreg
```diff
@@ -0,0 +1,15 @@
...
+            "SupportedFileTypeExtensions":
+            [
+                "sdf",
+                "urdf",
+                "world",
+                "xacro"
+            ]
+        }
+    }
+}
```
mbalfour-amzn, line 15:
```suggestion
{
    "O3DE":
    {
        "ROS2":
        {
            "SdfAssetBuilder":
            {
                "SupportedFileTypeExtensions":
                [
                    "sdf",
                    "urdf",
                    "world",
                    "xacro"
                ]
            }
        }
    }
}
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
mbalfour-amzn: Lol, yeah. I already changed the error tag to SdfAssetBuilderName, but I wasn't exactly sure why the original error message from `DefaultProceduralPrefabGroup::CreatePrefabGroupManifestUpdates()` was warning about NaNs so I left it there. I'll change the message to your proposed message though, it makes more sense. :)

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
mbalfour-amzn: Technically yes, but I'm expecting this part of the code to change pretty drastically over the next couple of PRs. I thought it was pretty likely that there would be different / better ways to write out prefabs than this, so I expected this to just be a stub. I'll add the error message logging anyways though, it's always a good idea to keep each PR as correct as possible. :)

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
mbalfour-amzn: Yeah, I have no explanation for why the Json/Prefab code uses Outcome<> instead of optional<> for these. My best guess is that it was a stylistic choice to keep API consistency with the other Json/Prefab APIs that use Outcome<>. 

Doing a quick search on Outcome<> in the engine, probably 90% of the uses could be replaced with optional<>. Most of them were probably written before C++ optional<> support was available.

lumberyard-employee-dm (approved):

Issue #410, opened 2023-07-13
Higher frequency of publication of JointStates


michalpelka:
Currently, the JointState publication rate is limited by the framerate of the Game/Editor.
To my knowledge, the data used to compute all necessary data is available on every tick of the physics engine.
Please change the mechanism of ticking of `JointStatePublisher`. 
Utilize `ROS2::Utils::PhysicsCallbackHandler`.
With this change, the frequency of JointState publication will be limited by the frequency of the PhysX engine (which is an order of magnitude higher than the framerate, in most of robotics simulations).

Issue #295, opened 2023-04-24
Refactor ROS2Sensor component data acquisition loops


adamdbrw:
**Is your feature request related to a problem? Please describe.**
This is an enhancement to the architecture of current data acquisition mechanism in ROS2Sensor component as introduced with #254.

**Describe the solution you'd like**
I would like to increase encapsulation and reduce the amount of code that the implementer of ROS2Sensor subclass needs to be aware of in order to succeed. This involves replacing inheritance with composition where needed and implementing NVI for the loops if applicable.

**Additional context**
The dual mechanism is here because one (onTick()) is simpler and refers to Engine updates, and the second one (physics events) provides a higher-frequency, less-jittery data streams important for sensors such as IMU and Odometry.

Antoni-Robotec:
The sensors update mechanism is similar to the ones in ROS Publishers, would be best to unify.

adamdbrw:
This means, when working on this task, also look at non-sensor publishers such as JointState.

PR #412, opened 2023-07-14
Increase max publishing frequency of joint_states messages

Antoni-Robotec:
Solves issue https://github.com/o3de/o3de-extras/issues/410

The solution uses one bus request per joint. This has a performance impact, but it's fairly small for publishers with a reasonable number of joints (for 9 joints about 90% of the performance impact is caused by the ros2 publishing itself). The performance can be further improved by using JointHandles and SimulatedBodyHandles instead of buses, but the code for it got so hideous I gave up on it and decided that this is the better solution after all.

### DIFF
--- a/Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
+++ b/Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
@@ -73,7 +73,7 @@ namespace ROS2
 
     void ROS2ImuSensorComponent::GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required)
     {
-        required.push_back(AZ_CRC_CE("PhysicsRigidBodyService"));
+        required.push_back(AZ_CRC_CE("PhysicsDynamicRigidBodyService"));
         required.push_back(AZ_CRC_CE("ROS2Frame"));
     }
 
--- a/Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
+++ b/Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
@@ -79,7 +79,17 @@ namespace ROS2
 
     void ROS2ImuSensorComponent::SetupRefreshLoop()
     {
-        InstallPhysicalCallback(m_entity->GetId());
+        InstallPhysicalCallback();
+    }
+
+    void ROS2ImuSensorComponent::OnPhysicsInitialization(AzPhysics::SceneHandle sceneHandle)
+    {
+        AzPhysics::RigidBody* rigidBody = nullptr;
+        AZ::EntityId entityId = GetEntityId();
+        Physics::RigidBodyRequestBus::EventResult(rigidBody, entityId, &Physics::RigidBodyRequests::GetRigidBody);
+        AZ_Assert(rigidBody, "Entity %s does not have rigid body.", entityId.ToString().c_str());
+
+        m_bodyHandle = rigidBody->m_bodyHandle;
     }
 
     void ROS2ImuSensorComponent::OnPhysicsSimulationFinished(AzPhysics::SceneHandle sceneHandle, float deltaTime)
--- a/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
@@ -31,13 +31,15 @@ namespace ROS2
         rosHeader.stamp = ROS2::ROS2Interface::Get()->GetROSTimestamp();
         m_jointStateMsg.header = rosHeader;
 
-        for (size_t i = 0; i < m_jointNames.size(); i++)
+        AZ_Assert(m_jointNames.size() == m_jointStateMsg.name.size(), "The expected message size doesn't match with the joint list size");
+
+        for (size_t i = 0; i < m_jointStateMsg.name.size(); i++)
         {
             m_jointStateMsg.name[i] = m_jointNames[i].c_str();
             JointInfo& jointInfo = m_jointInfos[i];
 
             auto jointStateData = Utils::GetJointState(jointInfo);
-            
+
             m_jointStateMsg.position[i] = jointStateData.position;
             m_jointStateMsg.velocity[i] = jointStateData.velocity;
             m_jointStateMsg.effort[i] = jointStateData.effort;
--- a/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
@@ -52,7 +52,7 @@ namespace ROS2
         ManipulationJoints manipulatorJoints;
         JointsManipulationRequestBus::EventResult(manipulatorJoints, m_context.m_entityId, &JointsManipulationRequests::GetJoints);
 
-        for (auto& [jointName, jointInfo] : manipulatorJoints)
+        for (const auto& [jointName, jointInfo] : manipulatorJoints)
         {
             m_jointNames.push_back(jointName);
             m_jointInfos.push_back(jointInfo);
@@ -66,7 +66,7 @@ namespace ROS2
         InstallPhysicalCallback();
     }
 
-    void JointStatePublisher::OnPhysicsSimulationFinished(AzPhysics::SceneHandle sceneHandle [[maybe_unused]], float deltaTime)
+    void JointStatePublisher::OnPhysicsSimulationFinished([[maybe_unused]] AzPhysics::SceneHandle sceneHandle, float deltaTime)
     {
         AZ_Assert(m_configuration.m_frequency > 0.f, "JointPublisher frequency must be greater than zero");
         auto frameTime = 1.f / m_configuration.m_frequency;
--- a/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
@@ -7,6 +7,7 @@
  */
 
 #include "JointStatePublisher.h"
+#include "ManipulationUtils.h"
 #include <ROS2/Manipulation/JointsManipulationRequests.h>
 #include <ROS2/ROS2Bus.h>
 #include <ROS2/Utilities/ROS2Names.h>
@@ -30,36 +31,40 @@ namespace ROS2
         rosHeader.stamp = ROS2::ROS2Interface::Get()->GetROSTimestamp();
         m_jointStateMsg.header = rosHeader;
 
+        for (size_t i = 0; i < m_jointNames.size(); i++)
+        {
+            m_jointStateMsg.name[i] = m_jointNames[i].c_str();
+            JointInfo& jointInfo = m_jointInfos[i];
+
+            auto jointStateData = Utils::GetJointState(jointInfo);
+            
+            m_jointStateMsg.position[i] = jointStateData.position;
+            m_jointStateMsg.velocity[i] = jointStateData.velocity;
+            m_jointStateMsg.effort[i] = jointStateData.effort;
+        }
+        m_jointStatePublisher->publish(m_jointStateMsg);
+    }
+
+    void JointStatePublisher::InitializePublisher(AZ::EntityId entityId)
+    {
         ManipulationJoints manipulatorJoints;
         JointsManipulationRequestBus::EventResult(manipulatorJoints, m_context.m_entityId, &JointsManipulationRequests::GetJoints);
+
+        for (auto& [jointName, jointInfo] : manipulatorJoints)
+        {
+            m_jointNames.push_back(jointName);
+            m_jointInfos.push_back(jointInfo);
+        }
 
         m_jointStateMsg.name.resize(manipulatorJoints.size());
         m_jointStateMsg.position.resize(manipulatorJoints.size());
         m_jointStateMsg.velocity.resize(manipulatorJoints.size());
         m_jointStateMsg.effort.resize(manipulatorJoints.size());
-        size_t i = 0;
-        for (const auto& [jointName, jointInfo] : manipulatorJoints)
-        {
-            AZ::Outcome<float, AZStd::string> result;
-            JointsManipulationRequestBus::EventResult(
-                result, m_context.m_entityId, &JointsManipulationRequests::GetJointPosition, jointName);
-            auto currentJointPosition = result.GetValue();
-            JointsManipulationRequestBus::EventResult(
-                result, m_context.m_entityId, &JointsManipulationRequests::GetJointVelocity, jointName);
-            auto currentJointVelocity = result.GetValue();
-            JointsManipulationRequestBus::EventResult(result, m_context.m_entityId, &JointsManipulationRequests::GetJointEffort, jointName);
-            auto currentJointEffort = result.GetValue();
 
-            m_jointStateMsg.name[i] = jointName.c_str();
-            m_jointStateMsg.position[i] = currentJointPosition;
-            m_jointStateMsg.velocity[i] = currentJointVelocity;
-            m_jointStateMsg.effort[i] = currentJointEffort;
-            i++;
-        }
-        m_jointStatePublisher->publish(m_jointStateMsg);
+        InstallPhysicalCallback();
     }
 
-    void JointStatePublisher::OnTick(float deltaTime)
+    void JointStatePublisher::OnPhysicsSimulationFinished(AzPhysics::SceneHandle sceneHandle [[maybe_unused]], float deltaTime)
     {
         AZ_Assert(m_configuration.m_frequency > 0.f, "JointPublisher frequency must be greater than zero");
         auto frameTime = 1.f / m_configuration.m_frequency;
--- a/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.h
+++ b/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.h
@@ -11,7 +11,7 @@
 #include <AzCore/Component/EntityId.h>
 #include <ROS2/Communication/PublisherConfiguration.h>
 #include <ROS2/Manipulation/JointInfo.h>
-#include <Utilities/PhysicsCallbackHandler.h>
+#include <ROS2/Utilities/PhysicsCallbackHandler.h>
 #include <rclcpp/publisher.hpp>
 #include <sensor_msgs/msg/joint_state.hpp>
 
--- a/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.h
+++ b/Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.h
@@ -10,6 +10,8 @@
 
 #include <AzCore/Component/EntityId.h>
 #include <ROS2/Communication/PublisherConfiguration.h>
+#include <ROS2/Manipulation/JointInfo.h>
+#include <Utilities/PhysicsCallbackHandler.h>
 #include <rclcpp/publisher.hpp>
 #include <sensor_msgs/msg/joint_state.hpp>
 
@@ -24,13 +26,13 @@ namespace ROS2
 
     //! A class responsible for publishing the joint positions on ROS2 /joint_states topic.
     //!< @see <a href="https://docs.ros2.org/latest/api/sensor_msgs/msg/JointState.html">jointState message</a>.
-    class JointStatePublisher
+    class JointStatePublisher : public ROS2::Utils::PhysicsCallbackHandler
     {
     public:
         JointStatePublisher(const PublisherConfiguration& configuration, const JointStatePublisherContext& context);
+        virtual ~JointStatePublisher() = default;
 
-        //! Update time tick. This will result in state publishing if timing matches frequency.
-        void OnTick(float deltaTime);
+        void InitializePublisher(AZ::EntityId entityId);
 
     private:
         void PublishMessage();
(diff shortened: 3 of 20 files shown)


adamdbrw (dismissed):
Similar mechanisms are in use in sensor abstraction. It would be best to unify. Please leave a note about this in #295

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.h
```diff
@@ -25,14 +26,18 @@ namespace ROS2
...
     class JointStatePublisher
+        : public ROS2::Utils::PhysicsCallbackHandler
     {
     public:
         JointStatePublisher(const PublisherConfiguration& configuration, const JointStatePublisherContext& context);
+        virtual ~JointStatePublisher() = default;
 
-        //! Update time tick. This will result in state publishing if timing matches frequency.
-        void OnTick(float deltaTime);
+        void Activate();
```
michalpelka, line 35: Could you document it? It is not a bug but looks like one, or better choose other name since it can be mistaken AZ::Component::Activate.

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
```diff
@@ -102,7 +102,15 @@ namespace ROS2
 
     void ROS2ImuSensorComponent::SetupRefreshLoop()
     {
-        InstallPhysicalCallback(m_entity->GetId());
+        auto entityId = m_entity->GetId();
```
michalpelka, line 105: I do not really like that. A number of sensors require a handler to rigid body / simulated body. The problem with that is:
- the code for taking rigid body handle is duplicated (even in this PR).
- this API is used outside of the project and it is breaking change.

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
```diff
@@ -53,7 +54,12 @@ namespace ROS2
         m_jointStatePublisher->publish(m_jointStateMsg);
     }
 
-    void JointStatePublisher::OnTick(float deltaTime)
+    void JointStatePublisher::Activate()
```
michalpelka, line 57: Activate can be mistaken with Component's activate.

Gems/ROS2/Code/Source/Odometry/ROS2OdometrySensorComponent.cpp
```diff
@@ -58,7 +58,15 @@ namespace ROS2
     // ROS2SensorComponent overrides ...
     void ROS2OdometrySensorComponent::SetupRefreshLoop()
     {
-        InstallPhysicalCallback(m_entity->GetId());
+        auto entityId = m_entity->GetId();
+        AzPhysics::RigidBody* rigidBody = nullptr;
+        Physics::RigidBodyRequestBus::EventResult(rigidBody, entityId, &Physics::RigidBodyRequests::GetRigidBody);
```
michalpelka, line 63: Duplicated code.

Gems/ROS2/Code/Source/Odometry/ROS2OdometrySensorComponent.h
```diff
@@ -46,6 +46,9 @@ namespace ROS2
         nav_msgs::msg::Odometry m_odometryMsg;
         AZ::Transform m_initialTransform;
 
+        // Handle to the simulated physical body
+        AzPhysics::SimulatedBodyHandle m_bodyHandle = AzPhysics::InvalidSimulatedBodyHandle;
```
michalpelka, line 50: The sensor code has unnecessary implementation detail.

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
```diff
@@ -53,7 +54,12 @@ namespace ROS2
...
+    {
+        InstallPhysicalCallback();
+    }
+
+    void JointStatePublisher::OnPhysicsSimulationFinished([[maybe_unused]] AzPhysics::SceneHandle sceneHandle, float deltaTime)
```
michalpelka, line 62: I do not like that. This code will be call 
` JointsManipulationRequestBus::EventResult(
  result, m_context.m_entityId, &JointsManipulationRequests::GetJointPosition, joint name);`
with a very large frequency (1 kHz). It is doing a number of string and map operations that can be expensive, especially with a large number of joints.  Currently, the manipulator structure cannot be changed during simulation, the less elegant but performant solution needs to be initialised here.

Gems/ROS2/Code/Source/Utilities/PhysicsCallbackHandler.cpp
```diff
@@ -11,30 +11,18 @@
 
 namespace ROS2::Utils
 {
-    void PhysicsCallbackHandler::InstallPhysicalCallback(AZ::EntityId entityId)
+    void PhysicsCallbackHandler::InstallPhysicalCallback()
```
michalpelka, line 14: I would propose to make AZ::EntityId entityId optional here to prevent breaking change. 
If the API consumer won't provide entiyId the callback will be installed, but the handle m_bodyHandle will be invalid.

michalpelka (changes requested):
I think this change in this form should not be merged.

michalpelka:
What is the status @Antoni-Robotec ? I can see new code, but my comments are not resolved.

Gems/ROS2/Code/Source/Gripper/FingerGripperComponent.cpp
michalpelka, line 1: Why this code is here? The gripper component is not a apart of this PR.

Gems/ROS2/Code/Source/Gripper/FingerGripperComponent.h
michalpelka, line 1: Why this code is here? The gripper component is not a apart of this PR.

Gems/ROS2/Code/Source/Gripper/GripperActionServer.cpp
michalpelka, line 1: Why this code is here? The gripper component is not a apart of this PR.

michalpelka:
I cannot review, since you committed multiple contributions. Please clear it to one contribution and rebase.

Antoni-Robotec:
> 

Sorry, this not ready for review yet, WIP. The changes are here because I did a rebase (needed a helper function from another PR).

Antoni-Robotec:
The code still uses AZ:Bus, but only once for each joint. I've also removed unordered maps which were the biggest performance issue. The publisher could be further optimized by keeping handles to the simulated bodies in case of articulation links and to joints in case of classical joints, as that'd remove the need for any bus usage, however it's not so simple to do so with the code being of acceptable quality.

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.cpp
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#include "ManipulationUtils.h"
+
+namespace ROS2::Utils
+{
+    JointStateData GetJointState(const JointInfo& jointInfo)
+    {
+        JointStateData result;
+        float position, velocity, effort;
```
adamdbrw, line 16: unnecessary to declare these - just assign straight to result

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+#include <ROS2/Manipulation/JointInfo.h>
+
+namespace ROS2::Utils
+{
+    struct JointStateData
+    {
+        float position;
+        float velocity;
+        float effort;
```
adamdbrw, line 18: default to zeros

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
```diff
@@ -30,36 +31,40 @@ namespace ROS2
         rosHeader.stamp = ROS2::ROS2Interface::Get()->GetROSTimestamp();
         m_jointStateMsg.header = rosHeader;
 
+        for (size_t i = 0; i < m_jointNames.size(); i++)
```
adamdbrw, line 34: Good practice - iterate over container you write to instead of data source.
Add a check whether the joint names matches the message name vector size.

Gems/ROS2/Code/Source/Utilities/PhysicsCallbackHandler.cpp
adamdbrw: I suggest to have only 1 InstallPhysicalCallback which has no arguments.
1. Use RequiredServices in implementing classes first and foremost, if possible, to define requirements for the entity. 
2. If requirements are not simple or static, implement OnPhysicsInitialization to check whether our Entity meets constraints. Allow it to fail and disconnect callback & inform user through error log. You can acquire enityId in implementation body.
3. OnPhysicsInitialization will also cache all bus calls / getters so that it is not repeated in OnPhysicsSimulationFinished.
4. Document accordingly, including how a new sensor component may want to derive the class AND ROS2SensorComponent.
5. Expose PhysicsCallbackHandler as an interface for developers.

Please also correct m_entityId passed as function argument - this is wrong naming convention for a function argument.

If so, adjust IMU / Odometry sensor - overriding OnPhysicsInitialization.

@michalpelka if you have other design in mind lets discuss.

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.cpp
```diff
@@ -0,0 +1,61 @@
...
+                    is_acceleration_driven = articulationJointRequests->IsAccelerationDrive(jointInfo.m_axis);
+                });
+            if (!is_acceleration_driven)
+            {
+                effort = stiffness * -(position - target_position) + damping * (target_velocity - velocity);
+                effort = AZ::GetClamp(effort, -max_effort, max_effort);
+            }
+            else
+            {
+                effort = 0.f;
```
adamdbrw, line 42: entirely skippable if defaults are used

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.cpp
```diff
@@ -0,0 +1,61 @@
...
+                    stiffness = articulationJointRequests->GetDriveStiffness(jointInfo.m_axis);
+                    damping = articulationJointRequests->GetDriveDamping(jointInfo.m_axis);
+                    target_position = articulationJointRequests->GetDriveTarget(jointInfo.m_axis);
+                    position = articulationJointRequests->GetJointPosition(jointInfo.m_axis);
+                    target_velocity = articulationJointRequests->GetDriveTargetVelocity(jointInfo.m_axis);
+                    velocity = articulationJointRequests->GetJointVelocity(jointInfo.m_axis);
+                    max_effort = articulationJointRequests->GetMaxForce(jointInfo.m_axis);
+                    is_acceleration_driven = articulationJointRequests->IsAccelerationDrive(jointInfo.m_axis);
+                });
+            if (!is_acceleration_driven)
```
adamdbrw, line 35: best move this inside the lambda since nothing happens in the lambda doesn't execute anyway.

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.cpp
```diff
@@ -0,0 +1,61 @@
...
+        else
+        {
+            PhysX::JointRequestBus::Event(
+                jointInfo.m_entityComponentIdPair,
+                [&](PhysX::JointRequests* jointRequests)
+                {
+                    position = jointRequests->GetPosition();
+                    velocity = jointRequests->GetVelocity();
+                });
+            effort = 0.f;
```
adamdbrw, line 54: same, skip

Gems/ROS2/Code/Source/Utilities/PhysicsCallbackHandler.cpp
michalpelka: @adamdbrw totally agree.
- remove 'AzPhysics::SimulatedBodyHandle m_bodyHandle' from PhysicsCallbackHandler.
- add virtual, empty `PhysicsCallbackHandler::OnPhysicsInitialization` whiteout parameters.
- To implement "fast physical sensor" the Developer need to:
  - derive ROS2SensorComponent
  - derive PhysicsCallbackHandler
  - implement `PhysicsCallbackHandler::OnPhysicsInitialization` where she or he will cache all needed pointers handlers and use them in implementation PhysicsCallbackHandler::OnPhysicsInitialization.
  
I would only propose to add utility function / static member to easily get something like `AZStd::unorderedmap<AZ::EntityId,AzPhysics::SimulatedBodyHandle>` from given Entity eg : ` AZStd::unorderedmap<AZ::EntityId,AzPhysics::SimulatedBodyHandle> getHandlesFromEntity(const AZ::EntityId &id);` - to limit redundant code in `PhysicsCallbackHandler::OnPhysicsInitialization`

Antoni-Robotec: But currently the ROS2SensorComponent doesn't do much other than providing a mechanism for frame based publication (so it is only really useful for slow sensors). This is related to https://github.com/o3de/o3de-extras/issues/295.
Maybe ROS2SensorComponent should derive from PhysicsCallbackHandler and always use the physics engine updates? Maybe it should just be called ROS2RegularPublisher?

Antoni-Robotec: Discussed, will implement Michał's solution

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.cpp
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "ManipulationUtils.h"
+#include <PhysX/Joint/PhysXJointRequestsBus.h>
```
michalpelka, line 10:
```suggestion
#include <PhysX/Joint/PhysXJointRequestsBus.h>
#include <PhysX/ArticulationJointBus.h>
```

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
```diff
@@ -73,13 +73,23 @@ namespace ROS2
 
     void ROS2ImuSensorComponent::GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required)
     {
-        required.push_back(AZ_CRC_CE("PhysicsRigidBodyService"));
+        required.push_back(AZ_CRC_CE("PhysicsDynamicRigidBodyService"));
```
michalpelka, line 76: Good catch :+1:

michalpelka (commented):
Looks, good, one header is missing in non-unity build.

michalpelka (approved):
LGTM

adamdbrw:
@PawelLiberadzki please review

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+#include <ROS2/Manipulation/JointInfo.h>
+
+namespace ROS2::Utils
+{
+    struct JointStateData
+    {
+        float position{ 0.f };
+        float velocity{ 0.f };
+        float effort{ 0.f };
```
PawelLiberadzki, line 18: In Inlude/ROS2/ManipulationJointInfo.h there are aliases declared for joint position, velocity and effort (they are still float). They can technically be used here, but in the same moment they seem to be not used in the whole project. I would either update this code here or remove those aliases.

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.cpp
```diff
@@ -0,0 +1,52 @@
...
+                    result.position = articulationJointRequests->GetJointPosition(jointInfo.m_axis);
+                    result.velocity = articulationJointRequests->GetJointVelocity(jointInfo.m_axis);
+                    bool is_acceleration_driven = articulationJointRequests->IsAccelerationDrive(jointInfo.m_axis);
+                    if (!is_acceleration_driven)
+                    {
+                        float stiffness = articulationJointRequests->GetDriveStiffness(jointInfo.m_axis);
+                        float damping = articulationJointRequests->GetDriveDamping(jointInfo.m_axis);
+                        float target_position = articulationJointRequests->GetDriveTarget(jointInfo.m_axis);
+                        float target_velocity = articulationJointRequests->GetDriveTargetVelocity(jointInfo.m_axis);
+                        float max_effort = articulationJointRequests->GetMaxForce(jointInfo.m_axis);
```
PawelLiberadzki, line 34: These three variables above seem to break naming convention -> CamelCase.

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.cpp
```diff
@@ -0,0 +1,52 @@
...
+            PhysX::ArticulationJointRequestBus::Event(
+                jointInfo.m_entityComponentIdPair.GetEntityId(),
+                [&](PhysX::ArticulationJointRequests* articulationJointRequests)
+                {
+                    result.position = articulationJointRequests->GetJointPosition(jointInfo.m_axis);
+                    result.velocity = articulationJointRequests->GetJointVelocity(jointInfo.m_axis);
+                    bool is_acceleration_driven = articulationJointRequests->IsAccelerationDrive(jointInfo.m_axis);
+                    if (!is_acceleration_driven)
+                    {
+                        float stiffness = articulationJointRequests->GetDriveStiffness(jointInfo.m_axis);
```
PawelLiberadzki, line 30: A lot of variables in this scope can be made const -> may affect compiler a little bit in terms of optimization.

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
```diff
@@ -30,36 +31,42 @@ namespace ROS2
...
-            m_jointStateMsg.velocity[i] = currentJointVelocity;
-            m_jointStateMsg.effort[i] = currentJointEffort;
-            i++;
-        }
-        m_jointStatePublisher->publish(m_jointStateMsg);
+        InstallPhysicalCallback();
     }
 
-    void JointStatePublisher::OnTick(float deltaTime)
+    void JointStatePublisher::OnPhysicsSimulationFinished(AzPhysics::SceneHandle sceneHandle [[maybe_unused]], float deltaTime)
```
PawelLiberadzki, line 69: Just for readability, I would move maybe_unused attribute to the left side of SceneHandle type - this is how it is used in rest of the project.

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
PawelLiberadzki, line 1: I am not sure, if I can address these in this pull request, however:

Line 83 in this file - is there maybe some way to notifiy user about this? I can be not aware, how serious this problem could be and if it is worth e.g. warning spam?
+
If-statement from line 81 looks like it can be moved earlier, right after line 72 - this would save few operations.

Gems/ROS2/Code/Source/Utilities/ArticulationsUtilities.cpp
```diff
@@ -52,4 +53,24 @@ namespace ROS2::Utils
...
     }
+
+    AZStd::unordered_map<AZ::EntityId, AzPhysics::SimulatedBodyHandle> GetSimulatedBodyHandles(AzPhysics::SceneHandle sceneHandle, AZ::EntityId entityId)
+    {
+        AZ::EntityId rootArticulationEntity = GetRootOfArticulation(entityId);
+        AZ::Entity* rootEntity = nullptr;
+        AZ::ComponentApplicationBus::BroadcastResult(rootEntity, &AZ::ComponentApplicationRequests::FindEntity, rootArticulationEntity);
+
+        PhysX::ArticulationLinkComponent* component = rootEntity->FindComponent<PhysX::ArticulationLinkComponent>();
+        AZStd::vector<AzPhysics::SimulatedBodyHandle> articulationHandles = component->GetSimulatedBodyHandles();
```
PawelLiberadzki, line 64: I believe, this and articulationHandle from line 67 can be made const.

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
Antoni-Robotec: I think a single warning may be justified (with some bool keeping track whether it was already sent). If we decide on it, we should also add it to sensors.

As for the if statement, I think it's unnecessary optimization, the impact would be very minimal and it'd lead to worse code (we can't just return in the if statement, we'd need to call the Publish method).

Also, please next time divide such comments into 2 separate ones, made at the right place instead of at the start of the file :)

Gems/ROS2/Code/Source/Manipulation/ManipulationUtils.h
Antoni-Robotec: I was considering using them, but decided against it since as you've mentioned, they're not used elsewhere. It's true that their usage is inconsistent and should somehow be unified.

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
PawelLiberadzki: I would personally go for this single warning - overhead with additional if-statement (well defined with one path always more likely) should be acceptable trade-off between performance (runtime test?) and debugability.

Regarding this minor optimization - you are right that this will not be a huge performance boost, however this could be some performance boost. Depending on the common practices here, we can work on it, discuss it or ignore it. I would at least discuss this and then decide.

PawelLiberadzki (approved):

adamdbrw (approved):

Issue #425, opened 2023-07-21
Ros2ProjectTemplate warning: json replace is not valid


jhanca-robotecai:
An issue in **Ros2ProjectTemplate** - the project created from the template causes warnings related to json replace while opening

![image](https://github.com/o3de/o3de-extras/assets/134940295/0872b04a-d030-4d07-8ab0-1715fa137a75)

o3de (develop from 6th of June): 02ec68a8
o3de-extras development branch: 17ddb04d467d82004f58b0bf0ca05d8bc7a30676

PR #427, opened 2023-07-24
Fix warnings related to json replace while opening

aniafafe:
Fixes: #425 
The problem was with the 31st patch replace operation in _ROSBot_slamtec.prefab_, because of an invalid argument value.
Deleting the patch solved the problem.


adamdbrw:
@aniafafe Could you explain shortly what the issue was and how this fixes it? Thank you!

michalpelka (approved):
LGTM!

adamdbrw (approved):

PR #426, opened 2023-07-21
Adding the sdformat13 3rdParty library as a dependency the `ROS2.Editor.Static` TARGET

lumberyard-employee-dm:
An `ly_associate_package` call has been added to allow the sdformat13 3rdParty library to be downloaded during CMake configure when building on Linux.

Both a x86_64 and Aarch64 package has been built.

NOTE: These packages have not been uploaded to the O3DE 3rdParty package repo and therefore this change cannot go in until then.

It is posted for review to streamline the process when the package is approved to be uploaded.

PR #429, opened 2023-07-25
Add SDFormat <sensor> tag support

jhanca-robotecai:
This is a **work in progress** branch with multiple changes that will be cherry-picked into respective PRs when possible.

Some changes, that might be obsolete:
* hardcoded `libsdformat` import (to be removed after a proper import is done, #426)
* added `RobotImporter/SDFormat/Parser.h` which uses `libsdformat` to parse data (xml file / string to `libsdformat` structure)
* added _SDFormat_ models for tests in `Tests/SdfModel.h` alongside with self-test/test_zero in `Tests/SdfModelTest.cpp`
* added simple methods to list `<sensor>` tags and `<plugin>` tags (with some tests)

Additionally, this PR partially implements the idea first mentioned by Derric McGarrah, in which O3DE components are dynamically mapped to _SDFormat_ `<sensor>` tag by adding a hook in the _Reflect_ method. Next, available O3DE components can be retrieved while importing SDFormat_ model. This part is not implemented yet, but the prototype for the hook and the partial implementation for `<sensor><camera>` to `ROS2CameraSensorComponent` is ready.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/SDFormat/SensorsMaker.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/SDFormat/SensorsMaker.cpp
@@ -7,22 +7,24 @@
  */
 
 #include "SensorsMaker.h"
+#include "SupportedTags.h"
 
 #include <AzCore/Component/Component.h>
 #include <AzToolsFramework/Entity/EditorEntityHelpers.h>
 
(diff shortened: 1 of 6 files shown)


jhanca-robotecai:
This PR is reopened in https://github.com/o3de/o3de-extras/pull/437

Issue #392, opened 2023-07-07
URDF Importer: Single xacro file with multiple configurations


arturkamieniecki:
When importing a robot from an xacro file with a configuration parameter the URDF importer does not take into consideration that the provided robot is different. 

Following the instructions at: https://docs.o3de.org/docs/user-guide/interactivity/robotics/importing-robot/. By changing parameters, we can import different kinds of robots using one file. This process will always create a folder to store meshes in Assets/UrdfImporter. The folder's name depends on the selected file, not the parameters. This will create a problem when importing a different robot with the same file but different arguments, the URDF Importer will use the same meshes, which can create broken prefabs.

Example: importing a Universal Robot.
When importing a ur5 arm and then ur10 URDF importer will use the same meshes.

jhanca-robotecai:
@arturkamieniecki 
@lewandowskikacper 
is the issue solved now? can it be closed?

arturkamieniecki:
Closed with #394 merge.

PR #394, opened 2023-07-10
Make output directories of RobotImporter unique

lewandowskikacper:
Added Uuid of parameters passed to XACRO to the name name of RobotImporter's output directory.
This makes the directory unique for each robot imported from the same xacro file. 
Addresses  https://github.com/o3de/o3de-extras/issues/392.

PR #431, opened 2023-07-26
Fixed directory name creation in URDF Importer

lewandowskikacper:
https://github.com/o3de/o3de-extras/pull/394 resolved https://github.com/o3de/o3de-extras/issues/392 by introducing `dirSuffix` - additional directory path part, that was added in order to make it unique. 

In https://github.com/o3de/o3de-extras/pull/394 `dirSuffix` is an  `AZStd::string_view` created from `paramsUuid.ToString<AZStd::string_view>()`, which is an error, as the string is a temporary value. 
This PR changes it to be a `AZStd::string` instead, so that it's value lives for the duration of call to `CopyAssetForURDFAndCreateAssetMap`.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
@@ -182,7 +182,7 @@ namespace ROS2
             auto collidersNames = Utils::GetMeshesFilenames(m_parsedUrdf->getRoot(), false, true);
             auto visualNames = Utils::GetMeshesFilenames(m_parsedUrdf->getRoot(), true, false);
 
-            AZStd::string dirSuffix = "";
+            AZ::Uuid::FixedString dirSuffix;
             if (!m_params.empty())
             {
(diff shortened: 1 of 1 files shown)


Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -182,7 +182,7 @@ namespace ROS2
             auto collidersNames = Utils::GetMeshesFilenames(m_parsedUrdf->getRoot(), false, true);
             auto visualNames = Utils::GetMeshesFilenames(m_parsedUrdf->getRoot(), true, false);
 
-            AZStd::string_view dirSuffix = "";
+            AZStd::string dirSuffix = "";
```
lumberyard-employee-dm, line 185: Line 195 below needs to be changed to create an `AZStd::string` or alternatively use the ToFixedString function.
```suggestion
            AZ::Uuid::FixedString dirSuffix;
            // ...
            dirSuffix = paramsUuid.ToFixedString();
```

michalpelka (approved):

adamdbrw (approved):

PR #432, opened 2023-07-26
[URDF] Functioning asset builder

mbalfour-amzn:
This PR modifies the SDFAssetBuilder to produce functional procedural prefabs for URDF model files. It works by calling into the RobotImporter code and saving the result as a procedural prefab in the Cache directory instead of a regular prefab in a source asset directory. 

Features:
* Parses the URDF and generates the procprefab using the same logic as the RobotImporter.
* Creates AP job dependencies on every referenced asset to guarantee the procprefab can be generated and used without race conditions.

To make the code shareable between the Editor RobotImporter and the Asset Builder, some pieces of functionality needed to be refactored slightly, since the Asset Builder has less services available to it. In particular, the code now generates the hierarchy of entities _before_ creating the prefab template out of it, instead of using the prefab system to build up the entity hierarchy inside a prefab incrementally. Also, the importer code currently uses a pattern of entity activate / change component config / entity deactivate for setting up the components. Pieces of this are starting to migrate towards setting up component configurations without requiring entity activation. This will be more efficient overall, and reduce the reliance even further on runtime services (like Atom Feature Processors), but will likely require some changes to core o3de gems and components to expose component configurations a littler further first.

Some notes about the changes:
* The wheel physicsmaterial asset is no longer loaded during import, just referenced. The import didn't _require_ it to be loaded, it just required the asset reference to get hooked up correctly in the component. This needed to be changed since the asset builder can't support blocking asset loads.
* All of the entities getting created by the importer are now getting explicitly tracked in a flat list. The importer used to _implicitly_ track them by setting them to selected on creation and letting the prefab code build a prefab of all selected entities. Since the asset builder doesn't have a concept of selected entities, we needed to move to a more explicit tracking model.
* The method for finding and resolving external asset references is starting to branch somewhat between the RobotImporter and the SDFAssetBuilder. The asset builder will evolve towards a simplified model that tries to resolve relative paths and either succeeds or fails. The RobotImporter code uses environment variables and more heuristics. Eventually, if the simplified model is proven effective, the more complex lookups in the RobotImporter code could be removed.

Example video demonstrating use:

https://github.com/o3de/o3de-extras/assets/82224783/5c3dafbd-5c66-479c-b1c3-c8ea3de196d8

### DIFF
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -160,6 +160,8 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
             PUBLIC
                 Gem::${gem_name}.Editor.Static
                 Gem::Atom_Feature_Common.Static
+            PRIVATE
+                AZ::AzToolsFramework
     )
 
     # By default, we will specify that the above target ROS2 would be used by
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -161,8 +161,6 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
             PUBLIC
                 Gem::${gem_name}.Editor.Static
                 Gem::Atom_Feature_Common.Static
-            PRIVATE
-                AZ::AzToolsFramework
     )
 
     # By default, we will specify that the above target ROS2 would be used by
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
@@ -19,6 +19,7 @@
 #include <AzCore/std/smart_ptr/make_shared.h>
 #include <AzCore/std/smart_ptr/shared_ptr.h>
 #include <AzCore/std/string/string.h>
+#include <AzToolsFramework/Prefab/PrefabIdTypes.h>
 #include <AzToolsFramework/Prefab/PrefabPublicInterface.h>
 #include <RobotImporter/Utils/SourceAssetsStorage.h>
 
@@ -45,12 +46,11 @@ namespace ROS2
 
         //! Create and return a prefab corresponding to the URDF model as set through the constructor.
         //! @return result which is either the root prefab entity containing the imported model based on URDF or an error.
-        AzToolsFramework::Prefab::CreatePrefabResult CreatePrefabFromURDF();
+        AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string> CreatePrefabFromURDF();
 
-        //! Create and return a prefab JSON string corresponding to the URDF model set in the constructor.
-        //! @param outputPrefab The string to put the output prefab into.
-        //! @return result which is either the root prefab entity containing the imported model based on URDF or an error.
-        AzToolsFramework::Prefab::CreatePrefabResult CreatePrefabStringFromURDF(AZStd::string& outputPrefab);
+        //! Create and return a prefab template ID corresponding to the URDF model set in the constructor.
+        //! @return result which is either the prefab template id or an error message.
+        AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string> CreatePrefabTemplateFromURDF();
 
         //! Get path to the prefab resulting from the import.
         //! @return path to the prefab.
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
@@ -47,13 +47,17 @@ namespace ROS2
 
         ~URDFPrefabMaker() = default;
 
+        //! On prefab creation this will contain a prefab template id when successful,
+        //! and an error string on failure.
+        using CreatePrefabResult = AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string>;
+
         //! Create and return a prefab corresponding to the URDF model as set through the constructor.
         //! @return result which is either the root prefab entity containing the imported model based on URDF or an error.
-        AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string> CreatePrefabFromURDF();
+        CreatePrefabResult CreatePrefabFromURDF();
 
         //! Create and return a prefab template ID corresponding to the URDF model set in the constructor.
         //! @return result which is either the prefab template id or an error message.
-        AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string> CreatePrefabTemplateFromURDF();
+        CreatePrefabResult CreatePrefabTemplateFromURDF();
 
         //! Get path to the prefab resulting from the import.
         //! @return path to the prefab.
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
@@ -44,8 +44,13 @@ namespace ROS2
         ~URDFPrefabMaker() = default;
 
         //! Create and return a prefab corresponding to the URDF model as set through the constructor.
-        //! @return result which is either a prefab containing the imported model based on URDF or an error.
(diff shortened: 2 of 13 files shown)


Gems/ROS2/Code/CMakeLists.txt
```diff
@@ -161,6 +161,8 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
             PUBLIC
                 Gem::${gem_name}.Editor.Static
                 Gem::Atom_Feature_Common.Static
+            PRIVATE
+                AZ::AzToolsFramework
```
lumberyard-employee-dm, line 165: This is not needed.
The `ROS2.Editor` MODULE target gets the AzToolsFramework dependency via the `ROS2.Editor.Static` target above since it exposes `AzToolsFramework` as a public dependency

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
```diff
@@ -47,8 +48,12 @@ namespace ROS2
...
 
         //! Create and return a prefab corresponding to the URDF model as set through the constructor.
-        //! @return result which is either a prefab containing the imported model based on URDF or an error.
-        AzToolsFramework::Prefab::CreatePrefabResult CreatePrefabFromURDF();
+        //! @return result which is either the root prefab entity containing the imported model based on URDF or an error.
+        AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string> CreatePrefabFromURDF();
+
+        //! Create and return a prefab template ID corresponding to the URDF model set in the constructor.
+        //! @return result which is either the prefab template id or an error message.
+        AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string> CreatePrefabTemplateFromURDF();
```
lumberyard-employee-dm, line 56: Can you make an alias for the `AZ::Outcome<T, E>` type?
```c++
using CreateURDFPrefabResult = AZ::Outcome<AzToolsFramework::Prefab::TemplateId, AZStd::string>;
```

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -122,6 +123,35 @@ namespace ROS2::Utils
...
 
+    AZ::Data::AssetId GetProductAssetId(const AZ::Uuid& sourceAssetUUID, const AZ::TypeId typeId)
+    {
+        AZStd::vector<AZ::Data::AssetInfo> productsAssetInfo;
+        using AssetSysReqBus = AzToolsFramework::AssetSystemRequestBus;
+        bool ok{ false };
+        AssetSysReqBus::BroadcastResult(ok, &AssetSysReqBus::Events::GetAssetsProducedBySourceUUID, sourceAssetUUID, productsAssetInfo);
+        if (ok)
+        {
+            for (auto& product : productsAssetInfo)
```
lumberyard-employee-dm, line 134: This should be const auto since the container is not being mutated
```suggestion
            for (const auto& product : productsAssetInfo)
```

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.h
mbalfour-amzn: I'll just call it `CreatePrefabResult` though, so that it's one less thing we need to change as we broaden to both SDF and URDF. :)

lumberyard-employee-dm: Good with me.

mbalfour-amzn: Oh, wait, looking at the new PR diff, I just realized that's the same name as AzToolsFramework::Prefab::CreatePrefabResult. Let me pick something else to disambiguate. (I know they have different namespaces, but it can still get confusing)

lumberyard-employee-dm (approved):

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -109,18 +114,125 @@ namespace ROS2
...
+        auto assetNames = Utils::GetMeshesFilenames(rootLink, true, true);
+
+        Utils::UrdfAssetMap assetMap;
+
+        using AssetSysReqBus = AzToolsFramework::AssetSystemRequestBus;
+
+        // Unlike the RobotImporter, the SDF Asset Builder does not use the AMENT_PREFIX_PATH
+        // to resolve file locations. There wouldn't be a way to guarantee identical results across
+        // machines or to detect the need to rebuild assets if the environment variable changes.
+        const AZStd::string emptyAmentPrefixPath;
```
michalpelka, line 129: Why can we not ask the environment for ament path here?
I was not able to perform https://github.com/RobotecAI/o3de-ros2-gem-testing/blob/main/URDFTest/04_IMPORTING_URDF_2/IMPORTING_URDF2_TESTCASE.md test case using URDF builder.
Maybe AMENT_PREFIX_PATH can live in assetinfo?

michalpelka (approved):
Looks good to me.
I was able successfully test it with our cases:
# Test1
https://github.com/RobotecAI/o3de-ros2-gem-testing/blob/main/URDFTest/01_IMPORTING_INVALID_URDF/IMPORTING_INVALID_URDF_TESTCASE.md
**Result** Job fails in similar fashion to Robot importer

# Test 2 
https://github.com/RobotecAI/o3de-ros2-gem-testing/blob/main/URDFTest/02_IMPORTING_INVALID_XACRO/IMPORTING_INVALID_XACRO_TESTCASE.md
**Result** Xacro is no supported (no AP jobs fired)

# Test 3 
https://github.com/RobotecAI/o3de-ros2-gem-testing/blob/main/URDFTest/02_IMPORTING_VALID_SIMPLE_URDF/IMPORTING_VALID_SIMPLE_URDF_TESTCASE.md
** Result ** Success

# Test 4
https://github.com/RobotecAI/o3de-ros2-gem-testing/blob/main/URDFTest/04_IMPORTING_URDF_2/IMPORTING_URDF2_TESTCASE.md
**Changes** I've generated URDF from Xacro
**Result** Meshes were not imported, since I have no way to add ament path.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
adamdbrw: @michalpelka what do you propose to address the test case?

mbalfour-amzn: I don't actually know what the ament path represents or why it exists, so I'd definitely like to understand the use of ament path further. All I can do is explain my concern with it from an O3DE point of view.

In general, the goal is that O3DE projects should be easily shareable. The project itself should contain both the assets it needs and any configuration needed to process and use those assets. If the asset builder requires environment variables to be set a specific way to locate and process the assets, then the project will presumably be more difficult to share because it will require a manual environment setup outside of O3DE. This will also make things like configuring build machines that much more difficult, because they would require extra manual configuration steps for each project.

That said, we _already_ have a dependency on the currently-active ROS version (via `source`), so if this is a similar type of dependency where everyone is expected to have their environments configured the same way, maybe that's OK.

However, the design that we've currently got proposed is that we'll have settings files globally for the asset builder as well as optionally per-file. The first setting that we'll add is a way to directly map URI prefixes to relative directories. This would replace the current use of environment variables and search heuristics. The benefits are that it allows us to solve some of the cases that the current heuristics seemingly fail to find and it makes the settings themselves shareable through the project files. The main drawback is that it requires manual setup.

The settings file would look something like this:
```
{
    "uri_prefixes": {
        "model://": "Assets/my_robot/models/",
        "package://": "Assets/",
        "http://": "Assets/web/",
        "aws/models":  "Assets/aws/models/"
    }
}
```

Thoughts?

(Also, thank you to the pointer to the test cases, I wasn't aware these existed. FWIW, here's the 'in the wild' model that I was using to test this: https://github.com/ANYbotics/anymal_b_simple_description ).

mbalfour-amzn:
Since the tests passed and I have PR approvals, I've merged the changes. Issues with path resolution and xacro file support will get addressed in future PRs.

PR #433, opened 2023-07-27
Constructors with parameters in ROS2SensorComponents

jhanca-robotecai:
* move configuration of `ROS2GNSSSensorComponent`, `ROS2ImuSensorComponent`, `ROS2LidarSensorComponent`, and `ROS2Lidar2DSensorComponent` to separate structs
* add constructors for _ROS2SensorComponents_ that take parameters

The goal of this PR is to simplify the implementation of methods for creating sensors (it will be used in _RobotImporter_).

### DIFF
--- a/Gems/ROS2/Code/Source/Imu/ImuSensorConfiguration.cpp
+++ b/Gems/ROS2/Code/Source/Imu/ImuSensorConfiguration.cpp
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "ImuSensorConfiguration.h"
+#include <AzCore/Serialization/EditContext.h>
+#include <AzCore/Serialization/EditContextConstants.inl>
+
+namespace ROS2
+{
+    void ImuSensorConfiguration::Reflect(AZ::ReflectContext* context)
+    {
+        if (AZ::SerializeContext* serialize = azrtti_cast<AZ::SerializeContext*>(context))
+        {
+            serialize->Class<ImuSensorConfiguration>()
+                ->Version(1)
+                ->Field("FilterSize", &ImuSensorConfiguration::m_filterSize)
+                ->Field("IncludeGravity", &ImuSensorConfiguration::m_includeGravity)
+                ->Field("AbsoluteRotation", &ImuSensorConfiguration::m_absoluteRotation)
+                ->Field("AccelerationVariance", &ImuSensorConfiguration::m_linearAccelerationVariance)
+                ->Field("AngularVelocityVariance", &ImuSensorConfiguration::m_angularVelocityVariance)
+                ->Field("OrientationVariance", &ImuSensorConfiguration::m_orientationVariance);
+
+            if (AZ::EditContext* ec = serialize->GetEditContext())
+            {
+                ec->Class<ImuSensorConfiguration>("ROS2 IMU sensor configuration", "IMU sensor configuration")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Slider,
+                        &ImuSensorConfiguration::m_filterSize,
+                        "Filter Length",
+                        "Filter Length, Large value reduce numeric noise but increase lag")
+                    ->Attribute(AZ::Edit::Attributes::Max, &ImuSensorConfiguration::m_maxFilterSize)
+                    ->Attribute(AZ::Edit::Attributes::Min, &ImuSensorConfiguration::m_minFilterSize)
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &ImuSensorConfiguration::m_includeGravity,
+                        "Include Gravitation",
+                        "Enable accelerometer to observe gravity force.")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &ImuSensorConfiguration::m_absoluteRotation,
+                        "Absolute Rotation",
+                        "Include Absolute rotation in message.")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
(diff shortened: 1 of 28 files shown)


Gems/ROS2/Code/Source/Lidar/LidarSensorConfiguration.cpp
```diff
@@ -0,0 +1,142 @@
...
+        {
+            serializeContext->Class<LidarSensorConfiguration>()
+                ->Version(1)
+                ->Field("lidarModel", &LidarSensorConfiguration::m_lidarModel)
+                ->Field("lidarImplementation", &LidarSensorConfiguration::m_lidarSystem)
+                ->Field("LidarParameters", &LidarSensorConfiguration::m_lidarParameters)
+                ->Field("IgnoreLayer", &LidarSensorConfiguration::m_ignoreLayer)
+                ->Field("IgnoredLayerIndex", &LidarSensorConfiguration::m_ignoredLayerIndex)
+                ->Field("ExcludedEntities", &LidarSensorConfiguration::m_excludedEntities)
+                ->Field("PointsAtMax", &LidarSensorConfiguration::m_addPointsAtMax);
```
adamdbrw, line 27: This is not related to the PR, but please raise an issue to refactor out the common configuration between Lidar2D and Lidar3D. This is a lot of repetition that is clearly visible here. Handle differences such as population of combo boxes.

adamdbrw (commented):
Nice changes! How was this PR tested?

Gems/ROS2/Code/Source/Lidar/LidarSensorConfiguration.cpp
jhanca-robotecai: Issue created: https://github.com/o3de/o3de-extras/issues/435

jhanca-robotecai:
> Nice changes! How was this PR tested?

It was tested manually to ensure that all options are still available in _Editor_ and with some unit tests (not pushed) to test the constructors. The unit tests were based on this code: https://github.com/RobotecAI/o3de-extras/pull/10 and can be found in closed #429 for `ROS2CameraSensorComponent`. I can refactor the test code and push it to this PR if necessary.

adamdbrw:
Test code would be great!

adamdbrw (approved):

Gems/ROS2/Code/Source/Imu/ImuSensorConfiguration.cpp
```diff
@@ -0,0 +1,67 @@
...
+ *
+ */
+
+#include "ImuSensorConfiguration.h"
+#include <AzCore/Serialization/EditContext.h>
+#include <AzCore/Serialization/EditContextConstants.inl>
+
+namespace ROS2
+{
+    void ImuSensorConfiguration::Reflect(AZ::ReflectContext* context)
```
michalpelka, line 15: I do not like that change. There are designs that use that component and this changes reflection.

Gems/ROS2/Code/Source/Lidar/LidarSensorConfiguration.h
```diff
@@ -0,0 +1,54 @@
...
+#include <AzCore/std/string/string.h>
+
+#include "LidarRegistrarSystemComponent.h"
+#include "LidarTemplate.h"
+#include "LidarTemplateUtils.h"
+
+namespace ROS2
+{
+    //! A structure capturing configuration of a lidar sensor (to be used with ROS2LidarSensorComponent).
+    class LidarSensorConfiguration
```
michalpelka, line 22: It breaks all lidars in other gems and other projects.

michalpelka (changes requested):
Change from code point of view looks nice, but it will break all prefabs using touched components.

michalpelka (commented):
Also from UX point of view it looks unclean:
Pr:
image
Development:
image
Since all components were released it  should introduce translation for versions.

adamdbrw:
@jhanca-robotecai the default view in Editor is simple to address can be fixed through an attribute (expand).
@michalpelka I believe following the pattern of configuration reflection is good. It is a shame for projects that have serialized previous versions. However, we also updated version of the Gem to 2.0 recently, since we had API breaking changes. We won't avoid doing that between versions. I believe we get interfaces into good place before the next release - and if that includes breaking changes, so be it. The next release will happen in October most likely.
@jhanca-robotecai could you make sure versions are updated in serialization of components?

zakmat:
Change management is an inherent part of the software engineering process. The sooner we are able to battle-test our approach to component migration the better. [Version converters](https://docs.o3de.org/docs/user-guide/programming/components/reflection/serialization-context/versioning/) might be possibly used to change existing data in an unbreakable way, and developed idioms shall be kept as unit tests for further reference

adamdbrw (commented):
Looks good! Does anything else need to be changed with prefabs? I see changes for ROSBot but not Proteus.

jhanca-robotecai:
Based on the internal conversation it was decided to break the backward compatibility and fix the design issue. If decided otherwise, the sample implementation of the version converter (for one of the sensors) can be found here: https://github.com/jhanca-robotecai/o3de-extras/pull/1

The Editor's view was set to the same as before using the attributes:
image

Additionally, as the backward compatibility is broken, the parameter `visualise` was changed to `visualize`, to follow the coding guidelines (US English). 

Finally, the _prefab_ with _Slamtec_ sensor was fixed (the configuration of 3D lidar sensor was requested for 2D lidar sensor), before:
image
after:
image

adamdbrw:
The demos need to be adjusted to development anyway, we should skip adjustments for this PR alone

jhanca-robotecai:
> Looks good! Does anything else need to be changed with prefabs? I see changes for ROSBot but not Proteus.
There is no configuration stored in Proteus' prefab.

> The demos need to be adjusted to development anyway, we should skip adjustments for this PR alone
It is quite easy to miss the change, as de-serialization simply ignores the unknown fields without notifying the user about problems. Therefore I would strongly recommend doing it as fast as possible. E.g. `ROSConDemo` docker configuration use `development` branch of `o3de-extras` repo, which means failure right after this PR is merged.

adamdbrw:
@michalpelka take a look please

michalpelka (approved):

PR #436, opened 2023-07-27
Fix to conveyor belt compilation

adamdbrw:
Missing asset builder header, moved static to internal


lumberyard-employee-dm (approved):

spham-amzn (approved):

michalpelka (approved):
Tested against non-unity build.

PR #438, opened 2023-07-27
TFs also published to namespaced tf topic

adamdbrw:
This PR replaces https://github.com/o3de/o3de-extras/pull/408.

### DIFF
--- a/Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
+++ b/Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
@@ -20,7 +20,7 @@ namespace ROS2
         , m_childFrame(AZStd::move(childFrame))
         , m_isDynamic(isDynamic)
     {
-        auto splitName = ROS2Names::SplitIntoNamespaceAndName(childFrame);
+        auto splitName = ROS2Names::SplitIntoNamespaceAndName(m_childFrame);
         if (splitName.first.empty())
         {   // No namespace in child frame, no need for a publisher.
             return;
@@ -32,7 +32,7 @@ namespace ROS2
         {
             qos = tf2_ros::DynamicBroadcasterQoS();
         }
-        AZStd::string fullTopic = ROS2Names::GetNamespacedName(splitName.first, splitName.second);
+        AZStd::string fullTopic = ROS2Names::GetNamespacedName(splitName.first, m_isDynamic ? "tf" : "tf_static");
         m_transformPublisher = ros2Node->create_publisher<geometry_msgs::msg::TransformStamped>(fullTopic.data(), qos);
     }
 
--- a/Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
+++ b/Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
@@ -9,6 +9,7 @@
 #include <ROS2/Frame/ROS2Transform.h>
 #include <ROS2/ROS2Bus.h>
 #include <ROS2/Utilities/ROS2Conversions.h>
+#include <ROS2/Utilities/ROS2Names.h>
 #include <tf2_ros/qos.hpp>
 #include <tf2_ros/transform_broadcaster.h>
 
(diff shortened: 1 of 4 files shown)


pijaro (changes requested):

michalpelka (approved):

Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -38,7 +53,20 @@ namespace ROS2
...
         }
-        ROS2Interface::Get()->BroadcastTransform(CreateTransformMessage(transform), m_isDynamic);
+        auto transformMessage = CreateTransformMessage(transform);
+        ROS2Interface::Get()->BroadcastTransform(transformMessage, m_isDynamic);
+
+        if (m_transformPublisher)
+        {
+            AZStd::string parent_id(transformMessage.header.frame_id.c_str());
+            AZStd::string child_frame_id(transformMessage.child_frame_id.c_str());
+            ROS2Names::StripNamespace(parent_id);
```
MateuszWasilewski, line 63: This behavior of stripping namespace should be optional because this prevents simple tf merging / filtering.

Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -19,6 +20,20 @@ namespace ROS2
...
+        }
+
+        auto ros2Node = ROS2Interface::Get()->GetNode();
+        rclcpp::QoS qos = tf2_ros::StaticBroadcasterQoS();
+        if (m_isDynamic)
+        {
+            qos = tf2_ros::DynamicBroadcasterQoS();
+        }
+        AZStd::string fullTopic = ROS2Names::GetNamespacedName(splitName.first, m_isDynamic ? "tf" : "tf_static");
+        m_transformPublisher = ros2Node->create_publisher<geometry_msgs::msg::TransformStamped>(fullTopic.data(), qos);
```
MateuszWasilewski, line 36: In my opinion not every namespace should have separate TF tree.
Ex. robotic arm with many namespaced joints does not need seperate tf topic for each joint.
I believe this should be an opt in feature.

Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
```diff
@@ -38,7 +53,20 @@ namespace ROS2
         { // Only publish static transforms once
             return;
         }
-        ROS2Interface::Get()->BroadcastTransform(CreateTransformMessage(transform), m_isDynamic);
+        auto transformMessage = CreateTransformMessage(transform);
+        ROS2Interface::Get()->BroadcastTransform(transformMessage, m_isDynamic);
```
MateuszWasilewski, line 57: I'd love to be able to publish tf messages to `robotX/tf` but not to `/tf`

Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
MateuszWasilewski, line 36: Using `geometry_msgs::msg::TransformStamped` over `tf2_msgs::msg::TFMessage` prevents rqt and nav2 from detecting tfs correctly.

MateuszWasilewski (changes requested):
Welcome feature but some customization would be a good thing.

Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
adamdbrw: You are correct. Static broadcaster publishes a vector of TransformStamped, but (dynamic) broadcaster publishes TFMessage, see here:
https://github.com/ros2/geometry2/blob/rolling/tf2_ros/src/transform_broadcaster.cpp.

We should template the ROS2Transform class with the message type.

Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
adamdbrw: Since this is global and has to be known before we create publishers, I think that adding a reg setting is best

Gems/ROS2/Code/Source/Frame/ROS2Transform.cpp
adamdbrw: I do not see a satisfying solution in terms of UX at the moment, any suggestions?

adamdbrw:
After some discussion we decided it is better handled on the navigation stack side, with a good example delivered by @pijaro. Closing this issue.

PR #440, opened 2023-07-28
Warehouse automation gem

Antoni-Robotec:
Separated out the warehouse automation gem, updated the conveyor belt assets to use the new ConveyorBeltComponent.
Other minor asset fixes, made ProximitySensor independent of ROS2SensorComponent.
Should be merged alongside this PR: https://github.com/RobotecAI/ROSCon2023Demo/pull/21.


Gems/WarehouseAutomation/Code/Platform/Android/warehouseautomation_private_files.cmake
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) Contributors to the Open 3D Engine Project.
```
adamdbrw, line 1: These platform files for non-Linux platform are not needed

Gems/WarehouseAutomation/Code/Source/Clients/WarehouseAutomationModule.cpp
```diff
@@ -0,0 +1,23 @@
+/*
```
adamdbrw, line 1: What is the meaning of the "Clients" folder?

Gems/WarehouseAutomation/Code/Source/ConveyorBelt/ConveyorBeltComponentConfiguration.cpp
```diff
@@ -10,7 +10,7 @@
 #include <AzCore/Serialization/EditContext.h>
```
adamdbrw, line 10: Changes from #436 need to be applied here as well!

Gems/WarehouseAutomation/Code/Source/ProximitySensor/ROS2ProximitySensor.h
```diff
@@ -13,11 +13,11 @@
 #include <rclcpp/publisher.hpp>
 #include <std_msgs/msg/bool.hpp>
```
adamdbrw, line 14: Discuss with @michalpelka about making this ROS-independent, as these systems are typically not handled by ROS.

Gems/WarehouseAutomation/gem.json
```diff
@@ -0,0 +1,26 @@
+{
```
adamdbrw, line 1: Please take a look at Gem versioning and version dependencies (a recent feature), for example in other Gems in o3de-extras.

adamdbrw (commented):
Nice work! Some things need attention, as per comments.

michalpelka:
Why this name? It should be factory automation.

adamdbrw:
> Why this name? It should be factory automation.

I suggested this name. Depends on where we want to go with further assets and components here. Lets' think:
- ProductionAutomation?
- LogisticsAutomation?
I am also ok with FactoryAutomation.
@michalpelka share your ideas.

Gems/WarehouseAutomation/Code/Source/Clients/WarehouseAutomationModule.cpp
Antoni-Robotec: I haven't been able to figure out why it's called that (it's part of the gem creation template). It has some of the files that were previously kept in the main Source directory so it was probably added to keep it clean.

Gems/WarehouseAutomation/Assets/Factory/Box1.prefab
```diff
@@ -0,0 +1,239 @@
+{
```
Antoni-Robotec, line 1: The assets have been moved improperly, they link to materials within their old directory and thus don't render.

Gems/WarehouseAutomation/Code/Source/Clients/WarehouseAutomationModule.cpp
adamdbrw: I think we can remove the directory

Antoni-Robotec: Should the Tools directory be removed as well then? It's similar, except for EditorComponents, and I also don't know where the name comes from.

Gems/WarehouseAutomation/Code/Source/ProximitySensor/ROS2ProximitySensor.h
Antoni-Robotec: Made the proximity sensor ROS-independent.

Gems/WarehouseAutomation/Assets/Factory/Box1.prefab
Antoni-Robotec: Fixed

Gems/WarehouseAutomation/Code/Source/ProximitySensor/ProximitySensor.cpp
```diff
@@ -0,0 +1,177 @@
+/*
```
adamdbrw, line 1: Does not look clang-formatted. Please apply formatting to entire PR.

Gems/WarehouseAutomation/Code/Source/WarehouseAutomationModuleInterface.cpp
```diff
@@ -0,0 +1,42 @@
...
+namespace WarehouseAutomation
+{
+    AZ_TYPE_INFO_WITH_NAME_IMPL(WarehouseAutomationModuleInterface,
+        "WarehouseAutomationModuleInterface", WarehouseAutomationModuleInterfaceTypeId);
+    AZ_RTTI_NO_TYPE_INFO_IMPL(WarehouseAutomationModuleInterface, AZ::Module);
+    AZ_CLASS_ALLOCATOR_IMPL(WarehouseAutomationModuleInterface, AZ::SystemAllocator);
+
+    WarehouseAutomationModuleInterface::WarehouseAutomationModuleInterface()
+    {
+        // Push results of [MyComponent]::CreateDescriptor() into m_descriptors here.
```
adamdbrw, line 25: Check comments such as this, which should not be left from the generated Gem

Gems/WarehouseAutomation/Code/Source/WarehouseAutomationSystemComponent.cpp
```diff
@@ -0,0 +1,83 @@
...
+        AZ::TickBus::Handler::BusConnect();
+    }
+
+    void WarehouseAutomationSystemComponent::Deactivate()
+    {
+        AZ::TickBus::Handler::BusDisconnect();
+        WarehouseAutomationRequestBus::Handler::BusDisconnect();
+    }
+
+    void WarehouseAutomationSystemComponent::OnTick([[maybe_unused]] float deltaTime, [[maybe_unused]] AZ::ScriptTimePoint time)
```
adamdbrw, line 79: I don't think that TickBus interface is needed at all

Gems/WarehouseAutomation/Code/Include/WarehouseAutomation/WarehouseAutomationBus.h
```diff
@@ -0,0 +1,39 @@
...
+#include <AzCore/Interface/Interface.h>
+
+namespace WarehouseAutomation
+{
+    class WarehouseAutomationRequests
+    {
+    public:
+        AZ_RTTI(WarehouseAutomationRequests, WarehouseAutomationRequestsTypeId);
+        virtual ~WarehouseAutomationRequests() = default;
+        // Put your public methods here
```
adamdbrw, line 22: There seems to be no methods. Are there any planned or should we discard the interface?

Antoni-Robotec: I don't really know the planned expansion direction for this gem so I decided to leave such files in (they're much easier to delete later than to add from scratch). I imagine there is some likelihood of "global methods" being needed in the future.
Do you think such methods are unlikely? If so, I'll delete this interface, and probably also this gem's SystemComponents.

Gems/WarehouseAutomation/Code/Source/WarehouseAutomationModuleInterface.cpp
Antoni-Robotec: In my opinion they are useful for clarification (future components will also need to be added here).
Should they be removed from the ROS2 gem as well then?

adamdbrw: Yes, these are not needed.

Gems/WarehouseAutomation/Code/Include/WarehouseAutomation/WarehouseAutomationBus.h
adamdbrw: This Gem is a "bunch of things" rather than a system, so yes I think the interface would not be needed.

michalpelka (changes requested):
This contribution has issue with LFS and cannot be checked out. 
Please make sure that PR is available for reviewers.

michalpelka:
Gem is cannot be build due to missing headers.

Gems/WarehouseAutomation/Code/Source/ProximitySensor/ProximitySensor.h
```diff
@@ -9,20 +10,19 @@
 
 #include <Atom/RPI.Public/AuxGeom/AuxGeomDraw.h>
```
michalpelka, line 11:
```suggestion
#include <Atom/RPI.Public/AuxGeom/AuxGeomDraw.h>
#include <AzCore/Component/Component.h>
#include <AzCore/Component/TickBus.h>
#include <AzCore/Math/Vector3.h>
```

Gems/WarehouseAutomation/Assets/Factory/ConveyorElements/ConveyorSegments_Turnout.prefab
```diff
@@ -0,0 +1,2218 @@
+{
```
michalpelka, line 1: Please adjust length of spline.
Screenshot from 2023-08-29 18-25-01

michalpelka (commented):
Technically this gem depends on ROS2 gem (due to usage of motorized joint). 
I would add for now dependency, but remove after moving motorized joint to another gem.

Gems/WarehouseAutomation/Assets/Factory/ConveyorElements/ConveyorSegments_Turnout.prefab
Antoni-Robotec: Good catch!

Antoni-Robotec:
> 

I feel like enabling it for windows is a bad idea if we don't add some automated check that at the very least makes sure it compiles on windows.

michalpelka (approved):

adamdbrw (approved):

PR #444, opened 2023-08-02
Spawning with namespace

pijaro:
Support for namespace parameter in the robot spawner.

### DIFF
--- a/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -91,6 +91,7 @@ namespace ROS2
     void ROS2SpawnerComponent::SpawnEntity(const SpawnEntityRequest request, SpawnEntityResponse response)
     {
         AZStd::string spawnableName(request->name.c_str());
+        AZStd::string spawnableNamespace(request->robot_namespace.c_str());
         AZStd::string spawnPointName(request->xml.c_str(), request->xml.size());
 
         auto spawnPoints = GetSpawnPoints();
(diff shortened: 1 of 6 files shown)


Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -159,10 +161,11 @@ namespace ROS2
         AZStd::string instanceName = AZStd::string::format("%s_%d", spawnableName.c_str(), m_counter++);
         for (AZ::Entity* entity : view)
         { // Update name for the first entity with ROS2Frame in hierarchy (left to right)
-            const auto* frameComponent = Utils::GetGameOrEditorComponent<ROS2FrameComponent>(entity);
+            auto* frameComponent = Utils::GetGameOrEditorComponent<ROS2FrameComponent>(entity);
             if (frameComponent)
             {
                 entity->SetName(instanceName);
+                frameComponent->UpdateNamespaceConfiguration(spawnableNamespace, NamespaceConfiguration::NamespaceStrategy::Custom);
```
zakmat, line 168: Have you considered checking if spawnableNamespace is empty? As I understand it will be empty if omitted by a service caller, perhaps in such cases namespace shall not be altered?

Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -91,6 +91,7 @@ namespace ROS2
     void ROS2SpawnerComponent::SpawnEntity(const SpawnEntityRequest request, SpawnEntityResponse response)
     {
         AZStd::string spawnableName(request->name.c_str());
+        AZStd::string spawnableNamespace(request->robot_namespace.c_str());
```
zakmat, line 94: RosifyName could be used to sanitize external input.

zakmat (commented):
I am also waiting for the namespace support in spawnables. I hope it will be merged soon! I have added some comments for consideration

Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
adamdbrw: Since the request was to spawn with a specific name / namespace, there could be a problem, especially with automated runs, if the actual name/namespace was different. Thus, I believe it is better to validate (using our utils) instead and return failure if the validation does not pass.

zakmat (approved):

adamdbrw (approved):
I have some ideas about NamespaceConfiguration refactor, so that it can be set dynamically as a whole configuration.

pawelbudziszewski (approved):

michalpelka (approved):

PR #445, opened 2023-08-02
Flexible publishing PoC

adamdbrw:
Sometimes, the user might want to replace publishing with other behavior (such as recording data to bags directly with API), or to do additional things before or after publishing. This PR enables that.


Gems/ROS2/Code/Source/ContactSensor/ROS2ContactSensorComponent.cpp
```diff
@@ -9,6 +9,7 @@
 #include "ROS2ContactSensorComponent.h"
 #include <AzCore/Component/ComponentApplicationBus.h>
 #include <AzCore/Debug/Trace.h>
+#include <AzCore/Serialization/EditContext.h>
```
adamdbrw, line 12: This change will be gone from final PR, but fixes compile error present at the branching so it's easier to test.

jhanca-robotecai:
I am closing this draft as inactive. @adamdbrw please reopen it when the code is ready for the review.

PR #446, opened 2023-08-02
add SDFormat sensor importer hooks

jhanca-robotecai:
This PR is based on #437

The context: the reflection hook for user-defined mappings will allow developers to add new plugins and support them with gems and components. This will work similarly to AssetBuilder registration system, using SerializeContext reflection system. 

This PR adds the following:
* Add _SDFormat_ sensor importer hook template
* Add sample implementation of the hook for `ROS2CameraSensorEditorComponent`
* Add _SDFormat_ sensor importer hook utilities
* Add tests

### DIFF
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -111,20 +111,6 @@ ly_create_alias(NAME ${gem_name}.Unified NAMESPACE Gem TARGETS Gem::${gem_name})
 # will also depend on ${gem_name}.Static
 if(PAL_TRAIT_BUILD_HOST_TOOLS)
     ly_add_target(
-        NAME ${gem_name}.Editor.API INTERFACE
-        NAMESPACE Gem
-        FILES_CMAKE
-            ros2_editor_api_files.cmake
-        INCLUDE_DIRECTORIES
-            INTERFACE
-                Include
-        BUILD_DEPENDENCIES
-            INTERFACE
-                AZ::AzToolsFramework
-                3rdParty::sdformat
-    )
-
-    ly_add_target(
         NAME ${gem_name}.Editor.Static STATIC
         NAMESPACE Gem
         AUTOMOC
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -68,7 +68,6 @@ ly_add_target(
             Gem::StartingPointInput
             Gem::PhysX.Static
             Gem::LmbrCentral.API
-            3rdParty::sdformat
 )
 
 target_depends_on_ros2_packages(${gem_name}.Static rclcpp builtin_interfaces std_msgs sensor_msgs nav_msgs tf2_ros ackermann_msgs gazebo_msgs)
@@ -111,6 +110,20 @@ ly_create_alias(NAME ${gem_name}.Unified NAMESPACE Gem TARGETS Gem::${gem_name})
 # If we are on a host platform, we want to add the host tools targets like the ${gem_name}.Editor target which
 # will also depend on ${gem_name}.Static
 if(PAL_TRAIT_BUILD_HOST_TOOLS)
+    ly_add_target(
+        NAME ${gem_name}.Editor.API INTERFACE
+        NAMESPACE Gem
+        FILES_CMAKE
+            ros2_editor_api_files.cmake
+        INCLUDE_DIRECTORIES
+            INTERFACE
+                Include
+        BUILD_DEPENDENCIES
+            INTERFACE
+                AZ::AzToolsFramework
+                3rdParty::sdformat
+    )
+
     ly_add_target(
         NAME ${gem_name}.Editor.Static STATIC
         NAMESPACE Gem
@@ -219,6 +232,7 @@ if(PAL_TRAIT_BUILD_TESTS_SUPPORTED)
                     PRIVATE
                         AZ::AzTest
                         Gem::${gem_name}.Editor
+                        3rdParty::sdformat
             )
 
             # Add ROS2.Editor.Tests to googletest
--- a/Gems/ROS2/Code/CMakeLists.txt
+++ b/Gems/ROS2/Code/CMakeLists.txt
@@ -68,6 +68,7 @@ ly_add_target(
             Gem::StartingPointInput
             Gem::PhysX.Static
             Gem::LmbrCentral.API
+            3rdParty::sdformat
 )
 
 target_depends_on_ros2_packages(${gem_name}.Static rclcpp builtin_interfaces std_msgs sensor_msgs nav_msgs tf2_ros ackermann_msgs gazebo_msgs)
@@ -165,8 +166,8 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
     )
 
     # By default, we will specify that the above target ROS2 would be used by
-    # Tool and Builder type targets when this gem is enabled.  If you don't want it
-    # active in Tools or Builders by default, delete one of both of the following lines:
+    # Tool and Builder type targets when this gem is enabled. If you don't want it
+    # active in Tools or Builders by default, delete one or both of the following lines:
     ly_create_alias(NAME ${gem_name}.Tools    NAMESPACE Gem TARGETS Gem::${gem_name}.Editor)
     ly_create_alias(NAME ${gem_name}.Builders NAMESPACE Gem TARGETS Gem::${gem_name}.Editor)
 endif()
--- a/Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
+++ b/Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/Component/Entity.h>
+#include <AzCore/std/containers/unordered_set.h>
+#include <AzCore/std/function/function_template.h>
+#include <AzCore/std/string/string.h>
+
+#include <sdf/Sensor.hh>
+
+namespace ROS2::SDFormat
+{
+    //! Hook used in ROS2 Gem Robot Importer to create ROS2 sensor components based on SDF model description file.
+    //! See the <sensor> element at http://sdformat.org/spec?ver=1.10&elem=sensor for more details.
+    struct SensorImporterHook
+    {
+        AZ_TYPE_INFO(SensorImporterHook, "{23f189e3-8da4-4498-9b89-1ef6c900940a}");
+
+        //! Types of the sensor in the SDF file that can be reused in O3DE.
+        //! Multiple SDF sensors can map to one O3DE component.
+        AZStd::unordered_set<sdf::SensorType> m_sensorTypes;
+
+        //! Names of the options of the sensor in the SDF file that can be reused in O3DE.
+        AZStd::unordered_set<AZStd::string> m_sensorOptions;
+
+        //! Names of the plugins associated with the sensor in the SDF file that can be reused in O3DE.
+        //! Multiple SDF plugins can map to one O3DE component.
+        AZStd::unordered_set<AZStd::string> m_pluginNames;
+
+        //! Names of the options of the plugin associated with the sensor in the SDF file that can be reused in O3DE.
+        AZStd::unordered_set<AZStd::string> m_pluginOptions;
+
+        //! Registered function, that is invoked when an plugin with the specified name
+        //! is processed by the SDF Importer.
+        //! A reference to an Entity is supplied which can be populated with one or more O3DE Components
+        //! that were converted using the SDF Plugin data.
+        using ConvertSDFSensorCallback = AZStd::function<void(AZ::Entity&, const sdf::Sensor&)>;
+        ConvertSDFSensorCallback m_sdfSensorToComponentCallback;
+    };
+} // namespace ROS2::SDFormat
--- a/Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
+++ b/Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
@@ -24,30 +24,32 @@ namespace sdf
 
 namespace ROS2::SDFormat
 {
-    //! Hook used in ROS2 Gem Robot Importer to create ROS2 sensor components based on SDF model description file.
-    //! See the <sensor> element at http://sdformat.org/spec?ver=1.10&elem=sensor for more details.
+    //! Hook used in ROS2 Gem Robot Importer to create ROS2 sensor components based on SDFormat model description.
+    //! It implements the parameters mapping between the SDFormat sensor and the particular O3DE component.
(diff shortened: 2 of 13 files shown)


Gems/ROS2/Code/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ ly_add_target(
             Gem::StartingPointInput
             Gem::PhysX.Static
             Gem::LmbrCentral.API
+            3rdParty::sdformat
```
lumberyard-employee-dm, line 71: The sdformat library can't be added as a dependency of the `${gem_name}.Static` target, as it not be available in the ${project}.GameLauncher for processing sdf files and only in Tools such as the Editor and AssetPRocessor.

I am assuming you added this because the `SDFormatSensorImporterHook.h` is in the `ros2_header_files.cmake` file which doesn't have the sdformat dependency.

Instead what could be done is either add ``SDFormatSensorImporterHook.h` to the `ros2_editor_files.cmake` so that it gets added to the `ROS2.Editor.Static` target and would be available OR
(My recommendation) add a `${gem_name}.Editor.API` similar to the `${gem_name}.API` target below which exposes this hook file to other Gem targets.

That second option keeps more inline with how we want the [Gem code structure](https://github.com/o3de/o3de/blame/development/Templates/DefaultGem/Template/Code/CMakeLists.txt#L101-L112) to be formed going forward with a clear separation between public API for other gems to use(${gem_name}.Editor.API) versus private implementations in static libraries that are only linked to the gems modules(${gem_name}.Editor.Static)

An API target for Tools applications can be added in the `PAL_TRAIT_BUILD_HOST_TOOLS` block below, so it only builds when tooling application such as the Editor/AssetProcessor available which are in non-monolithic context
```c++
    if(PAL_TRAIT_BUILD_HOST_TOOLS)
        NAME ${gem_name}.Editor.API INTERFACE
        NAMESPACE Gem
        FILES_CMAKE
            ros2_editor_api_files.cmake
        INCLUDE_DIRECTORIES
            INTERFACE
                Include
        BUILD_DEPENDENCIES
            INTERFACE
                AZ::AzToolsFramework
    )
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -324,4 +325,55 @@ namespace ROS2
         return resolvedPath.String();
     }
 
+    namespace Utils::SDFormat
+    {
+        AZStd::string GetPluginFilename(const sdf::Plugin& plugin)
+        {
+            const std::filesystem::path path = plugin.Filename();
+            return AZStd::string(path.filename().u8string().c_str(), path.filename().u8string().size());
+        }
```
lumberyard-employee-dm, line 334: Instead of adding the `<filesystem>` include an using std::filesystem, this code can go directly to an AZ::IO::Path.
I also think that would be a better return value than AZStd::string, since that enforces specifically what kind of value is the result of the function returns
```suggestion
        AZ::IO::Path GetPluginFilename(const sdf::Plugin& plugin)
        {
            const std::string& path = plugin.Filename();
            return AZStd::string(path.c_str(), path.size());
        }
```

Gems/ROS2/Code/Tests/SdfParserTest.cpp
```diff
@@ -0,0 +1,241 @@
...
+                   "            <max>10</max>\n"
+                   "            <resolution>0.01</resolution>\n"
+                   "          </range>\n"
+                   "        </ray>\n"
+                   "        <plugin name=\"laser_plug\" filename=\"librayplugin.so\"/>\n"
+                   "      </sensor>\n"
+                   "    </link>\n"
+                   "  </model>\n"
+                   "</sdf>\n";
+        }
```
lumberyard-employee-dm, line 91: FYI: For future references, if you need support for a multiline string, you can also use the Raw string literal `R"()"` support: https://en.cppreference.com/w/cpp/language/string_literal.

In that case the `'\n'` wouldn't need to be explicitly specified and the double quotes wouldn't need to be escape..
Since XML is being parsed the leading whitespace before tags isn't relevant to the parsing aspect".

For example
```
constexpr std::string_view foo = R"(<ObjectStream version="3">
                    <Class name="DataPatch" type="{3A8D5EC9-D70E-41CB-879C-DEF6A6D6ED03}">
                            <Class name="AZ::Uuid" field="m_targetClassId" value="{47E5CF10-3FA1-4064-BE7A-70E3143B4025}" type="{E152C105-A133-4D03-BBF8-3D4B2FBA3E2A}"/>
                            <Class name="AZStd::unordered_map" field="m_patch" type="{CF3B3C65-49C0-5199-B671-E75347EB25C2}">
                                    <Class name="AZStd::pair" field="element" type="{07DEDB71-0585-5BE6-83FF-1C9029B9E5DB}">
                                            <Class name="AddressType" field="value1" value="C705B33500000000" type="{90752F2D-CBD3-4EE9-9CDD-447E797C8408}"/>
                                            <Class name="ByteStream" field="value2" value="00000000031C72039442EB384D42A1ADCB68F7E0EEF6000000960000" type="{ADFD596B-7177-5519-9752-BC418FE42963}"/>
                                    </Class>
                            </Class>
                    </Class>
            </ObjectStream>
            )";
```

Gems/ROS2/Code/CMakeLists.txt
jhanca-robotecai: I applied your suggestion with a forward declaration, which makes this problem gone. Nevertheless, thank you for the detailed explanation.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
jhanca-robotecai: The goal of this method is to strip the possible path and keep the filename only. As pointed out by Mike, the filename tag could include the folder or not [source](http://sdformat.org/spec?ver=1.10&elem=model#plugin_filename). The matching is based on the plugin names solely. 

I removed the `<filesystem>` include and used `AZ::IO::Path` instead, which simplified the code. Thanks for reminding me about not using `std` when not needed.

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
```diff
@@ -0,0 +1,54 @@
...
+        AZStd::unordered_set<AZStd::string> m_pluginNames;
+
+        //! Names of the options of the plugin associated with the sensor in the SDF file that can be reused in O3DE.
+        AZStd::unordered_set<AZStd::string> m_pluginOptions;
+
+        //! Registered function, that is invoked when an plugin with the specified name
+        //! is processed by the SDF Importer.
+        //! A reference to an Entity is supplied which can be populated with one or more O3DE Components
+        //! that were converted using the SDF Plugin data.
+        using ConvertSDFSensorCallback = AZStd::function<void(AZ::Entity&, const sdf::Sensor&)>;
```
lumberyard-employee-dm, line 51: Should we have a callback return a boolean our an AZ::Outcome of a vector of component typeis that were added to the entity on success, but an error message on why the callback failed on error.
i.e
```c++
using ErrorString = AZStd::string;
using ComponentTypeIds = AZStd::vector<AZ::TypeId>;
using ConvertSensorOutcome = AZ::Outcome<ComponentTypeIds, ErrorString>;
using ConvertSDFSensorCallback = AZStd::function<ConvertSensorOutcome(AZ::Entity&, const sdf::Sensor&)>;
```

The user can then in their callback return a populate a vector that is returned which every component added to the entity on success or return an error message on failure.

Using the ROS2CameraSensor lambda callback below, some psuedocode which illustrates using a return value can be as follows
```c++
 importerHook.m_sdfSensorToComponentCallback = [](AZ::Entity& entity, const sdf::Sensor& sdfSensor)
 {
    SensorImporterHook::ComponentTypeIds newComponentTypeIds;
    // ... existing logic
    if (someFailureOccured)
    {
        SensorImporterHook::ErrorString error = "Failed to convert ROS2 Camera Sensor for reason <x>";
        return AZ::Failure(error);
    }
    // ... more existing logic
   entity.CreateComponent<ROS2CameraSensorEditorComponent>(sensorConfiguration, cameraConfiguration);
   newComponentTypeIds.push_back(AZ::AzTypeInfo<ROS2CameraSensorEditorComponent>::GetCanonicalTypeId());

   return newComponentTypeIds;
}
```

jhanca-robotecai: I missed the return type on purpose, as I am not sure how the hook will be called from the importer. Similarly, the current hook for `ROS2CameraSensor` does not check for the required and incompatible services, because I am not sure yet if this call should be made from within the hook or from the code that parses `<sensor>` tags and calls the hook.

Let me know if I should change it now.

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
```diff
@@ -0,0 +1,54 @@
...
+    inline namespace v13
+    {
+        enum class SensorType;
+        class Sensor;
+    } // namespace v13
+} // namespace sdf
+
+namespace ROS2::SDFormat
+{
+    //! Hook used in ROS2 Gem Robot Importer to create ROS2 sensor components based on SDF model description file.
```
adamdbrw, line 27: Add a single sentence to help developers understand how to use this hook.

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
```diff
@@ -0,0 +1,54 @@
...
+    //! See the <sensor> element at http://sdformat.org/spec?ver=1.10&elem=sensor for more details.
+    struct SensorImporterHook
+    {
+        AZ_TYPE_INFO(SensorImporterHook, "{23f189e3-8da4-4498-9b89-1ef6c900940a}");
+
+        //! Types of the sensor in the SDF file that can be reused in O3DE.
```
adamdbrw, line 33: What does it mean reused? Perhaps add a comment about equivalence or similarity or mapping to the struct top level doxy comment.

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
```diff
@@ -0,0 +1,54 @@
...
+        //! Multiple SDF sensors can map to one O3DE component.
+        AZStd::unordered_set<sdf::SensorType> m_sensorTypes;
+
+        //! Names of the options of the sensor in the SDF file that can be reused in O3DE.
```
adamdbrw, line 37:
```suggestion
        //! Configuration parameters the sensor in the SDF file that can be reused in O3DE.
```

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
```diff
@@ -0,0 +1,54 @@
...
+        AZStd::unordered_set<AZStd::string> m_sensorOptions;
+
+        //! Names of the plugins associated with the sensor in the SDF file that can be reused in O3DE.
+        //! Multiple SDF plugins can map to one O3DE component.
```
adamdbrw, line 41: A plugin can also map to several components at once (because of modularity of things like Vehicle Dynamics, we need to create several components, sometimes these components might now be strictly required by each other).

A plugin could also map to a few alternative components (not together), but this one could be a future extension.

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
```diff
@@ -0,0 +1,54 @@
...
+        AZStd::unordered_set<AZStd::string> m_pluginNames;
+
+        //! Names of the options of the plugin associated with the sensor in the SDF file that can be reused in O3DE.
```
adamdbrw, line 44:
```suggestion
        //! Configuration parameters of the plugin associated with the sensor in the SDF file that can be reused in O3DE.
```

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
```diff
@@ -0,0 +1,54 @@
...
+        AZStd::unordered_set<AZStd::string> m_pluginOptions;
```
adamdbrw, line 45: Adjust names to ConfigurationParameters or similar

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
adamdbrw: It would be great if we could return comprehensive information about unsupported parameters and other non-failing imperfections.

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
jhanca-robotecai: Renamed to terms such as `mapping` and `supported`

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
jhanca-robotecai: This particular hook will only be used with sensors. Based on the defined list of sensor types available in SDFormat, I believe we will always map to one O3DE component. 

Plugins related to models and worlds will be handled by another hook.

jhanca-robotecai:
> It would be great if we could return comprehensive information about unsupported parameters and other non-failing imperfections.

Such information will be available based on the predefined sets of supported sensor parameters and the plugin's parameters in the hook. Methods providing this information are implemented in the utilities and are provided with some tests. I do believe `ConvertSDFSensorCallback` should only create a component, while all calls giving additional data should be executed outside of this callback. Please let me know if you agree, otherwise I will add it into the callback.

Gems/ROS2/Code/Include/ROS2/RobotImporter/SDFormatSensorImporterHook.h
jhanca-robotecai: Such information will be available based on the predefined sets of supported sensor parameters and the plugin's parameters in the hook. Methods providing this information are implemented in the utilities and are provided with some tests. I do believe `ConvertSDFSensorCallback` should only create a component(s) and return success/failure, while the remaining calls, such as finding unsupported parameters, should be executed outside of this callback. Please let me know if you agree, otherwise, I will add it to the callback.

lumberyard-employee-dm: At the very least, the implementer of the hook should have some kind of way to return back important information about the operation.
Such as any components they added via converting a sensor or an error message indicating why a sensor couldn't be converted.

That way the RobotImporter can collect all of this information and display any errors in a UI widget later if required.

jhanca-robotecai: Pushed the code with `AZ::Success`/`AZ::Failure`. I believe we can change it later on to `AZStd::vector<AZ::TypeId>` or any other structure needed by the method triggering sensor creation.

lumberyard-employee-dm: That sounds good to me.

lumberyard-employee-dm (approved):

adamdbrw (approved):

Issue #442, opened 2023-08-01
Support position control of joints through ROS 2


adamdbrw:
We already expose JointsManipulationRequests including API for position control. However, this feature is not available through ROS 2 subscriber.

The subscriber should most likely be placed in JointsManipulationComponent and its callback should call MoveJointsToPostiions.

Check the solution by comparing with:
- ros2 position controller: https://github.com/ros-controls/ros2_controllers/blob/master/position_controllers/src/joint_group_position_controller.cpp
- Gazebo implementation is https://gazebosim.org/api/gazebo/4.1/classignition_1_1gazebo_1_1systems_1_1JointPositionController.html.

PR #454, opened 2023-08-09
Addded subscription of position control to JointsManipulationComponent

lewandowskikacper:
Solves https://github.com/o3de/o3de-extras/issues/442.

Introduces position control interface in `JointsManipulationComponent`. 
the interface is `std_msgs/msg/Float64MultiArray`. 

Example usage:
`ros2 topic pub --once /position_controller/commands  std_msgs/msg/Float64MultiArray "{data: [1.4, 0.4, 0, -1.2, -1.6, -0.1]}"`
Please note that the number of values must match the number of joints exactly. Values are mapped to joints by order (there is new parameter in "Initial Positions" to control this).

There is a script for migrating prefabs (but not overrides)  to the new version.
```
import sys
import json
import os

def find_and_replace(data):

    def search_for_components(item, foundComponents):
        if isinstance(item, dict):
            if "Components" in item:
                foundComponents = True
            if "JointsManipulationEditorComponent" in item and foundComponents:
                modifiedElement = item["JointsManipulationEditorComponent"]
                # modifiedElement["$type"] = "JointsManipulationEditorComponent"
                # del modifiedElement["m_template"]["$type"]

                joints = []
                index = 0
                for name in modifiedElement["Initial positions"]:
                    print("Found joint: ", name)
                    position = modifiedElement["Initial positions"][name]
                    print(f"Found joint-> name: {name}, position: {position}")
                    joints.append(
                        {
                            "Name" : name,
                            "Position": position,
                            "Index": index
                        }
                    )
                    index = index + 1

                modifiedElement["Initial positions ordered"] = joints

                del modifiedElement["Initial positions"]

                # configuration = {"Configuration" : modifiedElement["m_template"]}
                # modifiedElement["Controller"] = configuration
                # del modifiedElement["m_template"]
                # del item["ROS2FrameComponent"]
                item["JointsManipulationEditorComponent"] = modifiedElement
            # if "op" in item and item["op"] == "replace" and "path" in item:
                # item["path"] = item["path"].replace("ROS2FrameComponent/m_template", "ROS2FrameEditorComponent/Controller/Configuration")
            for value in item.values():
                search_for_components(value, foundComponents)
        elif isinstance(item, list):
            for element in item:
                search_for_components(element, foundComponents)

    search_for_components(data, False)


def write_json_to_file(data, filename):
    try:
        with open(filename, 'w') as file:
            json.dump(data, file, indent=4)
            print(f"Modified data written to '{filename}'.")
    except Exception as e:
        print(f"Error writing to '{filename}': {e}")

def find_prefab_files(directory):
    prefab_files = []
    if os.path.isfile(directory) and directory.endswith(".prefab"):
        prefab_files.append(directory)
    elif os.path.isdir(directory):
        for root, dirs, files in os.walk(directory):
            for file in files:
                if file.endswith(".prefab"):
                    prefab_files.append(os.path.join(root, file))
    return prefab_files

def process_prefab_file(file_path):
    try:
        with open(file_path, 'r') as file:
            data = json.load(file)
            find_and_replace(data)
        with open(file_path, 'w') as file:
            json.dump(data, file, indent=4)
            print(f"Modified data written to '{file_path}'.")
    except Exception as e:
        print(f"Error processing '{file_path}': {e}")

if __name__ == "__main__":
    if len(sys.argv) != 2:
        print("Usage: python script.py <directory_or_file>")
    else:
        input_path = sys.argv[1]
        prefab_files = find_prefab_files(input_path)
        if not prefab_files:
            print(f"No .prefab files found in '{input_path}'.")
        else:
            for file_path in prefab_files:
                print(f"Processing '{file_path}'...")
                process_prefab_file(file_path)
```

### DIFF
--- a/Gems/ROS2/Code/Source/Manipulation/JointsManipulationComponent.h
+++ b/Gems/ROS2/Code/Source/Manipulation/JointsManipulationComponent.h
@@ -14,7 +14,9 @@
 #include <AzCore/Name/Name.h>
 
 #include "JointStatePublisher.h"
+#include "std_msgs/msg/float64_multi_array.hpp"
 #include <ROS2/Manipulation/JointsManipulationRequests.h>
+#include <rclcpp/subscription.hpp>
 
 namespace ROS2
 {
@@ -28,7 +30,10 @@ namespace ROS2
     public:
         JointsManipulationComponent();
         JointsManipulationComponent(
-            const PublisherConfiguration& configuration, const AZStd::unordered_map<AZStd::string, JointPosition>& initialPositions);
+            const PublisherConfiguration& configuration,
+            const AZStd::unordered_map<AZStd::string, JointPosition>& initialPositions,
+            AZStd::vector<AZStd::string> jointNames,
+            AZStd::string positionCommandTopic);
         ~JointsManipulationComponent() = default;
         AZ_COMPONENT(JointsManipulationComponent, "{3da9abfc-0028-4e3e-8d04-4e4440d2e319}", AZ::Component);
 
@@ -70,6 +75,7 @@ namespace ROS2
         void OnTick(float deltaTime, AZ::ScriptTimePoint time) override;
 
         void MoveToSetPositions(float deltaTime);
+        void PositionCommandCallback(std_msgs::msg::Float64MultiArray command);
 
         AZStd::string GetManipulatorNamespace() const;
 
@@ -82,5 +88,8 @@ namespace ROS2
         ManipulationJoints m_manipulationJoints; //!< Map of JointInfo where the key is a joint name (with namespace included)
         AZStd::unordered_map<AZStd::string, JointPosition>
             m_initialPositions; //!< Initial positions where the key is joint name (without namespace included)
+        std::shared_ptr<rclcpp::Subscription<std_msgs::msg::Float64MultiArray>> m_jointPositionSubscriber;
+        AZStd::vector<AZStd::string> m_jointNames;
+        AZStd::string m_positionCommandTopic;
     };
 } // namespace ROS2
--- a/Gems/ROS2/Code/Source/Manipulation/JointsManipulationEditorComponent.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/JointsManipulationEditorComponent.cpp
@@ -7,6 +7,10 @@
  */
 
 #include "JointsManipulationEditorComponent.h"
+#include "AzCore/Debug/Trace.h"
+#include "AzCore/std/containers/unordered_map.h"
+#include "AzCore/std/containers/vector.h"
+#include "AzCore/std/string/string.h"
 #include "JointsManipulationComponent.h"
 #include <AzCore/Component/ComponentApplicationBus.h>
 #include <AzCore/Component/TransformBus.h>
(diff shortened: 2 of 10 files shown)


Gems/ROS2/Code/Source/Manipulation/JointsManipulationComponent.h
```diff
@@ -28,7 +31,10 @@ namespace ROS2
     public:
         JointsManipulationComponent();
         JointsManipulationComponent(
-            const PublisherConfiguration& configuration, const AZStd::unordered_map<AZStd::string, JointPosition>& initialPositions);
+            const PublisherConfiguration& configuration,
+            const AZStd::unordered_map<AZStd::string, JointPosition>& initialPositions,
+            AZStd::vector<AZStd::string> jointNames,
```
michalpelka, line 36: Why is redundant joint names ? We have joint names in the map.

lewandowskikacper: We have joint names in the map, but: 
- they are keys of hashmap, and ordering is important for position control, as the message used is just array of floats

- the order in which the map is filled is BFS, but order of siblings is not specified (Bus::Event is used, where the policy doesn't specify order of handlers). Because of that, if you have eg. multiple gripper joints, they will change they position as you add/remove their siblings.


This was more natural to me, as in controllers.yaml file user would specify the joints as a list.
(This will also be needed if in the future we will want to have separate topics for controlling parts of the robot - e.g. separate controllers for arm/gripper).

If your suggestion is to use the map instead, I can work through the changes necessary to make the ordering more predictable.

adamdbrw: What would these changes be for a map? Alphabetical order of joints might not be desired, perhaps they would typically be indexed from top-level down the kinematic chain?

If alphabetical order is bad, I think that the vector can stay, or that we could map joints to position&index. Perhaps it would be better rename the vector to indicate it is about indexes (m_orderedJointNames or sth)?

Whatever the structure is, we should populate it instead of requiring user manual input, especially when the information that user needs to input is redundant.

Gems/ROS2/Code/Source/Manipulation/JointsManipulationEditorComponent.cpp
```diff
@@ -50,6 +50,17 @@ namespace ROS2
         incompatible.push_back(AZ_CRC_CE("JointsManipulationService"));
     }
 
+    void JointsManipulationEditorComponent::Activate()
+    {
+        if (m_jointOrderedNames.empty())
```
adamdbrw, line 55: Consider someone running the Editor while this component is in the Level, then changing initial positions.
There would be inconsistency between the two collections. This inconsistency can also occur if someone inputs different joint names.
You can handle changes in editable fields by event attributes (such as OnChangeValidate).

But it is much better to have a single collection, see other comment.

Gems/ROS2/Code/Source/Manipulation/JointsManipulationEditorComponent.cpp
```diff
@@ -50,6 +50,17 @@ namespace ROS2
...
+        {
+            for (auto [jointName, jointInfo] : m_initialPositions)
```
adamdbrw, line 57: jointInfo argument name here is misleading

Gems/ROS2/Code/Source/Manipulation/JointsManipulationEditorComponent.h
```diff
@@ -32,10 +32,12 @@ namespace ROS2
         // AzToolsFramework::Components::EditorComponentBase overrides
         void BuildGameEntity(AZ::Entity* gameEntity) override;
 
+        void Activate();
+
     private:
         PublisherConfiguration m_jointStatePublisherConfiguration;
         AZStd::unordered_map<AZStd::string, float> m_initialPositions;
```
adamdbrw, line 39: How about replacing of these two collections with 
AZStd::vector<AZStd::pair<<AZStd::string, float>> and letting user reorder items in UI?
or AZStd::map from index to such pair if it is easier?

lewandowskikacper: Unfortunately, I had problems with reordering elements of vectors: https://github.com/o3de/o3de/issues/16648.

lewandowskikacper: There is a `set<{name, position, index}>`. It keeps all the information in one place and has nice input method: 
image


But because of that I see no way of straightforward migration of prefab overrides (as it would require knowledge of the prefab).

lewandowskikacper:
Opened new Pr from RobotecAI's fork: https://github.com/o3de/o3de-extras/pull/553

PR #457, opened 2023-08-10
Add the concept of global settings to the SdfAssetBuilder

mbalfour-amzn:
This PR changes the SdfAssetBuilder to read in global builder settings from a setreg file to apply as the defaults when building all assets. Changes to the settings in the setreg file will cause all affected assets to rebuild the next time the AP is run. Currently, the only global setting that exists is "UseArticulations", but other settings can be added easily.

The mechanism used for rebuilding the assets is the builder "fingerprint". Any time the fingerprint changes, the builder will rebuild all of its assets. The global settings are read in from the setreg file and then serialized back out into a JSON string that is used as the fingerprint. Changes to the settings changes the fingerprint string, which triggers the rebuild.

This PR also has a separate change to the ROS2 Gem to enable .dae and .obj files in the Scene Builder by default. These are common file types for robotics, so we want the scene builder to attempt to convert them. Note that enabling them also requires a change to the AssetProcessor settings to list the .assetinfo files for those new types - that setting is used to ensure that the asset reprocesses whenever its .assetinfo file changes.

### DIFF
--- a/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
+++ b/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
@@ -32,17 +32,24 @@ namespace ROS2
         {
             [[maybe_unused]] constexpr const char* SdfAssetBuilderName = "SdfAssetBuilder";
             constexpr const char* SdfAssetBuilderJobKey = "Sdf Asset Builder";
-            constexpr const char* SdfAssetBuilderSupportedFileExtensionsRegistryKey = "/O3DE/ROS2/SdfAssetBuilder/SupportedFileTypeExtensions";
         }
 
     SdfAssetBuilder::SdfAssetBuilder()
     {
+        // Read in all of the global settings from the settings registry.
+        m_globalSettings.LoadGlobalSettings();
+
+        // Turn our global settings into a cached fingerprint that we'll use on every job
+        // so that we can detect when to rebuild assets on global setting changes.
+        m_fingerprint = GetFingerprint();
+
         AssetBuilderSDK::AssetBuilderDesc sdfAssetBuilderDescriptor;
 
         sdfAssetBuilderDescriptor.m_name = SdfAssetBuilderJobKey;
         sdfAssetBuilderDescriptor.m_version = 1; // bump this to rebuild all sdf files
         sdfAssetBuilderDescriptor.m_busId = azrtti_typeid<SdfAssetBuilder>();
-        sdfAssetBuilderDescriptor.m_patterns = GetSupportedBuilderPatterns();
+        sdfAssetBuilderDescriptor.m_patterns = m_globalSettings.m_builderPatterns;
+        sdfAssetBuilderDescriptor.m_analysisFingerprint = m_fingerprint; // set the fingerprint to the global settings
 
         sdfAssetBuilderDescriptor.m_createJobFunction = [this](auto && request, auto && response) 
             { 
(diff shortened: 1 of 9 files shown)


Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.h
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include <AssetBuilderSDK/AssetBuilderSDK.h>
+
+namespace ROS2
+{
+    struct SdfAssetBuilderSettings
+    {
+    public:
+        AZ_RTTI(ROS2::SdfAssetBuilderSettings, "{7FE3FA92-E274-4624-8905-61E1587DDD30}");
```
lumberyard-employee-dm, line 18: The `ROS2::` should be removed from the AZ_RTTI macro.
It complicates creating an SdfAssetBuilderSettings object if it was ever reflected to the BehaviorContext as well if we ever need to change the Type Info name from the C++ type
```suggestion
        AZ_RTTI(SdfAssetBuilderSettings, "{7FE3FA92-E274-4624-8905-61E1587DDD30}");
```

Gems/ROS2/Registry/sceneassetimporter.setreg
```diff
@@ -0,0 +1,21 @@
...
+					".stl",
+					".gltf",
+					".glb",
+					".obj"
+				]
+			}
+		}
+	}
+}
+
```
lumberyard-employee-dm, line 21: There are tabs in this file
```suggestion
{
    "O3DE":
    {
        "SceneAPI":
        {
            "AssetImporter":
            {
                "SupportedFileTypeExtensions":
                [
                    ".dae",
                    ".fbx",
                    ".stl",
                    ".gltf",
                    ".glb",
                    ".obj"
                ]
            }
        }
    }
}

```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
```diff
@@ -0,0 +1,83 @@
...
+        auto VisitFileTypeExtensions = [&settingsRegistry, this]
+            (const AZ::SettingsRegistryInterface::VisitArgs& visitArgs)
+            {
+                if (AZ::SettingsRegistryInterface::FixedValueString value;
+                    settingsRegistry->Get(value, visitArgs.m_jsonKeyPath))
+                {
+                    // Ignore any entries that are either completely empty or *only* contain a '.'.
+                    // These will produce excessive (and presumably incorrect) wildcard matches.
+                    if (value.empty() ||
+                        ((value.size() == 1) && value.starts_with('.')))
```
lumberyard-employee-dm, line 62: There is a simpler check for a dot and that is just comparing to `.`.
```suggestion
                    if (value.empty() ||
                        value == '.')
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -184,6 +147,25 @@ namespace ROS2
...
+        // we'll still be able to detect that the settings don't match what they used to be.
+        AZ::JsonSerializerSettings jsonSettings;
+        jsonSettings.m_keepDefaults = true;
+
+        [[maybe_unused]] AZ::Outcome<void, AZStd::string> saveObjectResult =
+            AZ::JsonSerializationUtils::SaveObjectToStream(&m_globalSettings, stream, {}, &jsonSettings);
+        AZ_Assert(saveObjectResult.IsSuccess(), "Failed to save settings to fingerprint string: %s", 
+            saveObjectResult.GetError().c_str());
+
+        return settingsString;
```
spham-amzn, line 166: The fingerprint is the entire settings string? not a hash?

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
```diff
@@ -0,0 +1,83 @@
...
+             ;
+        }
+    }
+
+    void SdfAssetBuilderSettings::LoadGlobalSettings()
+    {
+        auto settingsRegistry = AZ::SettingsRegistry::Get();
+        if (settingsRegistry == nullptr)
+        {
+            AZ_Error(SdfAssetBuilderName, false, "Settings Registry not found, Sdf Asset Builder settings will use defaults.");
```
spham-amzn, line 44: If there is a fallback (the defaults), then shouldn't this be a warning? I thought AZ_Error will raise an assertion.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
```diff
@@ -0,0 +1,83 @@
...
+            return;
+        }
```
lumberyard-employee-dm, line 46: Instead of this being only available for global settings, instead we can supply a `SettingsRegistryInterace*` pointer as a parameter and have it default to AZ::SettingsRegistry::Get().

If we do that, we can rename this function to `LoadSettings`
```suggestion
    void SdfAssetBuilderSettings::LoadGlobalSettings(AZ::SettingsRegistryInterface* settingsRegistry)
    {
        if (settingsRegistry == nullptr)
        {
            AZ_Error(SdfAssetBuilderName, false, "Settings Registry not found, Sdf Asset Builder settings will use defaults.");
            return;
        }
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
mbalfour-amzn: AZ_Error doesn't raise an assert, AZ_Assert does. It probably _ought_ to be an assert though, because a missing settings registry is a programming failure, not a data failure.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
mbalfour-amzn: lol, um, yeah. thanks!

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
spham-amzn: I am fine with it the way it is, its just that reading `fingerprint` implies a hash.

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.h
```diff
@@ -0,0 +1,40 @@
...
+        virtual ~SdfAssetBuilderSettings() = default;
+
+        static void Reflect(AZ::ReflectContext* context);
+
+        //! Read in the builder settings from the settings registry.
+        //! This is done as a separate step from the constructor so that serialization has
+        //! the ability to create a default set of values by using the default constructor.
+        //! If we read these in the constructor, serialization would see all of the current values
+        //! as default values and would try to prune them from the output by default.
+        void LoadSettings(AZ::SettingsRegistryInterface* settingsRegistry);
```
lumberyard-employee-dm, line 35: You could have settingsRegistry parameter here be default value of `AZ::SettingsRegistry::Get()` to simplify call sites that want to use the Global registry.
```suggestion
        void LoadSettings(AZ::SettingsRegistryInterface* settingsRegistry = AZ::SettingsRegistry::Get());
```

lumberyard-employee-dm (approved):

spham-amzn (approved):

Issue #414, opened 2023-07-17
Handle ros exceptions


adamdbrw:
In general, we should avoid crashes due to unhandled ROS exception.

The task is to analyze where such exception occur and what the proper handling should be (AZ_Error etc.).

michalpelka:
As discussed in the issue triage meeting, the issue is being resolved.

aniafafe:
After deeper analysis, I've recognised two main situations when the ROS exception can occur:

###  Nonrecoverable critical error:

When a ROS exception occurs that halts further processing and offers no recourse for the user, such scenarios should not be allowed to continue. In these cases, the program isn't fit to function correctly, so it should terminate immediately as it does currently. For instance:

Creating Node or Executors in Activate method of _ROS 2 System Component_:
```cpp
void ROS2SystemComponent::Activate()
{
    InitClock();
    m_simulationClock->Activate();
    m_ros2Node = std::make_shared<rclcpp::Node>("o3de_ros2_node");
    m_executor = AZStd::make_shared<rclcpp::executors::SingleThreadedExecutor>();
    m_executor->add_node(m_ros2Node);
```
Given that ROS 2 components rely on this system component, its failure renders most engine actions unworkable, putting the engine in a critical state.

###  Noncritical recoverable error:

However, for ROS exceptions impacting only a limited number of components (often just one), I propose adding a component status. This status would indicate if the component is functioning correctly or if it wasn't initialized properly due to an exception. We could handle these exceptions, adjusting the component status accordingly. For example:
```cpp
void ROS2ProximitySensor::Activate()
{
    try
    {
        auto ros2Node = ROS2Interface::Get()->GetNode();
        AZ_Assert(
            m_sensorConfiguration.m_publishersConfigurations.size() == 1, "Invalid configuration of publishers for proximity sensor");

        const TopicConfiguration& publisherConfig = m_sensorConfiguration.m_publishersConfigurations["std_msgs::msg::bool"];
        AZStd::string fullTopic = ROS2Names::GetNamespacedName(GetNamespace(), publisherConfig.m_topic);

        m_detectionPublisher = ros2Node->create_publisher<std_msgs::msg::Bool>(fullTopic.data(), publisherConfig.GetQoS());
    } catch (std::exception& e)
    {
        m_componentStatus = State::ERROR;
        AZ_Error("ROS2ProximitySensor", false, "The Activate failed %s", e.what());
        return;
    }
```
[...]

PR #459, opened 2023-08-16
 handle ROS 2 exceptions

aniafafe:
fixes: #414
- Add validation for user-provided ROS 2 names
- Add exception handling in non-critical cases

Handling for critical errors will be implemented.


aniafafe:
@adamdbrw @michalpelka Please review the changes made for some non-critical errors:
- Incorrect names of ROS topics, services, etc. provided by the user
- States of ROS components (publishing and activation)"

Gems/ROS2/Code/Include/ROS2/Utilities/ROS2ErrorHandler.h
```diff
@@ -0,0 +1,40 @@
...
+        void SetComponentStateError();
+
+        //! Executes function @param ros2Operation and catches possible exceptions caused by it.
+        //! When an exception occurs, it prints an error message on the console
+        //! containing @param componentName, @param contextMessage and error message.
+        void ExecuteROS2Context(std::function<void()> ros2Operation, const char* componentName, const char* contextMessage);
+
+    private:
+        // Describes component's state based on whether a ROS 2 error has occurred.
+        enum class ROS2ComponentState
```
michalpelka, line 33: Why not use here Outcome pattern?

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
```diff
@@ -79,11 +80,21 @@ namespace ROS2
 
     void ROS2ImuSensorComponent::SetupRefreshLoop()
     {
+        if (!IsComponentValid())
```
michalpelka, line 83: I thinkl this pattern adding another level of complication. We can simply check if publisher is null.

Gems/ROS2/Code/Source/ProximitySensor/ROS2ProximitySensor.h
```diff
@@ -10,14 +10,17 @@
...
 #include <std_msgs/msg/bool.hpp>
 
 namespace ROS2
 {
     //! Simple proximity sensor based on raycasting
     //! This component publishes a bool topic depending on the object presence
-    class ROS2ProximitySensor : public ROS2SensorComponent
+    class ROS2ProximitySensor
+        : public ROS2SensorComponent
+        , public ROS2ErrorHandler
```
michalpelka, line 23: what about multi topic sensors such as camera? Idea of solving this issue through polymorphism is interesting one, but makes integration with ROS 2 not as seamless as before. Let us think about wrapping publisher code into some C-style macros with __VA_ARGS__ ?
Another approach would be to use try - catch pattern directly and hadnling those inside component's code. 
What are you thoughts @aniafafe . Since it is architecture decisions I would kindly ask @adamdbrw for his input.

Gems/ROS2/Code/Include/ROS2/Utilities/ROS2ErrorHandler.h
aniafafe: The reason for introducing the class is to represent the state of the entire component (perhaps the class name could be more descriptive). After every ROS2 exception, we reach a state where we can't proceed with any further actions. Calling subsequent methods in such a state usually results in undefined behavior. To adhere to the DRY principle and to isolate common error handling, it seems sensible to create a new class that can execute ROS2 functions that may throw exceptions and set the component's state.

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
aniafafe: IMHO, this pattern significantly simplifies everything. It boils down to checking if the component is in the proper state or not (always the same check). It doesn't matter if it's publisher number 1 or number X that's in an invalid condition, as producing partially correct or non-deterministic output can be really tricky and difficult to debug for output receivers.

Gems/ROS2/Code/Source/ProximitySensor/ROS2ProximitySensor.h
aniafafe: I'm certainly open to all proposals, but I genuinely dislike macros. In most situations, they are challenging to debug and test, and in my opinion, they should remain in the C language. Also, I don't see any issue with managing a more complex component like the multi-sensor camera you mentioned. As I pointed out in my previous comment, the proposed ROS2ErrorHandler class is designed to represent the entire component's state, not just a single publisher or any other individual entity that might cause errors.

Gems/ROS2/Code/Source/Communication/PublisherConfiguration.cpp
```diff
@@ -30,6 +31,7 @@ namespace ROS2
                         AZ::Edit::UIHandlers::Default, &PublisherConfiguration::m_publish, "Publish", "Whether the publisher is on")
                     ->DataElement(
                         AZ::Edit::UIHandlers::Default, &PublisherConfiguration::m_topicConfiguration, "Topic", "Topic configuration")
+                    ->Attribute(AZ::Edit::Attributes::ChangeValidate, &ROS2Names::ValidateFullTopicField)
```
adamdbrw, line 34: Can we guarantee that this does not impact UX negatively? 
For example, if the namespace is invalid, the validate error will persist whatever the input to PublisherConfiguration, which is a bad experience for the user.

If we guarantee that namespace part will always be valid here, then it is fine.

Gems/ROS2/Code/Source/Communication/TopicConfiguration.cpp
```diff
@@ -29,7 +30,7 @@ namespace ROS2
                     ->DataElement(AZ::Edit::UIHandlers::Default, &TopicConfiguration::m_type, "Type", "Type of topic messages")
                     ->Attribute(AZ::Edit::Attributes::ReadOnly, true)
                     ->DataElement(AZ::Edit::UIHandlers::Default, &TopicConfiguration::m_topic, "Topic", "Topic with no namespace")
-                    ->Attribute(AZ::Edit::Attributes::ChangeValidate, &ROS2Names::ValidateTopicField)
+                    ->Attribute(AZ::Edit::Attributes::ChangeValidate, &ROS2Names::ValidateFullTopicField)
```
adamdbrw, line 33: See the comment above

Gems/ROS2/Code/Source/ContactSensor/ROS2ContactSensorComponent.cpp
```diff
@@ -24,6 +24,7 @@
 #include <ROS2/ROS2GemUtilities.h>
 #include <ROS2/Utilities/ROS2Conversions.h>
 #include <ROS2/Utilities/ROS2Names.h>
+#include <exception>
```
adamdbrw, line 27: Does not seem to be used

Gems/ROS2/Code/Source/ContactSensor/ROS2ContactSensorComponent.cpp
```diff
@@ -71,12 +72,18 @@ namespace ROS2
         AZ::Entity* entity = nullptr;
         AZ::ComponentApplicationBus::BroadcastResult(entity, &AZ::ComponentApplicationRequests::FindEntity, m_entityId);
         m_entityName = entity->GetName();
-
         auto ros2Node = ROS2Interface::Get()->GetNode();
         AZ_Assert(m_sensorConfiguration.m_publishersConfigurations.size() == 1, "Invalid configuration of publishers for Contact sensor");
         const auto publisherConfig = m_sensorConfiguration.m_publishersConfigurations["gazebo_msgs::msg::ContactsState"];
         const auto fullTopic = ROS2Names::GetNamespacedName(GetNamespace(), publisherConfig.m_topic);
-        m_contactsPublisher = ros2Node->create_publisher<gazebo_msgs::msg::ContactsState>(fullTopic.data(), publisherConfig.GetQoS());
+        ExecuteROS2Context(
```
adamdbrw, line 79: I see some drawbacks to this construct:
- It requires the developer to know about it, as opposed to doing things as they would normally do in ROS. It's boilerplate code for each (exception-throwing) ROS call, removing users from the "ROS developer experience", which is among the most important selling points for O3DE.
- The developer might not use it (in their sensor, project) and the engine will still crash.
- There is no space to be more specific about exceptions to catch instead of all std::exceptions (are we sure we know that the best way to handle them all is the same?). (https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/exceptions/exceptions.hpp) for rclcpp.
- There is no indication on what to do if it throws (behavior is binary). I imagine that someone could make a ROS call that is not fatal if exception is thrown, including accessing some parameters or creating optional/diagnostic publisher. We might want to just warn in some cases.

Gems/ROS2/Code/Source/ContactSensor/ROS2ContactSensorComponent.cpp
```diff
@@ -112,6 +119,11 @@ namespace ROS2
 
     void ROS2ContactSensorComponent::FrequencyTick()
     {
+        if (!IsComponentValid())
```
adamdbrw, line 122: Is there a reason why this check is preferred to just having the Component Deactivate?

In any case, I prefer to have it in one place (depending on solution, ROS2SensorComponent, or a global handler).

Gems/ROS2/Code/Source/Utilities/ROS2ErrorHandler.cpp
```diff
@@ -0,0 +1,37 @@
...
+    {
+        return m_componentState == ROS2ComponentState::Ok;
+    }
+
+    void ROS2ErrorHandler::SetComponentStateError()
+    {
+        m_componentState = ROS2ComponentState::Error;
+    }
+
+    void ROS2ErrorHandler::ExecuteROS2Context(std::function<void()> ros2Operation, const char* componentName, const char* contextMessage)
```
adamdbrw, line 24: use AZStd::string instead of character strings (const time size, compatible with standard algorithms and signatures of utilities such as string format, guaranteed to be valid etc).

Gems/ROS2/Code/Source/Utilities/ROS2ErrorHandler.cpp
```diff
@@ -0,0 +1,37 @@
...
+    {
+        try
+        {
+            ros2Operation();
+        }
+        catch (std::exception& e)
+        {
+            AZStd::string errorInfo = AZStd::string(contextMessage) + " " + e.what();
```
adamdbrw, line 32: use AZ version of string::format

adamdbrw (commented):
Lets discuss possible solutions!

PR #461, opened 2023-08-17
Small Conveyor Belt Component fix

alek-kam-robotec-ai:
Fixes the error described by this message:

[Error] (Serialize) - Programmer error:
Classes deriving from AZ::Component are not declaring their base class to SerializeContext.
This will cause unexpected behavior such as components shifting around, or duplicating themselves.
Affected components:
- ConveyorBeltComponent {B7F56411-01D4-48B0-8874-230C58A578BD}

Reflection code should look something like this:
serializeContext->Class<MyComponent, AZ::Component, ... (other base classes, if any) ...>()
Make sure the Reflect function is called for all base classes as well.


michalpelka:
duplication: 452

PR #469, opened 2023-08-24
Reverted PR #400, fixes to URDF importer

michalpelka:
Revert #400 using old better logic.
The #400 was used to circle navigate issue in asset processor. 
Also missing case for resolving package was added.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
@@ -291,16 +291,20 @@ namespace ROS2::Utils
         AZ::Crc32 urdfFileCrc;
         urdfFileCrc.Add(urdfFilename);
         const AZ::IO::Path urdfPath(urdfFilename);
-
+        const AZStd::string directoryNameTmp = AZStd::string::format("$tmp_%u.tmp", AZ::u32(urdfFileCrc));
         const auto directoryNameDst = AZ::IO::FixedMaxPathString::format(
             "%u_%.*s%.*s", AZ::u32(urdfFileCrc), AZ_PATH_ARG(urdfPath.Stem()), AZ_STRING_ARG(outputDirSuffix));
 
+        const AZ::IO::Path importDirectoryTmp = AZ::IO::Path(AZ::Utils::GetProjectPath()) / "Assets" / "UrdfImporter" / directoryNameTmp;
         const AZ::IO::Path importDirectoryDst = AZ::IO::Path(AZ::Utils::GetProjectPath()) / "Assets" / "UrdfImporter" / directoryNameDst;
 
+        fileIO->DestroyPath(importDirectoryTmp.c_str());
         const auto outcomeCreateDstDir = fileIO->CreatePath(importDirectoryDst.c_str());
+        const auto outcomeCreateTmpDir = fileIO->CreatePath(importDirectoryTmp.c_str());
         AZ_Error("CopyAssetForURDF", outcomeCreateDstDir, "Cannot create destination directory : %s", importDirectoryDst.c_str());
+        AZ_Error("CopyAssetForURDF", outcomeCreateTmpDir, "Cannot create temporary directory : %s", importDirectoryTmp.c_str());
 
-        if (!outcomeCreateDstDir)
+        if (!outcomeCreateDstDir || !outcomeCreateTmpDir)
         {
             return urdfAssetMap;
         }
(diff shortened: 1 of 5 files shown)


jhanca-robotecai (approved):
The code works as expected.

Screenshot from 2023-08-25 10-29-17

adamdbrw (approved):

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -291,16 +291,20 @@ namespace ROS2::Utils
         AZ::Crc32 urdfFileCrc;
         urdfFileCrc.Add(urdfFilename);
         const AZ::IO::Path urdfPath(urdfFilename);
-
+        const AZStd::string directoryNameTmp = AZStd::string::format("$tmp_%u.tmp", AZ::u32(urdfFileCrc));
```
mbalfour-amzn, line 294:
```suggestion
        // By naming the temp directory '$tmp_*', the default configuration in AssetProcessorPlatformConfig.setreg will
        // exclude these files from processing.
        const AZStd::string directoryNameTmp = AZStd::string::format("$tmp_%u.tmp", AZ::u32(urdfFileCrc));
```

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -318,26 +322,46 @@ namespace ROS2::Utils
...
+                            "CopyAssetForURDF",
+                            "Move %s to %s, result: %d",
+                            targetPathAssetTmp.c_str(),
+                            targetPathAssetDst.c_str(),
+                            outcomeMoveDst.GetResultCode());
+                        if (outcomeMoveDst)
+                        {
+                            copiedFiles[unresolvedUrfFileName] = targetPathAssetDst.String();
+                        }
+                    };
```
mbalfour-amzn, line 364: There might be a race condition here where the AP will start processing the first copied file before the .assetinfo file is copied, which I think would cause an initial export without the .assetinfo to occur before exporting a second time with the settings from the .assetinfo. It might be avoidable by copying the assetinfo first, and the real file second.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -291,16 +291,20 @@ namespace ROS2::Utils
...
         const auto outcomeCreateDstDir = fileIO->CreatePath(importDirectoryDst.c_str());
+        const auto outcomeCreateTmpDir = fileIO->CreatePath(importDirectoryTmp.c_str());
         AZ_Error("CopyAssetForURDF", outcomeCreateDstDir, "Cannot create destination directory : %s", importDirectoryDst.c_str());
+        AZ_Error("CopyAssetForURDF", outcomeCreateTmpDir, "Cannot create temporary directory : %s", importDirectoryTmp.c_str());
 
-        if (!outcomeCreateDstDir)
+        if (!outcomeCreateDstDir || !outcomeCreateTmpDir)
         {
             return urdfAssetMap;
         }
```
mbalfour-amzn, line 310: If either directory was created, this really ought to destroy it again before returning.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -318,26 +322,46 @@ namespace ROS2::Utils
             }
 
             AZ::IO::Path resolvedPath(resolved);
+            const bool needsVisual = visuals.contains(unresolvedUrfFileName);
             const bool needsCollider = colliders.contains(unresolvedUrfFileName);
 
             const AZStd::string pathSuffix = needsCollider ? "_collider" : "";
             const AZStd::string newFilename = resolvedPath.Stem().String() + pathSuffix + resolvedPath.Extension().String();
```
mbalfour-amzn, line 329: Now that the exported model files retain the file suffix (ex: "model.dae" -> "model.dae.azmodel"), is the pathSuffix still needed here?

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -318,26 +322,46 @@ namespace ROS2::Utils
...
+
+                    if (assetInfoOk)
+                    {
+                        const auto outcomeMoveDst = fileIO->Copy(targetPathAssetTmp.c_str(), targetPathAssetDst.c_str());
+                        AZ_Printf(
+                            "CopyAssetForURDF",
+                            "Move %s to %s, result: %d",
+                            targetPathAssetTmp.c_str(),
+                            targetPathAssetDst.c_str(),
+                            outcomeMoveDst.GetResultCode());
```
lumberyard-employee-dm, line 359: Actually for speed and atomicity, it would be better to use the [FileIOBase::Rename](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/IO/FileIO.h#L157-L160) function here to perform the equivalent of a `mv` command on.

That way there is no danger of the AP pickup a partially cppied assetinfo file as it is being copied.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: Thanks, will do!

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: Wtih recent changes to o3de path suffix is not needed anymore. happy to remove this hack.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: Great suggestion! Thanks!

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: Thanks for a suggestion @mbalfour-amzn. But I have to disagree. 
Asset info is not copied, it is created in place using  `targetPathAssetInfo`.

So in the moment of creation of assetinfo:
 - the mesh file is still in the temp directory (ignored by AP)
 - `.assetinfo` is created in destination directory.

 Next step is to move (no copy as @lumberyard-employee-dm suggested) mesh file from temp directory to the destination directory.
 AP should pick up the mesh file with `assetinfo` already in place.

michalpelka: Of course 
```
CreateSceneManifest(targetPathAssetTmp.String(), targetPathAssetInfo.String(), needsCollider, needsVisual);
``` runs synchronously.

Comment (michalpelka) [reply Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp:364]: I also found some slight possibility for a race in another place. I've added `WaitForAPAndEscalate` function.
There is a call 'GetAvailableAssetInfo' that collects some info about source assets (mainly uids). To make sure that result will be complete, I proposed a function `WaitForAPAndEscalate` that will wait for AP to start report any status on freshly moved source asset.

**Comment by michalpelka** [comment]
Thank you @lumberyard-employee-dm,  @mbalfour-amzn, @adamdbrw for your review. 
I agree with all suggestions, I've left one conversation open to discuss further race condition with @mbalfour-amzn  .

File: Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp [hunk 6]
```diff
@@ -267,6 +267,30 @@ namespace ROS2::Utils
...
+                return result;
+            }
+        }
+        AZ_Warning(
+            "WaitForAPAndEscalate",
+            false,
+            "Cannot find job for %s, import may be incomplete or AssetProcessor is not running.",
+            relativePath.c_str());
+        return result;
+    }
```
mbalfour-amzn, line 292: Rather than calling this request bus in a tight loop, CompileAssetSync will attempt to immediately compile the asset and block until there is a result.
```
                    AzFramework::AssetSystem::AssetStatus status = AzFramework::AssetSystem::AssetStatus_Unknown;
                    AzFramework::AssetSystemRequestBus::BroadcastResult(
                        status, &AzFramework::AssetSystemRequestBus::Events::CompileAssetSync_FlushIO, relativePath);
```

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
mbalfour-amzn: Sorry, I misread the logic here, thanks for the correction. You're absolutely right. 

I've got a suggestion below for how to improve WaitForAPAndEscalate though.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
michalpelka: Maybe we should consider `GetAssetStatusById_FlushIO` here?

michalpelka:
I've resolved last issue, hope PR can be approved.

mbalfour-amzn (approved):

Issue #476, opened 2023-08-28
URDF importer a lot of limitations are itroduced with libSDF


michalpelka:
LibSDF introduced multiple limitations that makes a population of importable URDFs smaller. 
On one hand it enables more predictable experience after import, but blocks import of URDF at all. 
Here is a list of issue I found:
## Error Code 19
```
Error Code 19: Msg: urdf2sdf: link[panda_link4] has no <inertial> block defined. Please ensure this link has a valid mass to prevent any missing links or joints in the resulting SDF model.
```
URDF to reproduce:
```
<?xml version="1.0"?>
<robot name="urdf_test">
    <link name="box_link">
        <visual>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <box size="1 2 1"/>
            </geometry>
        </visual>
        <material name="lightgrey"/>
       <collision>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <box size="1 2 1"/>
            </geometry>
        </collision>
    </link>
</robot>
```

Explanation:
This error is missing inertia link in URDF.

## Error 2
```
ErrorCode=2, Message=Frame with name [LF_HAA] in model with name [anymal] has a name collision. Please rename this frame.
```
URDF to reproduce :
```
<?xml version="1.0"?>
<robot name="urdf_test">    
    <link name="box">
        <inertial>
        <origin xyz="0. 0. 0."/>
        <mass value="1."/>
        <inertia ixx="1." ixy="0." ixz="0." iyy="1." iyz="0." izz="1."/>
        </inertial>
        <visual>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <box size="1 2 1"/>
            </geometry>
        </visual>
        <material name="lightgrey"/>
        <collision>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <box size="1 2 1"/>
            </geometry>
        </collision>
    </link>

    <link name="sphere">
        <inertial>
        <origin xyz="0. 0. 0."/>
        <mass value="1."/>
        <inertia ixx="1." ixy="0." ixz="0." iyy="1." iyz="0." izz="1."/>
        </inertial>
        <visual>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <sphere radius="2"/>
            </geometry>
        </visual>
        <material name="lightgrey"/>
        <collision>
[...]

Issue #472, opened 2023-08-26
ROS2 Robot Importer cannot import the MoveIt Resouces Robot Models


lumberyard-employee-dm:
As brought up in PR #460 switching to using libsdformat has issues importing some URDF files, that URDFDom is able to import successfully https://github.com/o3de/o3de-extras/pull/460#pullrequestreview-1595595506

Investigate problems importing [panda franka and fanuc](https://github.com/ros-planning/moveit_resources) robots with SDFormat library; double-check the warnings triggered by  [pr2](https://github.com/ros-planning/moveit_resources) import

lgleim:
This issue appears to be resolved by https://github.com/o3de/o3de-extras/pull/481 @michalpelka ?

jhanca-robotecai:
I can confirm the robots are successfully imported with the fix in #481. This issue can be closed.

![image](https://github.com/o3de/o3de-extras/assets/134940295/4561a48a-80a0-4251-aab0-8c56834e8fed)

michalpelka:
@lgleim issue should be closed.
Thanks, @jhanca-robotecai for this test.

PR #481, opened 2023-08-29
Try to fix urdfs in memory before SDF import

michalpelka:
Some proposition to solve #476  and #472.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
@@ -8,11 +8,12 @@
 
 #include "FixURDF.h"
 #include <AzCore/XML/rapidxml_print.h>
-#include <AzCore/std/containers/set.h>
+#include <AzCore/std/containers/unordered_set.h>
 #include <iostream>
+
 namespace ROS2::Utils
 {
-    AZStd::vector<AZStd::string> AddMissingInertiaToLink(AZ::rapidxml::xml_node<>* urdf)
+    AZStd::vector<AZStd::string> AddMissingInertiaToLinks(AZ::rapidxml::xml_node<>* urdf)
     {
         AZStd::vector<AZStd::string> modifiedLinks;
         using namespace AZ::rapidxml;
@@ -67,7 +68,7 @@ namespace ROS2::Utils
     {
         using namespace AZ::rapidxml;
         AZStd::vector<AZStd::string> modifiedLinks;
-        AZStd::set<AZStd::string> linkAndJointsName;
+        AZStd::unordered_set<AZStd::string> linkAndJointsName;
         for (xml_node<>* link = urdf->first_node("link"); link; link = link->next_sibling("link"))
         {
             auto* name = link->first_attribute("name");
@@ -86,7 +87,7 @@ namespace ROS2::Utils
                     auto newName = AZStd::string(name->value()) + "_joint_dup";
                     name->value(urdf->document()->allocate_string(newName.c_str()));
                     linkAndJointsName.insert(newName);
-                    modifiedLinks.push_back(newName);
+                    modifiedLinks.push_back(AZStd::move(newName));
                 }
                 else
                 {
@@ -104,7 +105,7 @@ namespace ROS2::Utils
         xml_document<> doc;
         doc.parse<0>(const_cast<char*>(data.c_str()));
         xml_node<>* urdf = doc.first_node("robot");
-        auto links = AddMissingInertiaToLink(urdf);
+        auto links = AddMissingInertiaToLinks(urdf);
 
         if (links.size())
         {
--- a/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
@@ -1,10 +1,10 @@
 /*
-* Copyright (c) Contributors to the Open 3D Engine Project.
-* For complete copyright and license terms please see the LICENSE at the root of this distribution.
-*
-* SPDX-License-Identifier: Apache-2.0 OR MIT
-*
-*/
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
 
 #include "FixURDF.h"
 #include <AzCore/XML/rapidxml_print.h>
@@ -25,9 +25,8 @@ namespace ROS2::Utils
             }
         }
 
-        for (auto * link : links)
+        for (auto* link : links)
         {
-
             xml_node<>* inertial = urdf->document()->allocate_node(node_element, "inertial");
 
             xml_node<>* mass = urdf->document()->allocate_node(node_element, "mass");
@@ -59,7 +58,6 @@ namespace ROS2::Utils
                 modifiedLinks.push_back(name->value());
             }
             link->append_node(inertial);
-
         }
 
         return modifiedLinks;
@@ -90,17 +88,18 @@ namespace ROS2::Utils
                     linkAndJointsName.insert(newName);
                     modifiedLinks.push_back(newName);
                 }
-                else{
+                else
+                {
                     linkAndJointsName.insert(name->value());
                 }
-
             }
         }
         return modifiedLinks;
     }
 
-    std::string ModifyURDFInMemory(const std::string& data)
+    AZStd::pair<std::string, AZStd::vector<AZStd::string>> ModifyURDFInMemory(const std::string& data)
     {
+        AZStd::vector<AZStd::string> modifiedElements;
         using namespace AZ::rapidxml;
         xml_document<> doc;
         doc.parse<0>(const_cast<char*>(data.c_str()));
@@ -112,7 +111,8 @@ namespace ROS2::Utils
             AZ_Warning("ROS2", false, "Added missing inertia to links: ");
             for (auto& link : links)
             {
-                    AZ_Warning("ROS2", false, "  -> %s", link.c_str());
+                modifiedElements.push_back(link);
+                AZ_Warning("ROS2", false, "  -> %s", link.c_str());
             }
         }
 
@@ -122,13 +122,14 @@ namespace ROS2::Utils
             AZ_Warning("ROS2", false, "Renamed duplicated links and joints: ");
             for (auto& dup : renames)
             {
-                    AZ_Warning("ROS2", false, "  -> %s", dup.c_str());
+                modifiedElements.push_back(dup);
+                AZ_Warning("ROS2", false, "  -> %s", dup.c_str());
             }
         }
 
         std::string xmlDocString;
         AZ::rapidxml::print(std::back_inserter(xmlDocString), *urdf, 0);
 
-        return xmlDocString;
+        return { xmlDocString, modifiedElements };
     }
 } // namespace ROS2::Utils
--- a/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
@@ -7,12 +7,16 @@
  */
 
 #include "FixURDF.h"
+#include <AzCore/XML/rapidxml.h>
 #include <AzCore/XML/rapidxml_print.h>
-#include <AzCore/std/containers/unordered_set.h>
+#include <AzCore/std/containers/unordered_map.h>
 #include <iostream>
 
 namespace ROS2::Utils
 {
+    //! Modifies a parsed URDF in memory to add missing inertia to links, which prevents SDF error 19.
+    //! @param urdf URDF to modify.
+    //! @returns a list of names of links that were modified.
     AZStd::vector<AZStd::string> AddMissingInertiaToLinks(AZ::rapidxml::xml_node<>* urdf)
     {
         AZStd::vector<AZStd::string> modifiedLinks;
@@ -64,17 +68,22 @@ namespace ROS2::Utils
         return modifiedLinks;
     }
 
-    AZStd::vector<AZStd::string> ChangeDuplications(AZ::rapidxml::xml_node<>* urdf)
+    //! Handles a case of multiple joints and the link sharing a common names which causes SDF error2 (but is fine in URDF)
+    //! Function will add a suffix "_dup" to the name of the joint if it is also the name of a link.
+    //! If there are name collisions in links, this will not be able to fix it, the SDF parser will throw an error.
+    //! @param urdf URDF to modify.
+    //! @returns a list of links that were modified
+    AZStd::vector<AZStd::string> RenameDuplicatedJoints(AZ::rapidxml::xml_node<>* urdf)
     {
         using namespace AZ::rapidxml;
         AZStd::vector<AZStd::string> modifiedLinks;
-        AZStd::unordered_set<AZStd::string> linkAndJointsName;
+        AZStd::unordered_map<AZStd::string, unsigned int> linkAndJointsName;
         for (xml_node<>* link = urdf->first_node("link"); link; link = link->next_sibling("link"))
         {
             auto* name = link->first_attribute("name");
             if (name)
             {
-                linkAndJointsName.insert(name->value());
+                linkAndJointsName.insert(AZStd::make_pair(name->value(), 0));
             }
         }
         for (xml_node<>* joint = urdf->first_node("joint"); joint; joint = joint->next_sibling("joint"))
@@ -84,14 +93,15 @@ namespace ROS2::Utils
             {
                 if (linkAndJointsName.contains(name->value()))
                 {
-                    auto newName = AZStd::string(name->value()) + "_joint_dup";
+                    unsigned int& count = linkAndJointsName[name->value()];
+                    auto newName = AZStd::string::format("%s_dup%u", name->value(), count);
                     name->value(urdf->document()->allocate_string(newName.c_str()));
-                    linkAndJointsName.insert(newName);
+                    count++;
                     modifiedLinks.push_back(AZStd::move(newName));
                 }
                 else
                 {
-                    linkAndJointsName.insert(name->value());
+                    linkAndJointsName.insert(AZStd::make_pair(name->value(), 0));
                 }
             }
         }
@@ -106,31 +116,14 @@ namespace ROS2::Utils
         doc.parse<0>(const_cast<char*>(data.c_str()));
         xml_node<>* urdf = doc.first_node("robot");
         auto links = AddMissingInertiaToLinks(urdf);
+        modifiedElements.insert(modifiedElements.end(), AZStd::make_move_iterator(links.begin()), AZStd::make_move_iterator(links.end()));
 
-        if (links.size())
-        {
-            AZ_Warning("ROS2", false, "Added missing inertia to links: ");
-            for (auto& link : links)
-            {
-                modifiedElements.push_back(link);
-                AZ_Warning("ROS2", false, "  -> %s", link.c_str());
-            }
-        }
-
-        auto renames = ChangeDuplications(urdf);
-        if (renames.size())
-        {
-            AZ_Warning("ROS2", false, "Renamed duplicated links and joints: ");
-            for (auto& dup : renames)
-            {
-                modifiedElements.push_back(dup);
-                AZ_Warning("ROS2", false, "  -> %s", dup.c_str());
-            }
-        }
+        auto renames = RenameDuplicatedJoints(urdf);
+        modifiedElements.insert(
+            modifiedElements.end(), AZStd::make_move_iterator(renames.begin()), AZStd::make_move_iterator(renames.end()));
 
         std::string xmlDocString;
         AZ::rapidxml::print(std::back_inserter(xmlDocString), *urdf, 0);
-
         return { xmlDocString, modifiedElements };
     }
 } // namespace ROS2::Utils
--- a/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
@@ -1,0 +1,134 @@
+/*
+* Copyright (c) Contributors to the Open 3D Engine Project.
+* For complete copyright and license terms please see the LICENSE at the root of this distribution.
+*
+* SPDX-License-Identifier: Apache-2.0 OR MIT
+*
+*/
+
+#include "FixURDF.h"
+#include <AzCore/XML/rapidxml_print.h>
+#include <AzCore/std/containers/set.h>
+#include <iostream>
+namespace ROS2::Utils
+{
+    AZStd::vector<AZStd::string> AddMissingInertiaToLink(AZ::rapidxml::xml_node<>* urdf)
+    {
+        AZStd::vector<AZStd::string> modifiedLinks;
+        using namespace AZ::rapidxml;
+        AZStd::vector<xml_node<>*> links;
+        for (xml_node<>* link = urdf->first_node("link"); link; link = link->next_sibling("link"))
+        {
+            if (!link->first_node("inertial"))
+            {
+                links.push_back(link);
+            }
+        }
+
+        for (auto * link : links)
+        {
+
+            xml_node<>* inertial = urdf->document()->allocate_node(node_element, "inertial");
+
+            xml_node<>* mass = urdf->document()->allocate_node(node_element, "mass");
+            mass->append_attribute(urdf->document()->allocate_attribute("value", "1."));
+            inertial->append_node(mass);
+            xml_node<>* inertia = urdf->document()->allocate_node(node_element, "inertia");
+
+            inertia->append_attribute(urdf->document()->allocate_attribute("ixx", "1."));
+            inertia->append_attribute(urdf->document()->allocate_attribute("ixy", "0."));
+            inertia->append_attribute(urdf->document()->allocate_attribute("ixz", "0."));
+
+            inertia->append_attribute(urdf->document()->allocate_attribute("iyx", "0."));
+            inertia->append_attribute(urdf->document()->allocate_attribute("iyy", "1."));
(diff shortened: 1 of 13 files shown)


Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,33 @@
...
+       //! @param urdf - the in memory URDF to modify
+       //! @returns a list of links that were modified
+       AZStd::vector<AZStd::string> AddMissingInertiaToLink(AZ::rapidxml::xml_node<>* urdf);
+
+       //! Modifies names of links and joints to be unique, circle navigate SDF error 2
+       //! @param urdf - the in memory URDF to modify
+       //! @returns a list of links that were modified
+       AZStd::vector<AZStd::string> ChangeDuplications(AZ::rapidxml::xml_node<>* urdf);
+
+       std::string ModifyURDFInMemory(const std::string& data);
```
lumberyard-employee-dm, line 30: I think we should put this option to add missing `<inertial>` tags and rename duplicate links and joints into two new settings that is apart of the [SdfAssetBuilderSettings](https://github.com/o3de/o3de-extras/blob/development/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp#L60-L94).

We can source those settings from the [Settings Registry](https://github.com/o3de/o3de-extras/blob/development/Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp#L106-L114) and they would be available for the user to select via the File Selection Page in the Robot Importer.

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.cpp
```diff
@@ -47,6 +49,9 @@ namespace ROS2::UrdfParser
         }
 
         std::string xmlStr((std::istreambuf_iterator<char>(istream)), std::istreambuf_iterator<char>());
-        return Parse(xmlStr, parserConfig);
+        // modify in memory
+
+        std::string xmlStrChanged = (ROS2::Utils::ModifyURDFInMemory(xmlStr));
```
lumberyard-employee-dm, line 54: We can make the `Parse*` functions accept an `const SdfAssetBuilderSettings&` instance which can be used to select whether the URDF should be modified in memory before importing.

michalpelka:
Added option to enable and disable 
Screenshot from 2023-09-04 14-39-34
Added notification to the User:
Screenshot from 2023-09-04 14-43-13

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,135 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "FixURDF.h"
+#include <AzCore/XML/rapidxml_print.h>
+#include <AzCore/std/containers/set.h>
+#include <iostream>
```
adamdbrw, line 12:
```suggestion
#include <iostream>

```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,135 @@
...
+namespace ROS2::Utils
+{
+    AZStd::vector<AZStd::string> AddMissingInertiaToLink(AZ::rapidxml::xml_node<>* urdf)
```
adamdbrw, line 15: This affects multiple Links. Maybe AddMissingInertiaToLinks (plural)?

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+        for (xml_node<>* joint = urdf->first_node("joint"); joint; joint = joint->next_sibling("joint"))
+        {
+            auto* name = joint->first_attribute("name");
+            if (name)
+            {
+                if (linkAndJointsName.contains(name->value()))
+                {
+                    auto newName = AZStd::string(name->value()) + "_joint_dup";
```
adamdbrw, line 86: I am wondering whether it is priority to keep joint names unchanged (as opposed to keeping link names unchanged) since we actually use joint names for joint manipulation. However, we create entity names and derive namespaces based on links. What do you think?

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <AzCore/XML/rapidxml.h>
+#include <AzCore/std/containers/vector.h>
+#include <AzCore/std/string/string.h>
+
+namespace ROS2
+{
+    namespace Utils
+    {
+        //! Modifies in memory URDF to add missing inertia to links, circle navigate SDF error 19
```
adamdbrw, line 19:
```suggestion
        //! Modifies a parsed URDF in memory to add missing inertia to links, which prevents SDF error 19.
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        //! @param urdf - the in memory URDF to modify
```
adamdbrw, line 20:
```suggestion
        //! @param urdf URDF to modify.
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        //! @returns a list of links that were modified
```
adamdbrw, line 21:
```suggestion
        //! @returns a list of names of links that were modified.
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        AZStd::vector<AZStd::string> AddMissingInertiaToLink(AZ::rapidxml::xml_node<>* urdf);
+
+        //! Modifies names of links and joints to be unique, circle navigate SDF error 2
```
adamdbrw, line 24:
```suggestion
        //! Modifies names of links and joints to be unique, which means that a set of both contains no duplicates. This prevents SDF error 2.
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        //! @param urdf - the in memory URDF to modify
```
adamdbrw, line 25:
```suggestion
        //! @param urdf URDF to modify.
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        //! @returns a list of links that were modified
+        AZStd::vector<AZStd::string> ChangeDuplications(AZ::rapidxml::xml_node<>* urdf);
+
+        //! Modifies in memory URDF to add missing elements
+        //! @param urdf - the in memory URDF to modify
```
adamdbrw, line 30:
```suggestion
        //! @param urdf URDF to modify.
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        //! @returns a modified URDF and a list of XML element that were modified
+        AZStd::pair<std::string, AZStd::vector<AZStd::string>> ModifyURDFInMemory(const std::string& data);
+
```
adamdbrw, line 33:
```suggestion
```

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -147,21 +146,36 @@ namespace ROS2
...
                 AZ_Assert(false, "Unknown file extension : %s \n", m_urdfPath.c_str());
             }
             AZStd::string log;
             bool urdfParsedSuccess{ parsedUrdfOutcome };
+            bool urdfParsedWithWarnings{ parsedUrdfOutcome.m_modifiedURDFContent.size() > 0};
             if (urdfParsedSuccess)
             {
+                if (urdfParsedWithWarnings)
+                {
+                    report += "# " + tr("The URDF was parsed, but it was modified to be compatible") + "\n";
```
adamdbrw, line 162:
```suggestion
                    report += "# " + tr("The URDF was parsed, though results were modified to be compatible with SDFormat") + "\n";
```

adamdbrw: Not sure if this is right - but please add an explanation (compatible with what?)

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.cpp
```diff
@@ -116,13 +118,24 @@ namespace ROS2::UrdfParser
...
-                        std::string{ urdfFilePath.c_str(), urdfFilePath.Native().size() });
+            fileNotFoundResult.m_sdfErrors.emplace_back(
+                sdf::ErrorCode::FILE_READ,
+                std::string{ fileNotFoundMessage.c_str(), fileNotFoundMessage.size() },
+                std::string{ urdfFilePath.c_str(), urdfFilePath.Native().size() });
             return fileNotFoundResult;
         }
 
         std::string xmlStr((std::istreambuf_iterator<char>(istream)), std::istreambuf_iterator<char>());
+        // modify in memory
```
adamdbrw, line 129: Nit: move to the inside of curly brackets (like this):
```
{  // modify in memory
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
```diff
@@ -90,7 +92,13 @@ namespace ROS2
                         AZ::Edit::UIHandlers::Default,
                         &SdfAssetBuilderSettings::m_importReferencedMeshFiles,
                         "Import meshes",
-                        "Allows importing of referenced mesh content files such as .dae or .stl files when importing the URDF/SDF.");
+                        "Allows importing of referenced mesh content files such as .dae or .stl files when importing the URDF/SDF.")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &SdfAssetBuilderSettings::m_fixURDF,
+                        "Fix URDF to be compatible with libsdformat",
+                        "When set, fixes the URDF file before importing it. This is useful for fixing URDF files that have missing inertials."
```
adamdbrw, line 100:
```suggestion
                        "When set, fixes the URDF file before importing it. This is useful for fixing URDF files that have missing inertials or duplicate names within links and joints."
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.h
```diff
@@ -36,5 +36,7 @@ namespace ROS2
         bool m_urdfPreserveFixedJoints = true;
         // When true, .dae/.stl mesh files are imported into the project folder to allow the AP to process them
         bool m_importReferencedMeshFiles = true;
+        // When true URDF will be fixed to be compatible with SDF
```
adamdbrw, line 39:
```suggestion
        // When true URDF will be fixed to be compatible with SDFormat.
```

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.cpp
```diff
@@ -116,13 +118,24 @@ namespace ROS2::UrdfParser
...
+        if (Utils::IsFileUrdf(filePath) && settings.m_fixURDF)
+        {
+            const auto& [modifiedXmlStr, modifiedElements] = (ROS2::Utils::ModifyURDFInMemory(xmlStr));
+
+            auto result = Parse(modifiedXmlStr, parserConfig);
+            result.m_modifiedURDFTags = modifiedElements;
+            result.m_modifiedURDFContent = modifiedXmlStr;
+            return result;
```
lumberyard-employee-dm, line 137: Instead of using `const auto&` to bind to the each element of the pair as a const reference, a normal `auto` can be used, which can take advantage of return value optimization here. Because the types are now value types in this case, they can also be moved to the result structure to prevent an expensive copy of the parsed XML content.
```suggestion
            auto [modifiedXmlStr, modifiedElements] = (ROS2::Utils::ModifyURDFInMemory(xmlStr));

            auto result = Parse(modifiedXmlStr, parserConfig);
            result.m_modifiedURDFTags = AZStd::move(modifiedElements);
            result.m_modifiedURDFContent = AZStd::move(modifiedXmlStr);
            return result;
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        return modifiedLinks;
+    }
+
+    AZStd::vector<AZStd::string> ChangeDuplications(AZ::rapidxml::xml_node<>* urdf)
+    {
+        using namespace AZ::rapidxml;
+        AZStd::vector<AZStd::string> modifiedLinks;
+        AZStd::set<AZStd::string> linkAndJointsName;
```
lumberyard-employee-dm, line 71: Using an unordered_set should be quicker if there are a lot of duplicate joint names.
Though for small amount of duplicate joint names, this should be an issue.

Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
```diff
@@ -147,21 +146,36 @@ namespace ROS2
...
-                parsedUrdfOutcome = UrdfParser::ParseFromFile(m_urdfPath, parserConfig);
+                parsedUrdfOutcome = UrdfParser::ParseFromFile(m_urdfPath, parserConfig, sdfBuilderSettings);
             }
             else
             {
                 AZ_Assert(false, "Unknown file extension : %s \n", m_urdfPath.c_str());
             }
             AZStd::string log;
             bool urdfParsedSuccess{ parsedUrdfOutcome };
+            bool urdfParsedWithWarnings{ parsedUrdfOutcome.m_modifiedURDFContent.size() > 0};
```
lumberyard-employee-dm, line 157: Should we codify this into a function on the ParseResult struct in UrdfParser.h?
Perhaps a function such as
```
bool UrdfParsedWithModifiedContent() const;
```

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.h
```diff
@@ -10,16 +10,17 @@
 
 #include <AzCore/IO/Path/Path.h>
 #include <AzCore/std/utility/expected.h>
-#include <sdf/Root.hh>
+#include <SdfAssetBuilder/SdfAssetBuilderSettings.h>
```
lumberyard-employee-dm, line 13: Actually this include isn't needed in the UrdfParser.h.

The SdfAssetBuilderSettings can be forward declared in this header and then included in the cpp file.
That will help a little bit with compile times.
```suggestion
```

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.cpp
michalpelka: Good point!

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
michalpelka: Both are important. Links names affect frame names, other joint names.
Joint names are easier to change IHMO in frame component.

adamdbrw (approved):

adamdbrw:
image

Suggestions: 
1. Indicate what the modification was (for example "Modified names:", "Added Inertial").
2. Make it easy to copy / save the modified URDF.

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,136 @@
...
+                links.push_back(link);
+            }
+        }
+
+        for (auto* link : links)
+        {
+            xml_node<>* inertial = urdf->document()->allocate_node(node_element, "inertial");
+
+            xml_node<>* mass = urdf->document()->allocate_node(node_element, "mass");
+            mass->append_attribute(urdf->document()->allocate_attribute("value", "1."));
```
mbalfour-amzn, line 34: According to the URDF spec, this should actually default to a 0 mass if not specified: http://wiki.ros.org/urdf/XML/link . "<inertial> (optional: defaults to a zero mass and zero inertia if not specified)". I know the SDF converter requires this to be specified, but does the SDF converter also require this to be 1 and identity for inertia to avoid the error?

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,136 @@
...
+        using namespace AZ::rapidxml;
+        AZStd::vector<AZStd::string> modifiedLinks;
+        AZStd::unordered_set<AZStd::string> linkAndJointsName;
+        for (xml_node<>* link = urdf->first_node("link"); link; link = link->next_sibling("link"))
+        {
+            auto* name = link->first_attribute("name");
+            if (name)
+            {
+                linkAndJointsName.insert(name->value());
+            }
```
mbalfour-amzn, line 78: Does this need to check for duplicate link names too, or is that deliberately ignored here to let the SDF converter throw the error? If so, maybe add a comment here to explain that, something like "// Duplicate link names are invalid in a URDF file, so they won't be handled here. The SDF converter will detect the error case and throw an error."

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,136 @@
...
+        }
+        for (xml_node<>* joint = urdf->first_node("joint"); joint; joint = joint->next_sibling("joint"))
+        {
+            auto* name = joint->first_attribute("name");
+            if (name)
+            {
+                if (linkAndJointsName.contains(name->value()))
+                {
+                    auto newName = AZStd::string(name->value()) + "_joint_dup";
```
mbalfour-amzn, line 87: This looks like it will only work for the first name collision. If you have 3 joints all named "x", this looks like it would fail. Perhaps this should have an int that's initialized to 0 before the joint loop, and gets added to the name and incremented each time so that we'd have "x_joint_dup_0", "x_joint_dup_1", etc. You could be extra fancy and store a separate counter for each unique name by turning the unordered_set into an unordered_map, but but honestly it's probably enough to just have a global incrementing counter.

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        //! @param urdf URDF to modify.
+        //! @returns a list of names of links that were modified.
+        AZStd::vector<AZStd::string> AddMissingInertiaToLinks(AZ::rapidxml::xml_node<>* urdf);
+
+        //! Modifies names of links and joints to be unique, which means that a set of both contains no duplicates. This prevents SDF
+        //! error 2.
+        //! @param urdf URDF to modify.
+        //! @returns a list of links that were modified
+        AZStd::vector<AZStd::string> ChangeDuplications(AZ::rapidxml::xml_node<>* urdf);
+
```
mbalfour-amzn, line 29: Should these definitions get removed from the header, and then just move the comments into the cpp? I can't see why these need to be separately exposed publicly. Also, if you did that, you could move the rapidxml include into the cpp.

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.cpp
```diff
@@ -50,7 +51,7 @@ namespace ROS2::UrdfParser
     {
         return m_root;
     }
-    const sdf::Root& ParseResult::GetRoot() const &
+    const sdf::Root& ParseResult::GetRoot() const&
```
mbalfour-amzn, line 54: Is the & after the const a typo?

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.h
```diff
@@ -83,7 +96,9 @@ namespace ROS2
         //!        The relevant ParserConfig functions for URDF importing are
         //!        URDFPreserveFixedJoint() function to prevent merging of robot links bound by fixed joint
         //!        AddURIPath() function to provide a mapping of package:// and model:// references to the local filesystem
+        //! @paragraph settings structure that contains configuration options for the SDFAssetBuilder
```
mbalfour-amzn, line 99:
```suggestion
        //! @param settings structure that contains configuration options for the SDFAssetBuilder
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
michalpelka: I would not set it zero, since it will cause serious problems down the line in PhysX. I think unit mass and identity inertia is reasonable choice.

Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.cpp
michalpelka: It seems so, but it is not mine. It is a typo existing in codebase. I ran formatting here, what brought to this PR.

michalpelka: It is already exsting in codebase https://github.com/o3de/o3de-extras/blob/fc0e50e7a40e009110fb401f3add5e83f2cdc263/Gems/ROS2/Code/Source/RobotImporter/URDF/UrdfParser.h#L43C6-L43C47
I've here run clang-format.

adamdbrw: Regardless of source, let's remove the &

lumberyard-employee-dm: No the `const&` is not a typo.
It is a referenced qualifier which allows the GetRoot function to return a `const sdf::Root&` when the ParseResult is a const lvalue reference.

This differs when ParseResult is a non-const rvalue reference, such as a call to `UrdfParser::Parse("<xml", ...).GetRoot()` which would return an rvalue `sdf::Root&&` which would be moved automatically.

lumberyard-employee-dm (approved):

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
michalpelka: It was removed from the header. Thanks for the suggestion.

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
michalpelka: I went fancy road here :).

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.h
```diff
@@ -0,0 +1,35 @@
...
+        //! Modifies in memory URDF to add missing elements
+        //! @param urdf URDF to modify.
+        //! @returns a modified URDF and a list of XML element that were modified
+        AZStd::pair<std::string, AZStd::vector<AZStd::string>> ModifyURDFInMemory(const std::string& data);
```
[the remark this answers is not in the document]
michalpelka: Moved to CPP, expanded documentation slightly.

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
michalpelka: Done, change the name of the function to `RenameDuplicatedJoints`

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,145 @@
...
+        }
+        for (xml_node<>* joint = urdf->first_node("joint"); joint; joint = joint->next_sibling("joint"))
+        {
+            auto* name = joint->first_attribute("name");
+            if (name)
+            {
+                if (linkAndJointsName.contains(name->value()))
+                {
+                    unsigned int& count = linkAndJointsName[name->value()];
+                    auto newName = AZStd::string::format("%s_dup%d", name->value(), count);
```
mbalfour-amzn, line 97: This should be %u for an unsigned int.

```suggestion
                    auto newName = AZStd::string::format("%s_dup%u", name->value(), count);
```

mbalfour-amzn (approved):

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        auto renames = RenameDuplicatedJoints(urdf);
+        if (renames.size())
+        {
+            AZ_Warning("ROS2", false, "Renamed duplicated links and joints: ");
+            for (auto& dup : renames)
+            {
+                modifiedElements.push_back(dup);
+                AZ_Warning("ROS2", false, "  -> %s", dup.c_str());
```
lumberyard-employee-dm, line 137: I recommend not outputting any messages here, as the caller gets back all of the links that had missing inertia added and the renamed joints as part of the return value of this function.

The caller themselves can then log the messages, if they please.
This helps prevent console spam and allows the caller to choose to output or suppress the warning messages more easily for their use cases.

PR #482, opened 2023-08-29
Refactor ROS 2 sensor components events management

PawelLiberadzki:
This is a proposition of universal sensor events management, that solves issue #295.

I introduce event source interface, and two sample implementations of it - TickBasedSource and PhysicsBasedSource. They are used through event source adapter in a new sensor component base class ROS2SensorComponentBase (previously ROS2SensorComponent).


Gems/ROS2/Code/Include/ROS2/Sensor/Events/TickBasedSource.h
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include <AzCore/Component/TickBus.h>
+#include <AzCore/EBus/Event.h>
+#include <ROS2/Sensor/Events/SensorEventSource.h>
+
+namespace ROS2
+{
+    //! Class implementing system TickBus (draw calls) as sensor event source. Source event (ROS2::SensorEventSource) is signalled based on
+    //! system ticks. This event source can de enabled / disabled using m_publishingEnabled member of SensorConfiguration
```
adamdbrw, line 18: There are cases where the source could be active even though it is not publishing, such as simply acquiring data to visualize it or process simulation-side only.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/TickBasedSource.h
```diff
@@ -0,0 +1,41 @@
...
+        void Configure(const SensorConfiguration& sensorConfiguration) override;
+
+    protected:
+        // Override of AZ::TickBus::Handler.
+        void OnTick(float deltaTime, AZ::ScriptTimePoint time) override;
+
+    private:
+        bool m_sourceEnabled{ true }; ///< Toggle for event source enabled/disabled state.
+    };
+} // namespace ROS2
```
adamdbrw, line 41: nit: ensure newlines in files

Gems/ROS2/Code/Include/ROS2/Sensor/Events/SensorEventSource.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#pragma once
+
+namespace ROS2
+{
+    struct SensorConfiguration;
+
+    //! Base class template for sensor event sources (based e.g. on TickBus or engine physics callback). Developer that wants to implement
+    //! new event source, should derive it from this class and signal m_sensorEvent based on chosen approach. Derived classes should rather
```
adamdbrw, line 16:
```suggestion
    //! new event source should derive from this class and signal m_sensorEvent based on chosen approach. Derived classes should rather
```

Gems/ROS2/Code/Include/ROS2/Sensor/Events/SensorEventSource.h
```diff
@@ -0,0 +1,65 @@
...
+    //! not auto connect any handler to m_sensorEvent, unless it is done to manage their internal logic. This event is for the final user
+    //! that can connect prepared event handler to it using ConnectToSourceEvent method.
```
adamdbrw, line 18: This class documentation is a bit unclear to me. Could you rephrase it?
It could be nice to clarify distinction and refer to EventSourceAdapter since typical use of sensors involves frequency.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/SensorEventSource.h
```diff
@@ -0,0 +1,65 @@
...
+    class SensorEventSource
+    {
+    public:
+        using SourceBaseType = SensorEventSource<EventT, EventHandlerT, EventArgs...>;
+        using SourceEventType = EventT<EventArgs...>;
+        using SourceEventHandlerType = EventHandlerT<EventArgs...>;
+
+        virtual ~SensorEventSource() = default;
+
+        //! Sets up event source - ee event source description for more details. After call to this method event source is supposed to start
```
adamdbrw, line 31:
```suggestion
        //! Sets up event source - see event source description for more details. After call to this method event source is supposed to start
```

Gems/ROS2/Code/Include/ROS2/Sensor/Events/SensorEventSource.h
```diff
@@ -0,0 +1,65 @@
...
+        //! stop signalling source event.
+        virtual void Deactivate()
+        {
+        }
+
+        //! Configures event source based on sensor settings. Details depend on specific event source implementation.
+        //! @param sensorConfiguration Implemented sensor configuration.
+        //! @note In most cases it will be safer to call this before Activate, however at the end it depends on specific event source
+        //! implementations.
+        virtual void Configure([[maybe_unused]] const SensorConfiguration& sensorConfiguration)
```
adamdbrw, line 47: Not entirely sure about purpose of this member function. It seems fields other than frequency are not related.

Gems/ROS2/Code/Source/Lidar/ROS2Lidar2DSensorComponent.cpp
```diff
@@ -173,12 +172,28 @@ namespace ROS2
...
+        m_eventSourceAdapter.ConnectToSourceEvent(m_sourceEventHandler);
+
+        m_adaptedEventHandler = decltype(m_adaptedEventHandler)([this](auto... args)
+        {
+            this->FrequencyTick();
+        });
+        m_eventSourceAdapter.ConnectToAdaptedEvent(m_adaptedEventHandler);
+
+        m_eventSourceAdapter.Configure(m_sensorConfiguration);
+        m_eventSourceAdapter.Activate();
```
adamdbrw, line 188: There is nothing lidar-specific to this added code block, and it looks like it would be repeated in each sensor implementation (with Tick source). Is it possible to do it in ROS2SensorComponentBase or generalize otherwise?

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
```diff
@@ -0,0 +1,190 @@
...
+                serializeContext->Class<EventSourceAdapter<EventSourceT>>()
+                    ->Version(1)
+                    ->Field("Adapted frequency", &EventSourceAdapter<EventSourceT>::m_adaptedFrequency)
+                    ->Field("Event source configuration", &EventSourceAdapter<EventSourceT>::m_eventSource);
+
+                if (auto editContext = serializeContext->GetEditContext())
+                {
+                    editContext
+                        ->Class<EventSourceAdapter<EventSourceT>>(
+                            "Event Source Adapter", "Adapts sensor event source to specified working frequency")
```
adamdbrw, line 85: To discuss: I think that users of sensor components do not need to know about their inner workings (that there are event source adapters and such). This is knowledge for sensor developers.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
```diff
@@ -0,0 +1,190 @@
...
+        //! investigate specific event source implementation with such case in mind.
+        void Deactivate()
+        {
+            m_eventSource.Deactivate();
+            m_adaptedEventHandler.Disconnect();
+        }
+
+        //! Parses sensor configuration into event source configuration.
+        //! @param sensorConfiguration Configuration of sensor using this event source adapter.
+        void Configure(const SensorConfiguration& sensorConfiguration)
```
adamdbrw, line 134: Not sure if anything else than frequency from this configuration would ever be useful at this level.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
```diff
@@ -0,0 +1,190 @@
...
+        //! @param adaptedEventHandler Event handler for adapted event (frequency set through event source adapter configuration).
+        void ConnectToAdaptedEvent(typename EventSourceT::SourceEventHandlerType& adaptedEventHandler)
+        {
+            adaptedEventHandler.Connect(m_sensorAdaptedEvent);
+        }
+
+    private:
+        //! Uses tick counter, expected loop time and frequency set for adapter to support managing calls from event source.
+        //! @param expectedLoopTime Expected simulation loop time.
+        //! @return Whether it is time to signal adapted event.
```
adamdbrw, line 162: It is feasible to want a different mechanism for approximating frequencies, but this does not need to be in this PR.

Gems/ROS2/Code/Source/Sensor/Events/PhysicsBasedSource.cpp
```diff
@@ -0,0 +1,59 @@
...
+        AzPhysics::SceneHandle sceneHandle = sceneInterface->GetSceneHandle(AzPhysics::DefaultPhysicsSceneName);
+        sceneInterface->RegisterSceneSimulationFinishHandler(sceneHandle, m_onSceneSimulationEvent);
+    }
+
+    void PhysicsBasedSource::Deactivate()
+    {
+        m_onSceneSimulationEvent.Disconnect();
+    }
+
+    void PhysicsBasedSource::Configure(const SensorConfiguration& sensorConfiguration)
```
adamdbrw, line 55: I think that control logic of publishing should not be a part of the source.
m_publishingEnables whether the data is published, not whether it is acquired in the first place. Removing Configure seems reasonable in this EventSource hierarchy.

I believe that whether the source produces events should be controlled only by Active state of Component which uses it.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
PawelLiberadzki: We agreed to remove from editor context everything from event source adapter down. After few thoughts, I decided to leave event source adapter edit context, and stop exposing it in ROS2SensorComponentBase instead. This way, it will not be visible in sensors implementing ROS2SensorComponentBase, but someone using EventSourceAdapter for different purpose will be able to expose its parameters to user.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
```diff
@@ -89,57 +86,55 @@ namespace ROS2
...
         }
 
-        //! Parses sensor configuration into event source configuration.
-        //! @param sensorConfiguration Configuration of sensor using this event source adapter.
-        void Configure(const SensorConfiguration& sensorConfiguration)
+        //! Sets adapter working frequency. By design, adapter will not work correctly, if this frequency will be greater than used event
+        //! source frequency - e.g. adapter will be requested to work in 60Hz, when using event source working in 30Hz. In general, adapted
+        //! frequency should be equal or lower than event source frequency. Optimal (highest precision in timing events) working conditions
+        //! take place when event source frequency is an integer multiple of adapted frequency.
+        //! @param adaptedFrequency Adapter working frequency. When set to zero or less adapter will be assumed to work in 1Hz.
```
adamdbrw, line 129: This an unnecessarily limiting assumption. Once per day is a sensible minimum (about 0.000012). This needs to be applied in SensorConfiguration m_frequency field MinValue Attribute. Please raise a new issue to limit frequency input (also consider limiting max to what the global setting for physics currently allows).

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
```diff
@@ -77,25 +77,54 @@ namespace ROS2
...
+
+        m_linearAccelerationCovariance = ToDiagonalCovarianceMatrix(m_imuConfiguration.m_linearAccelerationVariance);
+        m_angularVelocityCovariance = ToDiagonalCovarianceMatrix(m_imuConfiguration.m_angularVelocityVariance);
+        m_orientationCovariance = ToDiagonalCovarianceMatrix(m_imuConfiguration.m_orientationVariance);
+
+        StartSensor(
+            m_sensorConfiguration.m_frequency,
+            [this](float imuDeltaTime, AzPhysics::SceneHandle sceneHandle, float physicsDeltaTime)
+            {
+                if (!m_sensorConfiguration.m_publishingEnabled)
```
adamdbrw, line 97: Perhaps moving condition check to the OnImuEvent makes more sense (as opposed to leaving it here for a tiny bit of optimization). But one can argue for either.

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
```diff
@@ -77,25 +77,54 @@ namespace ROS2
...
 
-    void ROS2ImuSensorComponent::OnPhysicsSimulationFinished(AzPhysics::SceneHandle sceneHandle, float deltaTime)
+    void ROS2ImuSensorComponent::OnPhysicsEvent(AzPhysics::SceneHandle sceneHandle)
     {
+        if (m_bodyHandle == AzPhysics::InvalidSimulatedBodyHandle)
+        {
+            AzPhysics::RigidBody* rigidBody = nullptr;
+            AZ::EntityId entityId = GetEntityId();
+            Physics::RigidBodyRequestBus::EventResult(rigidBody, entityId, &Physics::RigidBodyRequests::GetRigidBody);
+            AZ_Assert(rigidBody, "Entity %s does not have rigid body.", entityId.ToString().c_str());
```
adamdbrw, line 123:
```suggestion
            AZ_Assert(rigidBody, "Entity %s does not have a rigid body.", entityId.ToString().c_str());
```

adamdbrw: In release, there will be no assert and it will crash in next line. May we handle null rigid body by logging an error and disconnecting the source instead? We might not want to necessarily crash the entire app because one sensor has an issue.

Gems/ROS2/Code/Source/Lidar/ROS2Lidar2DSensorComponent.cpp
```diff
@@ -172,28 +174,25 @@ namespace ROS2
...
+            {
+                if (!m_sensorConfiguration.m_publishingEnabled)
+                {
+                    return;
+                }
+                FrequencyTick();
+            },
+            [this]([[maybe_unused]] auto&&... args)
+            {
+                Visualize();
```
adamdbrw, line 189: Lets discuss: 
"We should only visualise data with given frequency, otherwise the visualisation data will be misleading (even if looking smoother)"

adamdbrw (approved):
Good work!

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
PawelLiberadzki: Agree. I changed assert to error + return / stop.

Gems/ROS2/Code/Source/Lidar/ROS2Lidar2DSensorComponent.cpp
PawelLiberadzki: This will probably lead to points being visualized once every few frames - they will be not visible most of the time. If this is acceptable, I can change the code that way.

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
PawelLiberadzki: The reason behind placing this if-statement here was to expose it to user / make it more visible. I would stick to that, to be honest.

adamdbrw (approved):

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
```diff
@@ -0,0 +1,192 @@
...
+
+#include <ROS2/ROS2Bus.h>
+#include <ROS2/Sensor/Events/SensorEventSource.h>
+#include <ROS2/Sensor/SensorConfiguration.h>
+
+namespace ROS2
+{
+    namespace Internal
+    {
+        template<
```
michalpelka, line 19: maybe turn off clang format here.

michalpelka (commented):
Please make sure that the contribution compiles with non-unity build.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/PhysicsBasedSource.h
```diff
@@ -0,0 +1,35 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#pragma once
+
+#include <AzCore/EBus/OrderedEvent.h>
+#include <AzFramework/Physics/Common/PhysicsTypes.h>
```
michalpelka, line 12: This was preventing non-unity build for me:
```suggestion
#include <AzFramework/Physics/Common/PhysicsTypes.h>
#include <AzFramework/Physics/PhysicsScene.h>
```

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
```diff
@@ -77,25 +77,61 @@ namespace ROS2
...
+                StopSensor();
+                return;
+            }
+
+            m_bodyHandle = rigidBody->m_bodyHandle;
+        }
+
         auto* sceneInterface = AZ::Interface<AzPhysics::SceneInterface>::Get();
-        const auto gravity = sceneInterface->GetGravity(sceneHandle);
         auto* body = sceneInterface->GetSimulatedBodyFromHandle(sceneHandle, m_bodyHandle);
```
michalpelka, line 135: I would suggest multiple assertions here - the code can mistakenly be executed in a different context by mistake (e.g. in Asset Processor) and a good assertion can be helpful.

michalpelka: I belive gravity is not needed.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
PawelLiberadzki: We agreed, through discussion, that this formatting is acceptable here.

Gems/ROS2/Code/Source/Imu/ROS2ImuSensorComponent.cpp
PawelLiberadzki: I added an assert on body pointer - this is a good catch. I left sceneInterface as it is - do you maybe think that it is necessary to assert it, when it is a result from:
   AZ::Interface<AzPhysics::SceneInterface>::Get()
engine call? Here I assume, that it won't rather fail, and if it will, then we have some critical issue. Is such reasoning not correct?

michalpelka (approved):

Issue #493, opened 2023-09-05
Issues with building GameLauncher (alone) based on Ros2ProjectTemplate


adamdbrw:
```
CMake Error at /home/piotr/ROBOTEC/O3DE/o3de/cmake/LYWrappers.cmake:210 (add_library):
  Target "MyProject.Editor.Static" links to target "AZ::AzToolsFramework" but
  the target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
Call Stack (most recent call first):
  Gem/CMakeLists.txt:83 (ly_add_target)
```

The issue is likely due to Editor.Static target not being within:
```
if(PAL_TRAIT_BUILD_HOST_TOOLS)
endif()
```

PR #494, opened 2023-09-05
Ros2ProjectTemplate fixes to monolithic build

lewandowskikacper:
This applies suggestion from https://github.com/o3de/o3de-extras/issues/493.

~~There are linker issues still present after those changes.~~

The linker issues were caused by dependency on AWSCore Gem, that is now removed.

### DIFF
--- a/Gems/ROS2/Code/Source/Manipulation/JointsManipulationComponent.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/JointsManipulationComponent.cpp
@@ -125,7 +125,7 @@ namespace ROS2
                 auto* prismaticComponent =
                     azrtti_cast<PhysX::JointComponent*>(Utils::GetGameOrEditorComponent<PhysX::PrismaticJointComponent>(entity));
                 auto* articulationComponent = Utils::GetGameOrEditorComponent<PhysX::ArticulationLinkComponent>(entity);
-                bool classicJoint = hingeComponent || prismaticComponent;
+                [[maybe_unused]] bool classicJoint = hingeComponent || prismaticComponent;
                 AZ_Warning(
                     "JointsManipulationComponent",
                     (classicJoint && supportsClassicJoints) || !classicJoint,
--- a/Gems/ROS2/Code/Source/Gripper/VacuumGripperComponent.cpp
+++ b/Gems/ROS2/Code/Source/Gripper/VacuumGripperComponent.cpp
@@ -104,8 +104,7 @@ namespace ROS2
 
     void VacuumGripperComponent::OnTick(float delta, AZ::ScriptTimePoint timePoint)
     {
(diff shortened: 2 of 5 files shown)


Gems/ROS2/Code/Source/FactorySimulation/ConveyorBeltComponent.cpp
```diff
@@ -61,7 +61,7 @@ namespace ROS2
 
     void ConveyorBeltComponent::Activate()
     {
-        AzPhysics::SystemInterface* physicsSystem = AZ::Interface<AzPhysics::SystemInterface>::Get();
+        [[maybe_unused]] AzPhysics::SystemInterface* physicsSystem = AZ::Interface<AzPhysics::SystemInterface>::Get();
```
adamdbrw, line 64: in each case, we don't actually need to run it in release builds. Please move to asserts, as in:
` AZ_Assert(AZ::Interface<AzPhysics::SystemInterface>::Get(), "No physics system");`
Apply to other cases.

Gems/ROS2/Code/Source/Manipulation/JointsManipulationComponent.cpp
```diff
@@ -125,7 +125,7 @@ namespace ROS2
                 auto* prismaticComponent =
                     azrtti_cast<PhysX::JointComponent*>(Utils::GetGameOrEditorComponent<PhysX::PrismaticJointComponent>(entity));
                 auto* articulationComponent = Utils::GetGameOrEditorComponent<PhysX::ArticulationLinkComponent>(entity);
-                bool classicJoint = hingeComponent || prismaticComponent;
+                [[maybe_unused]] bool classicJoint = hingeComponent || prismaticComponent;
```
adamdbrw, line 128: this one can stay as is - good suggestion

adamdbrw (commented):
Does this PR introduce new issues, or does it simply not resolve all issues present?

lewandowskikacper:
> Does this PR introduce new issues, or does it simply not resolve all issues present?

It solves issues with this Template now.

lumberyard-employee-dm (approved):

lewandowskikacper:
I also checked the [Ros2FleetRobotTemplate](https://github.com/o3de/o3de-extras/tree/development/Templates/Ros2FleetRobotTemplate), [Ros2RoboticManipulationTemplate](https://github.com/o3de/o3de-extras/tree/development/Templates/Ros2RoboticManipulationTemplate). They both compile in profile config.

In release config, compilation fails due to "unused but set variable" caused by `result` in https://github.com/o3de/o3de/blob/60a1e5a368282a13654f45d8c1aa7be2192dfcaa/Gems/Atom/RPI/Code/Source/RPI.Public/RPISystem.cpp#L299C1-L306C21. If that is fixed, it succeeds.

michalpelka (approved):

Issue #477, opened 2023-08-28
URDF Importer - support meshes with textures with "Import meshes during URDF format".


michalpelka:
Assume that URDF has multiple DAE meshes with materials and textures like ![this URDF](https://github.com/ANYbotics/anymal_c_simple_description/tree/master/meshes).

Currently URDF/SDF/XACRO importer ommit jpg/png and other files during import (it copies only meshes like FBX, DAE, OBJ or STL). 
To import such model one has:
 - copy all assets manually
 - untick this option "Import meshes during URDF format"
With that, URDF importer scans all assets and search for copy against that in ROS2 package. 
Here is result for such import and for import with option "Import meshes during URDF format" ticked.
![image](https://github.com/o3de/o3de-extras/assets/3209244/d9685a3e-ef40-4b7b-bd40-6379d550247f)
** Note ** this issue is also present for other formats such GLTF! In GLTF in companion file geometry lives.

Related issue:
**https://github.com/o3de/o3de-extras/issues/358**

PR #497, opened 2023-09-05
Robot Importer: copy texture files

jhanca-robotecai:
This is a PoC code for copying textures (indicated by #477). I am open to discussions.

Please note, that this PR does not solve the problem of missing mesh (e.g. `gltf` file might store mesh in a separate binary file).

Test mesh:
[texture.zip](https://github.com/o3de/o3de-extras/files/12524562/texture.zip)
Outcome:
image


jhanca-robotecai:
Please note, that I also fixed some typos and removed unused code...

jhanca-robotecai:
TODO: there might be some problems with path resolving if the textures are located in different folder - this will be fixed later if the approach is approved.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -541,13 +560,46 @@ namespace ROS2::Utils
...
+
+    AZStd::unordered_set<AZStd::string> GetMeshAssets(const AZStd::shared_ptr<AZ::SceneAPI::Containers::Scene>& scene)
     {
-        auto assetInfoFilePath = AZ::IO::Path{ sourceAssetPath };
-        assetInfoFilePath.Native() += ".assetinfo";
-        return CreateSceneManifest(sourceAssetPath, sourceAssetPath.Native() + ".assetinfo", collider, visual);
+        AZStd::unordered_set<AZStd::string> assetsFilenames;
+
+        // Look for material files
+        AZStd::initializer_list<AZ::SceneAPI::DataTypes::IMaterialData::TextureMapType> allTextureTypes = {
```
michalpelka, line 580: consider const or even better constexpr.

jhanca-robotecai:
> TODO: there might be some problems with path resolving if the textures are located in different folder - this will be fixed later if the approach is approved.

Fixed. The code was tested for the following configurations:

```
texture
├── face.dae
├── face.jpg
└── face.urdf
texture1
├── face.urdf
└── mesh
    ├── face.dae
    └── textures
        └── face.jpg
texture2
├── face.urdf
├── mesh
│   └── face.dae
└── textures
    └── face.jpg
```

and resulted in the following structure in `Assets` folder:
```
├── Importer
│   ├── texture1.prefab
│   ├── texture2.prefab
│   └── texture.prefab
└── UrdfImporter
    ├── 1266655632_face
    │   ├── face.dae
    │   └── face.dae.assetinfo
    ├── 1535885513_face
    │   ├── face.dae
    │   ├── face.dae.assetinfo
    │   └── face.jpg
    ├── 3533555601_face
    │   ├── face.dae
    │   ├── face.dae.assetinfo
    │   └── textures
    │       └── face.jpg
    └── textures
        └── face.jpg

```

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
jhanca-robotecai: I am not sure if `constexpr` is possible in this configuration. Added `static const`.

jhanca-robotecai:
The code does not solve all possible problems with textures (textures might come from another package), but it solves multiple cases existing in robots (e.g. [ANYmal C](https://github.com/ANYbotics/anymal_c_simple_description) ) and it is a step forward comparing to the previous codebase.

michalpelka (approved):

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -541,13 +569,46 @@ namespace ROS2::Utils
         scene->GetManifest().SaveToFile(assetInfoFile.Native());
         AZ_Printf("CreateSceneManifest", "Saving scene manifest to %s\n", assetInfoFile.c_str());
 
+        assetsFilepaths = GetMeshAssets(scene);
+        AZ_Printf("CreateSceneManifest", "Found %d assets for %s scene\n", assetsFilepaths.size(), assetInfoFile.c_str());
```
mbalfour-amzn, line 573: It feels to me like GetMeshAssets() should get split out into a separate call from CreateSceneManifest(), since it's not really a part of creating the manifest.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -386,7 +386,31 @@ namespace ROS2::Utils
...
+                            const AZ::IO::Path assetFullPathSrc(AZ::IO::Path(resolvedPath.ParentPath()) / assetLocalPath);
+                            const AZ::IO::Path assetFullPathDst(importDirectoryDst / assetLocalPath);
+
+                            const auto outcomeMkdir = fileIO->CreatePath(AZ::IO::Path(assetFullPathDst.ParentPath()).c_str());
+                            if (!outcomeMkdir)
+                            {
+                                break;
+                            }
+
+                            const auto outcomeCopy = fileIO->Copy(assetFullPathSrc.c_str(), assetFullPathDst.c_str());
```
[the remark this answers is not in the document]
jhanca-robotecai: The importer copies all meshes from the source directory into temp folder in order to create `assetinfo` files. Next, it moves meshes from temp into the destination folder. However, the materials are not touched in the process, which means this code copies them from the source directory into the destination directory, without temp folder in the way.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -541,13 +569,46 @@ namespace ROS2::Utils
...
+        {
+            for (auto textureType : allTextureTypes)
+            {
+                if (const AZ::IO::Path filePath(material.GetTexture(textureType)); !filePath.empty())
+                {
+                    assetsFilepaths.emplace(filePath);
+                }
+            }
+        }
+        return assetsFilepaths;
```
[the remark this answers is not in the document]
jhanca-robotecai: There is the assumption that materials are linked with meshes. So if a material is in the same package as mesh, it should work. I believe this assumption is valid, as the code handles only cases when materials are called from within a mesh.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
jhanca-robotecai: I am aware it is not the most convenient way to put it. I did it as a first PoC because `CreateSceneManifest()`  is where the scene is available. If you think the rest of the code is valid, I will move it. Thanks!

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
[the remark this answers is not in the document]
mbalfour-amzn: ohhh, ok. I didn't realize these were skipping the temp directory, thanks!

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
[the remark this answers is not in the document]
mbalfour-amzn: The assumption isn't _entirely_ correct. SDF files can have a <material> tag that references files directly, without being referenced by the mesh itself. However, this code looks like it will work well for the case that the mesh _does_ reference the files directly.

The point that I was trying to bring up though is that there are at least some material files (like the ambulance one) that aren't actually valid the way they're written, since they're dependent on something being able to resolve the "model://" prefixes of the file names. I'm not sure how these material files are authored or tested. But the problem we have with this with O3DE is that AssImpMaterialImporter::ResolveTexturePath() has its own heuristics on how to resolve a path, and those heuristics currently can't handle resolving "model://" or even for that matter finding ambulance.png because it lives in a sibling directory, not the same directory or a parent directory. I think we'll need to modify ResolveTexturePath() to allow for similar heuristics as the ones we have for the SDF files to be able to correctly resolve these.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
mbalfour-amzn: Yep, I think the rest of the code is valid. It seems like a slightly painful way to get all of the texture references, but it looks like it's probably the best possible way.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -353,11 +353,35 @@ namespace ROS2::Utils
...
+                            {
+                                break;
+                            }
+
+                            const auto outcomeCopy = fileIO->Copy(assetFullPathSrc.c_str(), assetFullPathDst.c_str());
+                            if (outcomeCopy)
+                            {
+                                copiedFiles[assetFullPathSrc.String()] = assetFullPathDst.String();
+                            }
+                        }
```
mbalfour-amzn, line 383: I have a feeling this logic is going to need some iteration over time, but I'm OK with it going in and improving it as we hit issues. Some of the things I'm thinking about:
* If the mesh refers to textures that can't be found, this doesn't look like it's doing much in the way of error-detection and reporting yet.
* @lumberyard-employee-dm We might hit an odd problem with some upcoming work, where we add URI prefix resolution for material textures to the asset import. For the URI prefix resolution to work, it needs to be able to find the texture. If we try to do the resolution before copying the textures, the textures won't exist anywhere that we can find them yet. If we try to copy the textures first, we won't have done the prefix resolution yet to know how to copy them. It's possible we'll need to do it in both places... :(

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -535,10 +560,48 @@ namespace ROS2::Utils
...
 
-    bool CreateSceneManifest(const AZ::IO::Path& sourceAssetPath, bool collider, bool visual)
+    bool CreateSceneManifest(const AZ::IO::Path& sourceAssetPath, const bool collider, const bool visual)
     {
-        auto assetInfoFilePath = AZ::IO::Path{ sourceAssetPath };
-        assetInfoFilePath.Native() += ".assetinfo";
         return CreateSceneManifest(sourceAssetPath, sourceAssetPath.Native() + ".assetinfo", collider, visual);
     }
+
+    AZStd::unordered_set<AZ::IO::Path> GetMeshAssets(const AZ::IO::Path& sourceAssetPath)
```
mbalfour-amzn, line 568: Suggestion: Maybe `GetMeshTextureAssets` instead of `GetMeshAssets`. To me, `GetMeshAssets` sounds like it's returning meshes, not textures.

Also, maybe `sourceMeshAssetPath` instead of `sourceAssetPath` to make it clear that it's expecting the mesh file as input.

mbalfour-amzn (approved):

PR #505, opened 2023-09-07
Add configurable SDF/URDF path resolution settings, including prefix remappings

mbalfour-amzn:
In SDF/URDF models, the various asset paths and URIs are likely to have a combination of absolute paths, relative paths, and paths that start with "model://", "package://", "http://", etc.

This PR exposes a set of Path Resolution heuristics that can be used to resolve these paths into the desired file locations on disk:
* "Use AMENT_PREFIX_PATH" - uses the environment variable to try and resolve relative paths. (This heuristic already existed, it's just formalized as a setting now)
* "Search Parent Paths" - tries to resolve relative paths by moving up the directory tree from the base sdf/urdf file. (This heuristic already existed, it's also formalized as a setting now)
* "Prefix Replacements" - tries to replace a prefix on each path with other provided partial paths. This is a new heuristic, and it enables more specialized mappings like turning an "http://" path into a path that exists locally without requiring a modification to the input file.

The Prefix Replacements are also passed to the libsdformat library so that they can be used to try and locate any sdf files that have been included via an <include> tag. Without this, the <include> tag is unlikely to work correctly.

These settings have global defaults set in the sdfassetbuilder_settings.setreg file.

By default, the Path Resolution settings are collapsed in the Robot Importer window so that the user isn't overwhelmed by them and can hopefully generally rely on the defaults:
image

If the settings are expanded, here's what they look like:
image

Other changes:
* I also changed the File Selection dialog to no longer set a default directory - when using the dialog multiple times in a row, this lets the file selector default to the last directory used, instead of always going back to the project root directory.


Gems/ROS2/Code/Source/RobotImporter/Pages/FileSelectionPage.cpp
```diff
@@ -22,7 +22,6 @@ namespace ROS2
         , m_sdfAssetBuilderSettings(AZStd::make_unique<SdfAssetBuilderSettings>())
     {
         m_fileDialog = new QFileDialog(this);
-        m_fileDialog->setDirectory(QString::fromUtf8(AZ::Utils::GetProjectPath().data()));
```
lumberyard-employee-dm, line 25: Actually instead of removing setting the directory, I think it would be better to load the most recently used directory into the File Dialog.

We can store the directory being shown into the Editor Preferences section(`/O3DE/Preferences`) of the Settings Registry using the [QFileDialog::directory](https://doc.qt.io/qt-6/qfiledialog.html#directory) function to query it.

We can do this based on the `completeChanged` signal that gets emitted below when a file is selected or a file path is directly typed into the text editor

Gems/ROS2/Registry/sdfassetbuilder_settings.setreg
```diff
@@ -13,7 +13,12 @@
                     "xacro"
                 ],
                 "UseArticulations": true,
-                "URDFPreserveFixedJoint": true
+                "URDFPreserveFixedJoint": true,
+                "PathPrefixMap": {
+                    "model://": ["./", "../" ],
+                    "package://": ["./", "../" ],
+                    "file://":  ["/"]
```
lumberyard-employee-dm, line 20: For the `file://` the value of the path should be an empty string.

This is because the `file://` scheme already ascribes a `/` as meaning an [absolute path](https://en.wikipedia.org/wiki/File_URI_scheme)

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
```diff
@@ -850,12 +863,13 @@ namespace UnitTest
...
+        constexpr AZ::IO::PathView expectedResult = "/home/foo/ros_ws/install/foo_robot/meshes/bar.dae";
+        auto result = ROS2::Utils::ResolveAssetPath(
             dae,
-            urdf, "",
-            [](const AZ::IO::PathView&) -> bool
+            urdf, "", GetTestSettings(),
+            [expectedResult](const AZ::IO::PathView& p) -> bool
             {
-                return false;
+                return (p == expectedResult);
```
lumberyard-employee-dm, line 872: Why parenthesis around the condition in this case, but not the one that occurs later?
```suggestion
                return p == expectedResult;
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.h
```diff
@@ -8,11 +8,32 @@
...
 {
+    struct SdfAssetPathResolverSettings
+    {
+        public:
+            AZ_RTTI(SdfAssetPathResolverSettings, "{51EDDB99-FE82-4783-9C91-7DF403AD4EFA}");
+
+            SdfAssetPathResolverSettings() = default;
+            virtual ~SdfAssetPathResolverSettings() = default;
+
+            static void Reflect(AZ::ReflectContext* context);
```
lumberyard-employee-dm, line 25: This is over indented by 1 one level.
```suggestion
    public:
        AZ_RTTI(SdfAssetPathResolverSettings, "{51EDDB99-FE82-4783-9C91-7DF403AD4EFA}");

        SdfAssetPathResolverSettings() = default;
        virtual ~SdfAssetPathResolverSettings() = default;

        static void Reflect(AZ::ReflectContext* context);
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.h
```diff
@@ -8,11 +8,32 @@
...
+
+        using UriPrefixMap = AZStd::unordered_map<AZStd::string, AZStd::vector<AZStd::string>>;
```
lumberyard-employee-dm, line 27: I recommend changing the value type to be a vector of `AZ::IO::Path`, since that is what is being mapped.
```suggestion
        using UriPrefixMap = AZStd::unordered_map<AZStd::string, AZStd::vector<AZ::IO::Path>>;
```

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
```diff
@@ -142,6 +189,9 @@ namespace ROS2
             };
         AZ::SettingsRegistryVisitorUtils::VisitArray(*settingsRegistry, VisitFileTypeExtensions, SdfAssetBuilderSupportedFileExtensionsRegistryKey);
 
+        // Get the URI prefix map
+        settingsRegistry->GetObject(m_resolverSettings.m_uriPrefixMap, SdfAssetBuilderUriPrefixMapRegistryKey);
```
lumberyard-employee-dm, line 193: As the `SdfAssetPathResolverSettings` settings are reflected to the SerializeContext, this code can directly read into the `m_resolverSettings` structure using the SettingsRegistry, as the Settings Registry will use the JSON Serializer when GetObject is invoked.
```suggestion
        // Get the URI prefix map
        settingsRegistry->GetObject(m_resolverSettings, SdfAssetBuilderResolverRegistryRootKey);
```

Gems/ROS2/Code/Source/RobotImporter/Pages/FileSelectionPage.cpp
mbalfour-amzn: Great suggestion, the only tweak is that I'd be inclined to follow a similar implementation to the AssetImporterManager and store the directory into QSettings instead of the editorpreferences.setreg. Looking at the other settings in the editor preferences, I feel that something like this would just be clutter and doesn't really need to be exposed to the user.

lumberyard-employee-dm: Well I don't know where QSettings are stored on Linux for a user. Perhaps it is a location like (`~/.cache/qt/...')?

I agree it is not an editor preferences settings, but more like Editor runtime metadata that is populated based on users actions.

lumberyard-employee-dm:
Looks like you have a rebase issue here.
I believe you meant to rebase your changes on latest o3de-extras/development ?

Gems/ROS2/Code/Source/RobotImporter/Pages/FileSelectionPage.cpp
mbalfour-amzn: The QSettings are being stored in `~/.config/O3DE/O3DE Editor.conf` .

I've changed this to make the file dialog default to the last selected file and directory, but only when the OK button is pressed. In addition, if you change the file in the text box and then open the file dialog, it now defaults to that file correctly too. If the file doesn't exist, it will fall back to defaulting back to the project path again.

lumberyard-employee-dm: Sounds good to me!

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -341,161 +342,224 @@ namespace ROS2::Utils
...
+
+                // If we successfully matched the prefix, and the replacement path is completely empty, we don't need to look any further.
+                // There's no match.
+                if (replacedUriPath.empty())
                 {
-                    // Strip the number of characters from the Uri scheme from beginning of the path
-                    AZ::IO::PathView strippedUriPath = uriPath.Native().substr(schemePrefix.size());
-                    if (strippedUriPath.empty())
+                    AZ_Trace("ResolveAssetPath", R"(Resolved Path is empty: "%.*s" -> "")" "\n", AZ_PATH_ARG(unresolvedPath));
+                    return "";
```
lumberyard-employee-dm, line 480: This can return `{}` to use the default constructor.
It is slightly faster than initializing an AZ::IO::Path with an empty `const char*`, as a call to strlen still needs to occur internally to determine that the string is empty
```suggestion
                    return {};
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -341,161 +342,224 @@ namespace ROS2::Utils
...
+                        // The file didn't exist, so continue.
+                        continue;
                     }
+                }
 
-                    // The URI path cannot be resolved within the any ament prefix path,
-                    // so try the directory containing the URDF file as well as any of its parent directories
-                    for (const AZ::IO::Path& urdfAncestorPath : urdfAncestorPaths)
+                // The URI path cannot be resolved within the any ament prefix path,
+                // so try the directory containing the root file as well as any of its parent directories
```
lumberyard-employee-dm, line 501: Can you fix the comment to no longer mention the ament prefix path?
```suggestion
                // The URI path is not absolute, so attempt to append it to the ancestor directories of the URDF/SDF file
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -341,161 +342,224 @@ namespace ROS2::Utils
...
-            if (auto resolvedPath = uriPrefix.m_schemeResolver(unresolvedPath);
-                resolvedPath.has_value())
+            else
             {
-                AZ_Printf("ResolveURDFPath", R"(Resolved Path using URI Prefix "%.*s" -> "%.*s")" "\n", AZ_PATH_ARG(unresolvedPath),
-                    AZ_PATH_ARG(resolvedPath.value()));
-                return resolvedPath.value();
+                AZ_Trace("ResolveAssetPath", R"(Failed to resolve Absolute Path: "%.*s")" "\n", 
+                    AZ_PATH_ARG(unresolvedPath));
+                return "";
```
lumberyard-employee-dm, line 530: repeat: returning curly braces causes the `AZ::IO::Path` default constructor be used over the `const char*` constructor which is minuscule-ly faster.
```suggestion
                return {};
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -341,161 +342,224 @@ namespace ROS2::Utils
...
+            return relativePath;
         }
 
-        // The path is relative path, so append to the directory containing the .urdf file
-        const AZ::IO::Path resolvedPath = AZ::IO::Path(urdfFilePath.ParentPath()) / unresolvedPath;
-        AZ_Printf("ResolveURDFPath", "Input Path %s is being returned as is\n", unresolvedPath.c_str());
-        return resolvedPath;
+        AZ_Trace("ResolveAssetPath", R"(Failed to resolve Relative Path: "%.*s" -> "%.*s")" "\n", 
+            AZ_PATH_ARG(unresolvedPath), AZ_PATH_ARG(relativePath));
+        return "";
```
lumberyard-employee-dm, line 547: repeat: returning using list initialization causes the default constructor to be invoked which is infinitesimally quicker. 
```suggestion
        return {};
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -549,4 +613,42 @@ namespace ROS2::Utils::SDFormat
...
+                    uriPath.append(std::string(":"));
+                }
+
+                uriPath.append(std::string(path.c_str(), path.Native().size()));
+            }
+            if (!prefix.empty() && !uriPath.empty())
+            {
+                std::string uriPrefix(prefix.c_str(), prefix.size());
+                sdfConfig.AddURIPath(uriPrefix, uriPath);
+                AZ_Info("SdfParserConfig", "Added URI mapping '%s' -> '%s'", uriPrefix.c_str(), uriPath.c_str());
```
lumberyard-employee-dm, line 640: Well based on the SDFormat documentation for AddURIPath, I guess this will never work for Windows style absolute paths.

I am guessing Robotic model import usage is pretty much doesn't occur on Windows?

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
```diff
@@ -846,46 +859,147 @@ namespace UnitTest
...
             dae,
-            urdf, "",
-            [](const AZ::IO::PathView&) -> bool
+            urdf, "", GetTestSettings(),
+            [expectedResult](const AZ::IO::PathView& p) -> bool
             {
+                // Only a file name that matches the expected result will return that it exists on disk.
+                return p.LexicallyNormal() == expectedResult;
+            });
+        EXPECT_EQ(result.LexicallyNormal(), expectedResult);
```
lumberyard-employee-dm, line 877: There shouldn't be a need to lexically normal any of these paths.
When comparing `AZ::IO::Path*` classes normalization is not needed.

Normalization is only needed when performing string comparison, such when storing the path into an AZStd::string/AZStd::fixed_string.

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
```diff
@@ -846,46 +859,147 @@ namespace UnitTest
...
+    }
+
+    TEST_F(UrdfParserTest, TestPathResolve_InvalidAbsolutePath_ReturnsEmptyPath)
+    {
+        // Verify that an absolute path that isn't found on disk returns an empty path.
+        constexpr AZ::IO::PathView dae = "file:///usr/ros/humble/meshes/bar.dae";
```
lumberyard-employee-dm, line 883: Let me go ahead and make this path locally on file system, just so I can troll O3DE by claiming the ROS2.Editor Unit Test fails.

Gems/ROS2/Code/Source/RobotImporter/Pages/FileSelectionPage.cpp
```diff
@@ -14,15 +14,31 @@
 #include <QFileInfo>
 #include <QHBoxLayout>
 #include <QVBoxLayout>
+#include <QSettings>
 
 namespace ROS2
 {
+    static const char FileSelectionPageDefaultFile[] = "RobotImporter/SelectFileDefaultFile";
```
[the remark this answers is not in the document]
mbalfour-amzn: I've had problems in the past with trying to declare a constexpr string that's used in a lambda. IIRC Visual Studio complains if you _do_ include it in the lambda capture, and clang complains if you _don't_.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
mbalfour-amzn: Yeah, I would guess that as well. There's very little about the ROS / Gazebo stuff so far that seems very tolerant of running under Windows. I'm sure it's _possible_ but painful and unusual enough that it seems pretty unlikely to be done very much in practice.

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
mbalfour-amzn: Guess I can update the comments to put "disk" in quotes...

Gems/ROS2/Code/Source/RobotImporter/Pages/FileSelectionPage.cpp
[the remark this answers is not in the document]
mbalfour-amzn: Nvm, seems to build fine with constexpr, maybe it only failed for me when the static constexpr is local to a function.

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
mbalfour-amzn: I updated the comments to stop referring to a disk.

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
mbalfour-amzn: The unit tests would seem to disagree. I can remove it in most places, but the following is NOT seen as equivalent:
    `/home/foo/ros_ws/install/foo_robot/./meshes/bar.dae`
    `/home/foo/ros_ws/install/foo_robot/meshes/bar.dae`

I'll remove LexicallyNormal() everywhere except for the one test that requires it though.

lumberyard-employee-dm: Yeah if those aren't comparing equal, then that is a bug in the AZ::IO::Path [compare relative path code](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/IO/Path/PathParser.inl#L689-L707)

lumberyard-employee-dm: I'll fix it,

mbalfour-amzn: Ty! Here's a unit test that fails for testing the fix:
```
    TEST_F(PathFixture, OperatorEquals_PathAndPathViewCompareSuccessfullyWithLexicallyEqualPaths)
    {
        AZ::IO::Path path1 = "/home/foo/ros_ws/install/foo_robot/./meshes/bar.dae";
        AZ::IO::PathView path2 = "/home/foo/ros_ws/install/foo_robot/meshes/bar.dae";
        EXPECT_EQ(path1, path2);
    }
```

lumberyard-employee-dm: Thanks, I'll have a fix in a couple of minutes.

lumberyard-employee-dm: Here is the PR with the fix.
https://github.com/o3de/o3de/pull/16674

lumberyard-employee-dm (approved):

michalpelka:
I was not able to test this feature due to the reflection of AZ::IO::Path that wants to open a file.
image

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
```diff
@@ -52,18 +52,57 @@ namespace ROS2
...
+                        AZ::Edit::UIHandlers::Default,
+                        &SdfAssetPathResolverSettings::m_useAmentPrefixPath,
+                        "Use AMENT_PREFIX_PATH",
+                        "Uses the AMENT_PREFIX_PATH environment variable to try and locate asset references")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &SdfAssetPathResolverSettings::m_useAncestorPaths,
+                        "Search parent paths",
+                        "Tries to resolve partial paths by traversing parent folders to look for partial path matches")
+                    ->DataElement(
```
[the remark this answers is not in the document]
mbalfour-amzn: oh, thanks for catching this! i'll switch it back from Path to string. i didn't try editing it in the UI after changing it, i just opened the UI to verify that the existing setreg file was still getting read in correctly.

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
mbalfour-amzn: Removed the last LexicallyNormal() calls, thanks!

Gems/ROS2/Code/Source/SdfAssetBuilder/SdfAssetBuilderSettings.cpp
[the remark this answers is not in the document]
mbalfour-amzn: I've switched it back to using string instead of Path.

lumberyard-employee-dm:
It looks like you have a DCO issue.

jhanca-robotecai:
I am not really sure how to use the feature. I tried to import [Robotis OP3](https://github.com/ROBOTIS-GIT/ROBOTIS-OP3) robot, which uses `package://op3_description`, e.g.
```
<mesh filename="package://op3_description/meshes/rl3.stl"  scale="0.001 0.001 0.001"/>
```
After adding the paths to `package://` resolver:
image
the package is not found:
```
error: package not found: "package 'op3_description' not found, searching: ['/opt/ros/humble']"
when processing file: /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description/urdf/robotis_op3.urdf.xacro
```

I tried adding a slash at the end of the directory and toggling _AMENT_PREFIX_PATH_ on/off.

Both paths are valid:
```bash
jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description                                                                                                                                        
CHANGELOG.rst  CMakeLists.txt  doc  launch  meshes  package.xml  rviz  urdf

jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/                                                                                                                                                       
LICENSE  op3_description  op3_gazebo  README.md  robotis_op3_common
```

jhanca-robotecai:
The feature works correctly for [ANYbotics ANYmal C](https://github.com/ANYbotics/anymal_c_simple_description):

image
image

Gems/ROS2/Registry/sdfassetbuilder_settings.setreg
```diff
@@ -13,7 +13,16 @@
                     "xacro"
                 ],
                 "UseArticulations": true,
-                "URDFPreserveFixedJoint": true
+                "URDFPreserveFixedJoint": true,
+                "AssetResolverSettings":
+                {
+                    "URIPrefixMap":
```
jhanca-robotecai, line 19: I am not really sure how much we want to keep in `.setreg` files, but shouldn't you add _UseAmentPrefixPath_ and _UseAncestorPaths_ as well?

jhanca-robotecai (approved):
The code looks correct and works as expected for `urdf` and `sdf` files and should be an added value for simple `sdf` models. It is unfortunate it cannot be used with `xacro` files, which are used for robots. It is also unfortunate the parameters cannot be added with a file dialog, as it is easy to make a mistake entering long paths.

lumberyard-employee-dm:
> I was not able to test this feature due to the reflection of AZ::IO::Path that wants to open a file. I was trying to replace package:// path with disabled AMENT_PREFIX_PATH. The rendered Qt dialog wants us to save a file not open directory. image

We could make an another Property File Path Handler for AZ::IO::Path that allows for directory selection.
The currently one uses [QFileDialog::getSaveFileName](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyFilePathCtrl.cpp#L137-L151) to allow a user to select a file.

We could use the [QFileDialog::Option::ShowDirsOnly](https://github.com/o3de/o3de/blob/development/Code/Editor/AssetImporter/AssetImporterManager/AssetImporterManager.cpp#L205-L210) option to show a directory picker only and then associate a specialized Property Handler name such as "Directory" for it like we do with the ["FilePath"](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyFilePathCtrl.cpp#L274-L277) handler.

mbalfour-amzn:
> It looks like you have a DCO issue.

Yes, but not ones I directly introduced. These are ones that already exist in the o3de-extras branch that just got detected and highlighted during my accidental rebasing. I'm not quite sure what to do with them:

```
Commit sha: [f7e22ba](https://github.com/o3de/o3de-extras/pull/505/commits/f7e22baf199ac4e38004e85ba5330b81f4b53453), Author: Antoni-Robotec, Committer: Mike Balfour; Expected "Antoni-Robotec [138497503+Antoni-Robotec@users.noreply.github.com](mailto:138497503+Antoni-Robotec@users.noreply.github.com)", but got "Antoni Puch [antoni.puch@robotec.ai](mailto:antoni.puch@robotec.ai)".
Commit sha: [fe57176](https://github.com/o3de/o3de-extras/pull/505/commits/fe5717609dd470222360f96268fa9ef59d202c8b), Author: Jan Hanca, Committer: Mike Balfour; Expected "Jan Hanca [134940295+jhanca-robotecai@users.noreply.github.com](mailto:134940295+jhanca-robotecai@users.noreply.github.com)", but got "Jan Hanca [jan.hanca@robotec.ai](mailto:jan.hanca@robotec.ai)".
Commit sha: [553128d](https://github.com/o3de/o3de-extras/pull/505/commits/553128df65cddf0d09aabb2cfea5cf9a149e8bbc), Author: Jan Hanca, Committer: Mike Balfour; Expected "Jan Hanca [134940295+jhanca-robotecai@users.noreply.github.com](mailto:134940295+jhanca-robotecai@users.noreply.github.com)", but got "Jan Hanca [jan.hanca@robotec.ai](mailto:jan.hanca@robotec.ai)".
Commit sha: [8a4f59a](https://github.com/o3de/o3de-extras/pull/505/commits/8a4f59a8e6e330cb9b047938b52429e761b6f561), Author: Michał Pełka, Committer: Mike Balfour; Expected "Michał Pełka [michal.pelka@robotec.ai](mailto:michal.pelka@robotec.ai)", but got "Michał Pełka [michalpelka@gmail.com](mailto:michalpelka@gmail.com)".
```

mbalfour-amzn:
> Edit: I have just realized that the feature does not work for `.xacro`, so my comment does not make any sense. However, I will keep it to point out this problem.
> 
> I am not really sure how to use the feature. I tried to import [Robotis OP3](https://github.com/ROBOTIS-GIT/ROBOTIS-OP3) robot, which uses `package://op3_description`, e.g.
> 
> ```
> <mesh filename="package://op3_description/meshes/rl3.stl"  scale="0.001 0.001 0.001"/>
> ```
> 
> After adding the paths to `package://` resolver: image the package is not found:
> 
> ```
> error: package not found: "package 'op3_description' not found, searching: ['/opt/ros/humble']"
> when processing file: /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description/urdf/robotis_op3.urdf.xacro
> ```
> 
> I tried adding a slash at the end of the directory and toggling _AMENT_PREFIX_PATH_ on/off.
> 
> Both paths are valid:
> 
> ```shell
> jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
> ± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description                                                                                                                                        
> CHANGELOG.rst  CMakeLists.txt  doc  launch  meshes  package.xml  rviz  urdf
> 
> jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
> ± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/                                                                                                                                                       
> LICENSE  op3_description  op3_gazebo  README.md  robotis_op3_common
> ```

Hmmm. I'll take a look, it _ought_ to work for xacro, I'm not sure what's going wrong there. I can also add a slash at the end of the path replacement if one doesn't exist. This would eliminate the "feature" of allowing a partial name substitution - for "model://test1/test2" having "model://t" -> "b" to turn it into "best1/test2" - but I'm ok with that. It seems like a pretty questionable feature.

mbalfour-amzn:
> > I was not able to test this feature due to the reflection of AZ::IO::Path that wants to open a file. I was trying to replace package:// path with disabled AMENT_PREFIX_PATH. The rendered Qt dialog wants us to save a file not open directory. image
> 
> We could make an another Property File Path Handler for AZ::IO::Path that allows for directory selection. The currently one uses [QFileDialog::getSaveFileName](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyFilePathCtrl.cpp#L137-L151) to allow a user to select a file.
> 
> We could use the [QFileDialog::Option::ShowDirsOnly](https://github.com/o3de/o3de/blob/development/Code/Editor/AssetImporter/AssetImporterManager/AssetImporterManager.cpp#L205-L210) option to show a directory picker only and then associate a specialized Property Handler name such as "Directory" for it like we do with the ["FilePath"](https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyFilePathCtrl.cpp#L274-L277) handler.

I can add this into the backlog as a separate task, but I think it's lower-priority than some of our other work right now. There are also I think a few extra bits beyond what you're saying:
* We'd need to optionally make this control NOT require paths or files that exist inside of a scan path: (https://github.com/o3de/o3de/blob/development/Code/Framework/AzToolsFramework/AzToolsFramework/UI/PropertyEditor/PropertyFilePathCtrl.cpp#L206-L207)
* The picker should be able to pick directories *or* files, because you might want to do full file-name substitutions
* The dialog should have a customizable name to it, because "Open/Save File" doesn't make sense in this case
* The control should allow for typing as well as the browser, it shouldn't _require_ the browser

Gems/ROS2/Registry/sdfassetbuilder_settings.setreg
mbalfour-amzn: They can both be set this way as well, I just didn't add them because I didn't want to change the default settings. I can add them for visibility and clarity though.

The idea behind the setreg file is that this is a way to globally control the default settings for the importer. You can also create your own global overrides by creating a similar file in your user/Registry folder. 

Eventually, it would be nice to get to a point where we are able to load/save these settings either on a per-file basis or just as arbitrary settings that you can load and save from the Robot Importer wizard, so that way if you keep re-importing the same robot you don't have to keep recreating the same settings.

mbalfour-amzn:
> > Edit: I have just realized that the feature does not work for `.xacro`, so my comment does not make any sense. However, I will keep it to point out this problem.
> > I am not really sure how to use the feature. I tried to import [Robotis OP3](https://github.com/ROBOTIS-GIT/ROBOTIS-OP3) robot, which uses `package://op3_description`, e.g.
> > ```
> > <mesh filename="package://op3_description/meshes/rl3.stl"  scale="0.001 0.001 0.001"/>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After adding the paths to `package://` resolver: image the package is not found:
> > ```
> > error: package not found: "package 'op3_description' not found, searching: ['/opt/ros/humble']"
> > when processing file: /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description/urdf/robotis_op3.urdf.xacro
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I tried adding a slash at the end of the directory and toggling _AMENT_PREFIX_PATH_ on/off.
> > Both paths are valid:
> > ```shell
> > jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
> > ± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description                                                                                                                                        
> > CHANGELOG.rst  CMakeLists.txt  doc  launch  meshes  package.xml  rviz  urdf
> > 
> > jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
> > ± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/                                                                                                                                                       
> > LICENSE  op3_description  op3_gazebo  README.md  robotis_op3_common
> > ```
> 
> Hmmm. I'll take a look, it _ought_ to work for xacro, I'm not sure what's going wrong there. I can also add a slash at the end of the path replacement if one doesn't exist. This would eliminate the "feature" of allowing a partial name substitution - for "model://test1/test2" having "model://t" -> "b" to turn it into "best1/test2" - but I'm ok with that. It seems like a pretty questionable feature.

Ok, so looking further into this, the specific error of "'op3_description' not found" is an xacro error, not a urdf parsing error. This appears to be happening before we even reach the urdf piece. I haven't fully tracked down where the error comes from yet, but I believe it's related to the xacro lines that look like this:
```
  <xacro:include filename="$(find op3_description)/urdf/robotis_op3.visuals.xacro" />
```

Since these files look like they can execute arbitrary shell commands in them, I'm not sure there's an easy way to avoid going through the colcon install process for the package to make the op3_description directory findable. We could theoretically do a text search/replace inside the xacro file before handing it over to xacro, but we would still have problems with nested xacro files. We'd also need to determine all of the most common ways in which xacro files refer to other files to know what patterns to search for.

Gems/ROS2/Registry/sdfassetbuilder_settings.setreg
jhanca-robotecai: I was thinking mainly about visibility and clarity: it is easier for users to switch known parameters in the file than to look for possible names in the code. On the other hand, all parameters can be changed in the wizard.

Thanks for the explanation.

mbalfour-amzn: I've added them for the improved visibility.

lumberyard-employee-dm (approved):

mbalfour-amzn:
> > > Edit: I have just realized that the feature does not work for `.xacro`, so my comment does not make any sense. However, I will keep it to point out this problem.
> > > I am not really sure how to use the feature. I tried to import [Robotis OP3](https://github.com/ROBOTIS-GIT/ROBOTIS-OP3) robot, which uses `package://op3_description`, e.g.
> > > ```
> > > <mesh filename="package://op3_description/meshes/rl3.stl"  scale="0.001 0.001 0.001"/>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > After adding the paths to `package://` resolver: image the package is not found:
> > > ```
> > > error: package not found: "package 'op3_description' not found, searching: ['/opt/ros/humble']"
> > > when processing file: /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description/urdf/robotis_op3.urdf.xacro
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I tried adding a slash at the end of the directory and toggling _AMENT_PREFIX_PATH_ on/off.
> > > Both paths are valid:
> > > ```shell
> > > jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
> > > ± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/op3_description                                                                                                                                        
> > > CHANGELOG.rst  CMakeLists.txt  doc  launch  meshes  package.xml  rviz  urdf
> > > 
> > > jhanca@jhanca-buntu ~/devroot/o3de-extras [pr505-path_resolution_setting]
> > > ± % ls /home/jhanca/devroot/robots/ROBOTIS-OP3-Common/                                                                                                                                                       
> > > LICENSE  op3_description  op3_gazebo  README.md  robotis_op3_common
> > > ```
> > 
> > 
> > Hmmm. I'll take a look, it _ought_ to work for xacro, I'm not sure what's going wrong there. I can also add a slash at the end of the path replacement if one doesn't exist. This would eliminate the "feature" of allowing a partial name substitution - for "model://test1/test2" having "model://t" -> "b" to turn it into "best1/test2" - but I'm ok with that. It seems like a pretty questionable feature.
> 
> Ok, so looking further into this, the specific error of "'op3_description' not found" is an xacro error, not a urdf parsing error. This appears to be happening before we even reach the urdf piece. I haven't fully tracked down where the error comes from yet, but I believe it's related to the xacro lines that look like this:
> 
> ```
>   <xacro:include filename="$(find op3_description)/urdf/robotis_op3.visuals.xacro" />
> ```
> 
> Since these files look like they can execute arbitrary shell commands in them, I'm not sure there's an easy way to avoid going through the colcon install process for the package to make the op3_description directory findable. We could theoretically do a text search/replace inside the xacro file before handing it over to xacro, but we would still have problems with nested xacro files. We'd also need to determine all of the most common ways in which xacro files refer to other files to know what patterns to search for.

OK, so just to document what I learned on this one...
The "$(find op3_description)" isn't a shell macro, it's a call into roslaunch.substitution_args:
https://github.com/ros/xacro/blob/noetic-devel/src/xacro/__init__.py#L294-L306

The _find() method then calls down into rospack:
http://docs.ros.org/en/melodic/api/roslaunch/html/roslaunch.substitution_args-pysrc.html

And rospack goes looking for various things in the directories:
http://docs.ros.org/en/independent/api/rospkg/html/_modules/rospkg/rospack.html

Somewhere it's definitely relying on AMENT_PREFIX_PATH, but the problem is that the AMENT_PREFIX_PATH is pointing to the install directory that's coming from `colcon build && source/install.sh`, the directory doesn't exist before that point. So for example, for Universal Robots URx, the ur.urdf.xacro file has `<xacro:include filename="$(find ur_description)/urdf/ur_macro.xacro"/>` ( https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/blob/rolling/urdf/ur.urdf.xacro#L6 ) but there isn't a "ur_description" directory inside of https://github.com/UniversalRobots/Universal_Robots_ROS2_Description . It gets created during the install. I really don't think there's a way to deal with this without requiring the colcon build step.

michalpelka (approved):

lumberyard-employee-dm (approved):

Issue #504, opened 2023-09-07
Robot Importer does not apply scale to colliders


lewandowskikacper:
When Importing robot the `scale` attribute is ignored in colliders. 
It is applied correctly in visual elements.

The example urdf file:

```
<?xml version="1.0"?>
<robot name="urdf_test">  
    <link name="face_link">
        <inertial>
            <origin xyz="0. 0. 0."/>
            <mass value="1."/>
            <inertia ixx="1." ixy="0." ixz="0." iyy="1." iyz="0." izz="1."/>
        </inertial>
        <visual>
            <material name="blue"/>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <mesh filename="face.dae" scale="0.1 0.1 0.1"/>
            </geometry>
        </visual>
        <collision>
            <origin rpy="0 0 0" xyz="0 0 0"/>
            <geometry>
                <mesh filename="face.dae" scale="0.1 0.1 0.1"/>
            </geometry>
        </collision>
    </link>  
</robot>
```

collider - scale is 1, but should be 0.1
![Screenshot from 2023-09-07 16-50-45](https://github.com/o3de/o3de-extras/assets/138626322/dee81f76-7b92-45e4-b59e-6f6835d13478)

visual - scale is correctly 0.1
![Screenshot from 2023-09-07 16-50-55](https://github.com/o3de/o3de-extras/assets/138626322/dcfc03e1-a721-4b63-a055-6e68047afff1)

All needed files (urdf, meshes) are included in this zip archive:
[texture_scale.zip](https://github.com/o3de/o3de-extras/files/12550494/texture_scale.zip)

PR #506, opened 2023-09-08
Robot Importer: set scale of colliders during import

lewandowskikacper:
Fixes https://github.com/o3de/o3de-extras/issues/504.

on the left: before this change, collider is not scaled.
on the right: after change, the scale is applied correctly
Screenshot from 2023-09-08 10-34-54

Collider has the scale set:
image

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
@@ -295,6 +295,8 @@ namespace ROS2
                 assetId.ToString<AZStd::string>().c_str());
 
             Physics::PhysicsAssetShapeConfiguration shapeConfiguration;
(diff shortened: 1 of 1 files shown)


michalpelka (approved):
Very good change!

michalpelka:
I've accepted, but please do me a favor and check if those changes should be applied to primitive colliders as well.

jhanca-robotecai:
> I've accepted, but please do me a favor and check if those changes should be applied to primitive colliders as well.

Both [urdf specs](http://wiki.ros.org/urdf/XML/link) and [sdf specs](http://sdformat.org/spec?ver=1.10&elem=geometry) allow `scale` parameter for mesh only.

jhanca-robotecai (approved):
Tested: works as expected.

pawelbudziszewski (approved):

mbalfour-amzn (approved):

Issue #510, opened 2023-09-11
missing links in robots imported from urdfs with incomplete inertia tags


lewandowskikacper:
https://github.com/o3de/o3de-extras/pull/481 introduced mechanism to fix URDFs in memory.
It adds missing `inertia` links. However, if the `inertia` tag is present but has missing elements, the fix is not applied and import of this link fails.
This can happen for example if user specified `mass`, but not `inertia` (inertia matrix).

PR #511, opened 2023-09-11
fix incomplete inertial tags in urdfs before import

lewandowskikacper:
Fixes https://github.com/o3de/o3de-extras/issues/510.
Each tag inside `inertia` is checked and the value is set to default if missing.

Before this change importing this robot (https://github.com/RobotnikAutomation/rbvogui_common/blob/melodic-devel/rbvogui_description/robots/rbvogui_std_ur10.urdf.xacro) results in missing wheels:
Screenshot from 2023-09-11 13-30-21
Wheel links are not modified:
image

After this change, the wheels are visible:
Screenshot from 2023-09-11 13-29-54
Wheel links are modified:
image

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
@@ -21,48 +21,65 @@ namespace ROS2::Utils
     {
         AZStd::vector<AZStd::string> modifiedLinks;
         using namespace AZ::rapidxml;
-        AZStd::vector<xml_node<>*> links;
+
         for (xml_node<>* link = urdf->first_node("link"); link; link = link->next_sibling("link"))
         {
-            if (!link->first_node("inertial"))
+            bool modified = false;
+            bool inertiaPresent = true;
+
+            auto name_xml = link->first_attribute("name");
+            AZStd::string name = "unknown_link";
+            if (name_xml)
             {
-                links.push_back(link);
+                name = name_xml->value();
             }
-        }
 
-        for (auto* link : links)
-        {
-            xml_node<>* inertial = urdf->document()->allocate_node(node_element, "inertial");
-
-            xml_node<>* mass = urdf->document()->allocate_node(node_element, "mass");
-            mass->append_attribute(urdf->document()->allocate_attribute("value", "1."));
-            inertial->append_node(mass);
-            xml_node<>* inertia = urdf->document()->allocate_node(node_element, "inertia");
-
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixx", "1."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixy", "0."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixz", "0."));
-
-            inertia->append_attribute(urdf->document()->allocate_attribute("iyx", "0."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("iyy", "1."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("iyz", "0."));
-
-            inertia->append_attribute(urdf->document()->allocate_attribute("izx", "0."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("izy", "0."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("izz", "1."));
-
-            inertial->append_node(inertia);
-
-            xml_node<>* origin = urdf->document()->allocate_node(node_element, "origin");
-            origin->append_attribute(urdf->document()->allocate_attribute("xyz", "0. 0. 0."));
-            inertial->append_node(origin);
-
-            auto* name = link->first_attribute("name");
-            if (name)
+            auto inertial = link->first_node("inertial");
+            if (!inertial)
             {
-                modifiedLinks.push_back(name->value());
+                AZ_Warning("URDF", false, "Missing inertial tag in link %s, applying default tag.", name.c_str());
+                inertial = urdf->document()->allocate_node(node_element, "inertial");
+                link->append_node(inertial);
+                modified = true;
+                inertiaPresent = false;
             }
-            link->append_node(inertial);
+
+            auto mass = inertial->first_node("mass");
+            if (!mass)
+            {
+                AZ_Warning("URDF", !inertiaPresent, "Missing mass tag inside the inertial tag link %s, applying default mass tag.", name.c_str());
+                mass = urdf->document()->allocate_node(node_element, "mass");
+                mass->append_attribute(urdf->document()->allocate_attribute("value", "1."));
+                inertial->append_node(mass);
+                modified = true;
+            }
+
+            auto inertia = inertial->first_node("inertia");
+            if (!inertia)
+            {
+                AZ_Warning("URDF", !inertiaPresent, "Missing inertia tag inside the inertial tag link %s, applying default inertia tag.", name.c_str());
+                inertia = urdf->document()->allocate_node(node_element, "inertia");
+
+                inertia->append_attribute(urdf->document()->allocate_attribute("ixx", "1."));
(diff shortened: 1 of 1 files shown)


Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -16,53 +16,94 @@ namespace ROS2::Utils
...
+        AZStd::vector<AZStd::string> modifiedLinks;
+
+        for (auto [link, tags] : modifiedLinksTags)
+        {
+            AZStd::string modified = link + ":";
+
+            for (auto tag : tags)
             {
-                modifiedLinks.push_back(name->value());
+                modified += " " + tag + ", ";
```
michalpelka, line 103: I think that needs to be addressed in a more systematic way. Currently, it is even more confusing. Please revert this logic of return value here. 
Let it fix correctly in next PR, ok?

lewandowskikacper: I reverted back to showing just the link names.

lewandowskikacper: I opened an issue https://github.com/o3de/o3de-extras/issues/512 for discussion of the information for user.

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -21,48 +21,67 @@ namespace ROS2::Utils
...
         for (xml_node<>* link = urdf->first_node("link"); link; link = link->next_sibling("link"))
         {
-            if (!link->first_node("inertial"))
+            bool modified = false;
+
+            auto inertial = link->first_node("inertial");
+            if (!inertial)
             {
-                links.push_back(link);
+                inertial = urdf->document()->allocate_node(node_element, "inertial");
```
spham-amzn, line 32: I think a warning when we fall back to default values would be helpful in these situations
```suggestion
                AZ_Warning("URDF", false, "Missing inertial tag, applying default tag.");
                inertial = urdf->document()->allocate_node(node_element, "inertial");
```

michalpelka: +1 on that

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -21,48 +21,70 @@ namespace ROS2::Utils
...
+            }
 
-            xml_node<>* mass = urdf->document()->allocate_node(node_element, "mass");
-            mass->append_attribute(urdf->document()->allocate_attribute("value", "1."));
-            inertial->append_node(mass);
-            xml_node<>* inertia = urdf->document()->allocate_node(node_element, "inertia");
+            auto mass = inertial->first_node("mass");
+            if (!mass)
+            {
+                mass = urdf->document()->allocate_node(node_element, "mass");
```
spham-amzn, line 48: A similar warning would be helpful here as well, but only if `modified` is `false`, since i am assuming a default tag won't have this.
```suggestion
                if (!modified)
                {
                    AZ_Warning("URDF", false, "Missing mass tag inside the inertial tag link %s, applying default mass tag.", name.c_str());
                }
                mass = urdf->document()->allocate_node(node_element, "mass");
                
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -21,48 +21,70 @@ namespace ROS2::Utils
...
+                modified = true;
+            }
 
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixx", "1."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixy", "0."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixz", "0."));
+            auto inertia = inertial->first_node("inertia");
+            if (!inertia)
+            {
+                inertia = urdf->document()->allocate_node(node_element, "inertia");
```
spham-amzn, line 57: Same comment about the `mass` warning above

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -21,48 +21,70 @@ namespace ROS2::Utils
...
+                inertial->append_node(inertia);
+                modified = true;
+            }
 
-            auto* name = link->first_attribute("name");
-            if (name)
+            auto origin = inertial->first_node("origin");
+            if (!origin)
+            {
+                origin = urdf->document()->allocate_node(node_element, "origin");
```
spham-amzn, line 78: Same comment about the mass warning above

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
lewandowskikacper: When the 'inertial' tag is missing, also all subtags are added.
But since they are always added in this case, I think it is unnecessary to warn user about each one separately. 

I added a warning in case the `inertial` tag is present, but the subtag is missing.

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
lewandowskikacper: Origin tag is [optional](http://wiki.ros.org/urdf/XML/link).  libsdformat [only uses it if available](https://github.com/gazebosim/sdformat/blob/b999425da3ea58aa659c6a9d389b8f9f6b82b9e3/src/urdf/urdf_parser/link.cpp#L281C1-L287C4).
I think there is no need for a warning, and in fact for adding this value here.

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -21,48 +21,72 @@ namespace ROS2::Utils
...
-            mass->append_attribute(urdf->document()->allocate_attribute("value", "1."));
-            inertial->append_node(mass);
-            xml_node<>* inertia = urdf->document()->allocate_node(node_element, "inertia");
+            auto mass = inertial->first_node("mass");
+            if (!mass)
+            {
+                if(inertiaPresent)
+                {
+                    AZ_Warning("URDF", false, "Missing mass tag inside the inertial tag link %s, applying default mass tag.", name.c_str());
+                }
```
mbalfour-amzn, line 53: Suggestion: You can simplify this a little bit.
```suggestion
                AZ_Warning("URDF", !inertiaPresent, "Missing mass tag inside the inertial tag link %s, applying default mass tag.", name.c_str());
```

Gems/ROS2/Code/Source/RobotImporter/FixURDF/FixURDF.cpp
```diff
@@ -21,48 +21,72 @@ namespace ROS2::Utils
...
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixx", "1."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixy", "0."));
-            inertia->append_attribute(urdf->document()->allocate_attribute("ixz", "0."));
+            auto inertia = inertial->first_node("inertia");
+            if (!inertia)
+            {
+                if(inertiaPresent)
+                {
+                    AZ_Warning("URDF", false, "Missing inertia tag inside the inertial tag link %s, applying default inertia tag.", name.c_str());
+                }
```
mbalfour-amzn, line 66: You could simplify a bit here too:
```suggestion
                AZ_Warning("URDF", !inertiaPresent, "Missing inertia tag inside the inertial tag link %s, applying default inertia tag.", name.c_str());
```

mbalfour-amzn (approved):

michalpelka (approved):

PR #523, opened 2023-09-19
Updated the URDFPrefabMaker to avoid referencing SDF link names 

lumberyard-employee-dm:
Link names within an SDF are only unique with a <model> tag, not unique across the entire SDF document.

The `CreatePrefabTemplateFromUrdfOrSdf` function has been modified to create a JointMapper object which stores a mapping of joints to all their parent and child links. It also creates a mapping of joints to the model they are directly attached to.

A similar LinkMapper structure has been added for SDF links, which stores a mapping of the link to their attached model.

Instead of `CreatePrefabTemplateFromUrdfOrSdf` creating a mapping of SDF link name to O3DE Entity, that mapping has now been updated toi map the SDF link pointer to O3DE Entity. This ensures that SDF links with same name are not lost when mapping to O3DE entities. This is needed as SDF links and joints are only required to be unique relative to their attached model. Two links on different models, can have the same link name (http://sdformat.org/tutorials?tut=composition_proposal#1-6-proposed-parsing-stages)

Updated the Robot Importer Visitor callback to require "ModelStack" parameters, which passes in the stack of `sdf::Model` objects that were visited on the way to the sdf element being visited be that a link, joint or nested model. Also updated the GetAllJoints and GetAllLinks function to store a fully qualified name for the joint/link including the ancestor model names when returning the name to joint and name to link map respectively.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.cpp
@@ -222,7 +222,8 @@ namespace ROS2
         };
 
         // Use the URDF/SDF file name stem the prefab name
-        AZStd::string prefabName = AZStd::string(AZ::IO::PathView(filePath).Stem().Native());
+        auto fileStem = AZ::IO::PathView(filePath).Stem();
+        AZStd::string prefabName = AZStd::string::format("%.*s.prefab", AZ_PATH_ARG(fileStem));
 
         if (prefabName.empty())
         {
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
@@ -72,7 +72,7 @@ namespace ROS2
             };
 
             // Iterate over all visuals to get their materials
-            auto VisitAllModels = [&GetVisualsFromModel](const sdf::Model& model) -> Utils::VisitModelResponse
+            auto VisitAllModels = [&GetVisualsFromModel](const sdf::Model& model, const Utils::ModelStack&) -> Utils::VisitModelResponse
             {
                 GetVisualsFromModel(model);
                 // Continue to visit all models within the SDF document and query their <visual> tags
@@ -82,40 +82,6 @@ namespace ROS2
 
             m_visualsMaker = VisualsMaker(AZStd::move(materialMap), urdfAssetsMapping);
         }
-    }
-
-    void URDFPrefabMaker::BuildAssetsForLink(const sdf::Link* link)
-    {
-        m_collidersMaker.BuildColliders(link);
-
-        auto GetAssetsForLinkInModel = [this, link](const sdf::Model& model) -> Utils::VisitModelResponse
-        {
-            // Find the links which are children in a joint where this link is a parent
-            auto BuildAssetsFromJointChildLinks = [this, &model](const sdf::Joint& joint)
-            {
-                if (const sdf::Link* childLink = model.LinkByName(joint.ChildName()); childLink != nullptr)
-                {
-                    BuildAssetsForLink(childLink);
-                }
-
-                return true;
-            };
-
-            // Make sure the link is a child of the model being visited
-            // Before visiting the joints of the model
-            if (const sdf::Link* searchLink = model.LinkByName(link->Name()); searchLink != nullptr)
-            {
-                // Don't visit nested models using the VisitJoints function as the outer call
-                // to VisitModels already visits nested models
-                constexpr bool visitNestedModelLinks = false;
-                Utils::VisitJoints(model, BuildAssetsFromJointChildLinks, visitNestedModelLinks);
-            }
-
-            return Utils::VisitModelResponse::VisitNestedAndSiblings;
-        };
-
-        constexpr bool visitNestedModels = true;
-        Utils::VisitModels(*m_root, GetAssetsForLinkInModel, visitNestedModels);
     }
 
     URDFPrefabMaker::CreatePrefabTemplateResult URDFPrefabMaker::CreatePrefabTemplateFromUrdfOrSdf()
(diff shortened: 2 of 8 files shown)


Gems/ROS2/Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.cpp
```diff
@@ -190,7 +190,8 @@ namespace ROS2
         };
 
         // Use the URDF/SDF file name stem the prefab name
-        AZStd::string prefabName = AZStd::string(AZ::IO::PathView(filePath).Stem().Native());
+        auto fileStem = AZ::IO::PathView(filePath).Stem();
+        AZStd::string prefabName = AZStd::string::format("%.*s.prefab", AZ_PATH_ARG(fileStem));
```
mbalfour-amzn, line 194: Suggestion: As an alternative, you could do...
```
AZ::IO::Path prefabName = filePath;
prefabName.ReplaceExtension(".prefab");
```

Not sure whether that or string::format would run faster, I've got 0 objections with leaving it the way it is too.

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -84,103 +84,140 @@ namespace ROS2
...
-        m_collidersMaker.BuildColliders(link);
+        {
+            AZStd::lock_guard<AZStd::mutex> lck(m_statusLock);
+            m_status.clear();
+        }
+
+        if (!ContainsModel())
+        {
+            return AZ::Failure(AZStd::string("URDF/SDF doesn't contain any models."));
+        }
```
mbalfour-amzn, line 97: Is this check necessary? Can't we have a valid <world> without a <model>? 

Admittedly, it would probably be a pretty useless conversion, so maybe it's ok to check this...

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -84,103 +84,140 @@ namespace ROS2
...
+        auto GetAllLinksFromModel = [&linksMapper](const sdf::Model& model, const Utils::ModelStack&) -> Utils::VisitModelResponse
         {
             // As the VisitModels function visits nested models by default, gatherNestedModelLinks is set to false
             constexpr bool gatherNestedModelLinks = false;
             auto linksForModel = Utils::GetAllLinks(model, gatherNestedModelLinks);
-            links.insert(linksForModel.begin(), linksForModel.end());
-
+            for (const auto& [fullyQualifiedName, link] : linksForModel)
+            {
+                // Push back the mapping of link to attached model into the orderd vector
```
mbalfour-amzn, line 171:
```suggestion
                // Push back the mapping of link to attached model into the ordered vector
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -43,38 +43,47 @@ namespace ROS2::Utils
...
+    using JointVisitorCallback = AZStd::function<bool(const sdf::Joint& joint, const ModelStack& modelStack)>;
     //! Visit joints from URDF/SDF
     //! @param sdfModel Model object of SDF document corresponding to the <model> tag. It used to query joints
     //! @param visitNestedModelJoints When true recurses to any nested <model> tags of the Model object and invoke visitor on their joints
     //! as well
     //! @returns void
     void VisitJoints(const sdf::Model& sdfModel, const JointVisitorCallback& jointVisitorCB, bool visitNestedModelJoints = false);
 
-    //! Retrieve all joints in URDF/SDF.
+    //! Retrieve all joints in URDF/SDF wehre the key is the full
```
mbalfour-amzn, line 82:
```suggestion
    //! Retrieve all joints in URDF/SDF where the key is the full
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -43,38 +43,47 @@ namespace ROS2::Utils
...
     //! @param sdfModel Model object of SDF document corresponding to the <model> tag. It used to query joints
-    //! @param gatherNestedModelJoints When true recurses to any nested <model> tags of the Model object and also gathers their joints as
+    //! @param gatherNestedModelJoints When true recurses to any nested <model> tags of the Model object and also gathers their joints as well
     //! well
-    //! @returns mapping from joint name to joint pointer
+    //! @returns mapping from fully qualified joint name(such as "model_name::joint_name") to joint pointer
```
mbalfour-amzn, line 86: Line 85 (`//! well`) should get deleted

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -129,7 +141,8 @@ namespace ROS2::Utils
 
     //! Returns the SDF model object which contains the specified link
     //! @param root - reference to SDF Root object representing the root of the parsed SDF xml document
-    //! @param linkName - Name of SDF link to lookup in the SDF document
+    //! @param linkName - Fully qualified name of of SDF link to lookup in the SDF document
+    //! A fully qualified can be of the form "modelname1::nested_modelname1::linkname"
```
mbalfour-amzn, line 145:
```suggestion
    //! A fully qualified name is of the form "modelname1::nested_modelname1::linkname"
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -138,9 +151,10 @@ namespace ROS2::Utils
     //! Returns the SDF model object which contains the specified joint
     //! @param root - reference to SDF Root object representing the root of the parsed SDF xml document
     //! @param jointName - Name of SDF joint to lookup in the SDF document
+    //! A fully qualified can be of the form "modelname1::nested_modelname1::jointname"
```
mbalfour-amzn, line 154:
```suggestion
    //! A fully qualified name is of the form "modelname1::nested_modelname1::jointname"
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
```diff
@@ -138,9 +151,10 @@ namespace ROS2::Utils
...
     const sdf::Model* GetModelContainingJoint(const sdf::Root& root, AZStd::string_view jointName);
     //! @param root - reference to SDF Root object representing the root of the parsed SDF xml document
-    //! @param jointName - Name of SDF joint to lookup in the SDF document
+    //! @param joint- SDF joint pointer to lookup in the SDF document
```
mbalfour-amzn, line 157: Technically we're not supposed to have a '-' here for DOxygen, but for consistency with the parameter above it there should be a space before and after the dash.
```suggestion
    //! @param joint - SDF joint pointer to lookup in the SDF document
```

Gems/ROS2/Code/Tests/UrdfParserTest.cpp
```diff
@@ -601,10 +601,9 @@ namespace UnitTest
         EXPECT_EQ("FooRobot", model->Name());
         ASSERT_NE(nullptr, model);
 
-        // Due to joint reduction there should be 2 links and 43 frames
+        // Due to joint reduction there should be 2 links and 4 frames
         // This is different from the previous test, as the root link
-        // can now partipate in joint reduction because it has a name
-        // that isn't "world"
+        // can now partipate in joint reduction because its name isn't "world"
```
mbalfour-amzn, line 606:
```suggestion
        // can now participate in joint reduction because its name isn't "world"
```

mbalfour-amzn (approved):

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.h
lumberyard-employee-dm: I can remove the dashes from all parameters

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
lumberyard-employee-dm: Technically the check isn't necessary, It would just result in a prefab with no entities in it.

mbalfour-amzn: Once we support `<light>`, we could technically have `<world><light> ... </light></world>` that would have an entity but no models I guess. Still seems like a pretty silly and contrived example though, so I'm pretty ambivalent on the check, just pointing it out.

lumberyard-employee-dm: Well at that point we can update the function to check if the SDF contains a model or a light.

Gems/ROS2/Code/Source/RobotImporter/ROS2RobotImporterEditorSystemComponent.cpp
lumberyard-employee-dm: The logic would still have to call `.File()` has the filePath is the full path containing any parent directories.
Technically it could be something such as
```
AZ::IO::Path prefabName = AZ::IO::Path(filePath).Filename();
prefabName.ReplaceExtension(".prefab");
```

Now the AZ::IO::Path call would construct a temporary AZStd::string under the hood and then create another AZStd::string that contains the full file name, such as  `/home/user/URDF/ur.urdf.xacro` -> `ur.urdf.xacro`.
The replace extension call would update the string in place which probably will not need to allocate if the string is <22 characters or the allocated memory can fit an additional character due to replacing "xacro" with "prefab"

michalpelka:
I've rebased it locally to test but I've one test failing:
```
143: [  FAILED  ] UrdfParserTest.TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly
```
Here is branch : https://github.com/RobotecAI/o3de-extras/tree/mp/sdf-multi-model-support-fixes_rb
How to test PR? I've tested it against the URDF zoo, but I think better would be to run it against multimodel SDFs. Is there any test files?

lumberyard-employee-dm:
> I've rebased it locally to test but I've one test failing:
> 
> ```
> 143: [  FAILED  ] UrdfParserTest.TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly
> ```
> 
> Here is rebased branch : https://github.com/RobotecAI/o3de-extras/tree/mp/sdf-multi-model-support-fixes_rb How to test PR? I've tested it against the URDF, but I think better would be to run it against multimodel SDFs. Is there any test files? Edit: The test `TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly` fails for development as well 🙈

That test failure isn't due to this change, but do combination that the [UrdfParserTests.cpp](https://github.com/o3de/o3de-extras/blame/development/Gems/ROS2/Code/Tests/UrdfParserTest.cpp#L984) is using an updated version of the AZ::IO::Path CompareRelative internal function to fix comparing paths with <dot> paths in them added in this commit https://github.com/o3de/o3de/blame/3d70b5c813119f14723409dd01b339a6be8afded/Code/Framework/AzCore/AzCore/IO/Path/PathParser.inl#L773-L776.
That commit is only available in the o3de **development** not **stabilization/2310** currently, so the test only pass in when using the `o3de` repo development branch
The test passes in o3de **development**

lumberyard-employee-dm:
> > I've rebased it locally to test but I've one test failing:
> > ```
> > 143: [  FAILED  ] UrdfParserTest.TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Here is rebased branch : https://github.com/RobotecAI/o3de-extras/tree/mp/sdf-multi-model-support-fixes_rb How to test PR? I've tested it against the URDF, but I think better would be to run it against multimodel SDFs. Is there any test files? Edit: The test `TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly` fails for development as well 🙈
> 
> That test failure isn't due to this change, but do combination that the [UrdfParserTests.cpp](https://github.com/o3de/o3de-extras/blame/development/Gems/ROS2/Code/Tests/UrdfParserTest.cpp#L984) is using an updated version of the AZ::IO::Path CompareRelative internal function to fix comparing paths with paths in them added in this commit https://github.com/o3de/o3de/blame/3d70b5c813119f14723409dd01b339a6be8afded/Code/Framework/AzCore/AzCore/IO/Path/PathParser.inl#L773-L776. That commit is only available in the o3de **development** not **stabilization/2310** currently, so the test only pass in when using the `o3de` repo development branch The test passes in o3de **development**

Now there are three choices here that can resolve or mitigate the issue.
1. As the failure is only in the `UrdfParserTest.TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly` Unit Test and not in any actual functionality due to that test comparing a path of 
```c++
"/home/foo/ros_ws/install/foo_robot/meshes/bar.dae" == "/home/foo/ros_ws/install/foo_robot/./meshes/bar.dae"
```
The test failure can be ignored.

2. The AZ IO Path [changes](https://github.com/o3de/o3de/commit/183d594453d712e4d9991f96e502bd0f3e847faa) can be cherry picked into stabilization/2310 and that would resolve the test failure
3. The test code can be modified to use a call to `LexicallyNormal()` before the comparison, which would have the comparison succeed as in an [older revision of the code](https://github.com/o3de/o3de-extras/pull/505/commits/3daf84ba43ac19ed0a37e68bdea47fe8f84cce3b#diff-51028f9e7742ab6c69b09457939ea407a0b1ff06edb5d451eaf5e8c0e057db36L982-L985)

lumberyard-employee-dm:
> Here is rebased branch : https://github.com/RobotecAI/o3de-extras/tree/mp/sdf-multi-model-support-fixes_rb How to test PR? I've tested it against the URDF, but I think better would be to run it against multimodel SDFs. Is there any test files?


For test files, the `aws-robomaker-hospital-world` repo [hospital.world](https://github.com/aws-robotics/aws-robomaker-hospital-world/blob/ros1/worlds/hospital.world) SDF file and the `aws-robomaker-bookstore-world` repo [bookstore.world](https://github.com/aws-robotics/aws-robomaker-bookstore-world/blob/ros1/worlds/bookstore.world) contains a <world> tag with multiple models within them and those files are used for testing this change.

lumberyard-employee-dm:
> > > I've rebased it locally to test but I've one test failing:
> > > ```
> > > 143: [  FAILED  ] UrdfParserTest.TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Here is rebased branch : https://github.com/RobotecAI/o3de-extras/tree/mp/sdf-multi-model-support-fixes_rb How to test PR? I've tested it against the URDF, but I think better would be to run it against multimodel SDFs. Is there any test files? Edit: The test `TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly` fails for development as well 🙈
> > 
> > 
> > That test failure isn't due to this change, but do combination that the [UrdfParserTests.cpp](https://github.com/o3de/o3de-extras/blame/development/Gems/ROS2/Code/Tests/UrdfParserTest.cpp#L984) is using an updated version of the AZ::IO::Path CompareRelative internal function to fix comparing paths with paths in them added in this commit https://github.com/o3de/o3de/blame/3d70b5c813119f14723409dd01b339a6be8afded/Code/Framework/AzCore/AzCore/IO/Path/PathParser.inl#L773-L776. That commit is only available in the o3de **development** not **stabilization/2310** currently, so the test only pass in when using the `o3de` repo development branch The test passes in o3de **development**
> 
> Now there are three choices here that can resolve or mitigate the issue.
> 
>     1. As the failure is only in the `UrdfParserTest.TestPathResolve_ValidPathRelativeToAncestorPath_ResolvesCorrectly` Unit Test and not in any actual functionality due to that test comparing a path of
> 
> 
> ```c++
> "/home/foo/ros_ws/install/foo_robot/meshes/bar.dae" == "/home/foo/ros_ws/install/foo_robot/./meshes/bar.dae"
> ```
> 
> The test failure can be ignored.
> 
>     2. The AZ IO Path [changes](https://github.com/o3de/o3de/commit/183d594453d712e4d9991f96e502bd0f3e847faa) can be cherry picked into stabilization/2310 and that would resolve the test failure
> 
>     3. The test code can be modified to use a call to `LexicallyNormal()` before the comparison, which would have the comparison succeed as in an [older revision of the code](https://github.com/o3de/o3de-extras/pull/505/commits/3daf84ba43ac19ed0a37e68bdea47fe8f84cce3b#diff-51028f9e7742ab6c69b09457939ea407a0b1ff06edb5d451eaf5e8c0e057db36L982-L985)

The AZ::IO::Path changes have been cherry picked to the `o3de` repo stabilization/2310 branch in PR https://github.com/o3de/o3de/pull/16798, so the Unit Test will now succeed when using the stabilization branch

michalpelka (approved):
My previous concerns were unrelated to this change. Approving.

Issue #484, opened 2023-08-30
IMU is normalized by PhysX frequency not sensors frequency


michalpelka:
Create such setup:
 - Dynamic RigidBody component (without any damping and colliders)
![image](https://github.com/o3de/o3de-extras/assets/3209244/ae802ad8-4667-467e-a38c-b90680590853)
 - IMU with graviational disabled and frequency set to 10 Hz.
![image](https://github.com/o3de/o3de-extras/assets/3209244/85a90046-55a2-4c65-b02c-a12641a77cd2)
- Start simulation and subscribe to IMU message.

What should be expected - acceleration about 10 m/s^2.
But we observe 200 m/s2 for 10 Hz:
![image](https://github.com/o3de/o3de-extras/assets/3209244/43b38930-cf32-478d-89d7-81db369d3f71)
Changing sensors frequency to 100 Hz we observe 20 m/s2:
![image](https://github.com/o3de/o3de-extras/assets/3209244/2d03fb47-aa8e-411b-b951-7ef5cb666592)

Please make sure that logic around computing and filtering acceleration's magnitude is correct.

michalpelka:
It was resovled with https://github.com/o3de/o3de-extras/pull/564.

PR #526, opened 2023-09-22
Feature/sensor base update

PawelLiberadzki:
This code closes the refactor of sensor event management topic - #295. After reworking lidars and imu, this request introduced similar changes to other sensors and removes old approach (ROS2SensorComponent and PhysicsCallbackHandler).

Sensor tests were performed and described here:
https://github.com/RobotecAI/o3de-ros2-gem-testing/pull/10


Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.cpp
```diff
@@ -107,21 +108,24 @@ namespace ROS2
...
+        if (!m_cameraSensor)
         {
-            std_msgs::msg::Header messageHeader;
-            messageHeader.stamp = timestamp;
-            messageHeader.frame_id = m_frameName.c_str();
-            m_cameraSensor->RequestMessagePublication(transform, messageHeader);
+            return;
         }
+
+        const AZ::Transform transform = GetEntity()->GetTransform()->GetWorldTM();
```
michalpelka, line 116:
```suggestion
        const AZ::Transform& transform = GetEntity()->GetTransform()->GetWorldTM();
```

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -40,34 +41,36 @@ namespace ROS2
...
         // AzToolsFramework::Components::EditorComponentBase overrides ..
         void Activate() override;
         void Deactivate() override;
 
         // CameraCalibrationRequestBus::Handler overrides ...
-        AZ::Matrix3x3 GetCameraMatrix() const override;
-        int GetWidth() const override;
-        int GetHeight() const override;
-        float GetVerticalFOV() const override;
+        [[nodiscard]] AZ::Matrix3x3 GetCameraMatrix() const override;
```
michalpelka, line 53: I would not omit [[nodiscard]] since it is part of CameraCalibrationRequestBus::Handler. It should be called directly from ebus system. I cannot see such a pattern used in o3de source code, so we are probably challenging other compilers (GCC, VS here). We should not use nodiscard at all for events in all this contribution.
I think it will cause some problems in future.

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.h
```diff
@@ -40,34 +41,36 @@ namespace ROS2
...
 
     private:
         //! Helper that adds an image source.
         //! @tparam CameraType type of camera sensor (eg 'CameraColorSensor')
         template<typename CameraType>
-        void AddImageSource()
+        void SetImageSource()
         {
-            auto cameraName = GetCameraNameFromFrame(GetEntity());
+            const auto cameraName = GetCameraNameFromFrame(GetEntity());
```
michalpelka, line 64:
```suggestion
            const auto& cameraName = GetCameraNameFromFrame(GetEntity());
```

Gems/ROS2/Code/Source/GNSS/ROS2GNSSSensorComponent.h
```diff
@@ -36,15 +37,14 @@ namespace ROS2
...
 
-        //////////////////////////////////////////////////////////////////////////
-        // ROS2SensorComponent overrides
-        void FrequencyTick() override;
-        //////////////////////////////////////////////////////////////////////////
-
-        AZ::Transform GetCurrentPose() const;
+        //! Returns current entity position.
+        //! @return Current entity position.
+        [[nodiscard]] AZ::Transform GetCurrentPose() const;
```
michalpelka, line 45: For visibility - that should be event handled by ebus.

Gems/ROS2/Code/Source/Manipulation/JointStatePublisher.cpp
```diff
@@ -47,7 +54,7 @@ namespace ROS2
         m_jointStatePublisher->publish(m_jointStateMsg);
     }
 
-    void JointStatePublisher::InitializePublisher(AZ::EntityId entityId)
+    void JointStatePublisher::InitializePublisher([[maybe_unused]] AZ::EntityId entityId)
```
michalpelka, line 57: Remove the parameter if it is not needed anymore. We have already broken API.

Gems/ROS2/Code/Source/Odometry/ROS2OdometrySensorComponent.cpp
```diff
@@ -55,34 +53,28 @@ namespace ROS2
...
+            auto* sceneInterface = AZ::Interface<AzPhysics::SceneInterface>::Get();
+            auto* simulatedBodyPtr = sceneInterface->GetSimulatedBodyFromHandle(sceneHandle, m_bodyHandle);
+            auto* rigidbodyPtr = azrtti_cast<AzPhysics::RigidBody*>(simulatedBodyPtr);
+            AZ_Assert(rigidbodyPtr, "Requested simulated body is not a rigid body")
+            m_initialTransform = rigidbodyPtr->GetTransform();
+        }
 
-    void ROS2OdometrySensorComponent::OnPhysicsSimulationFinished(AzPhysics::SceneHandle sceneHandle, float deltaTime)
-    {
         auto* sceneInterface = AZ::Interface<AzPhysics::SceneInterface>::Get();
```
michalpelka, line 74: AZ_Assert(sceneInterface, "No sceneInterface")

Gems/ROS2/Code/Source/Odometry/ROS2WheelOdometry.cpp
```diff
@@ -114,18 +106,31 @@ namespace ROS2
         m_odometryMsg.child_frame_id = GetFrameID().c_str();
 
         auto ros2Node = ROS2Interface::Get()->GetNode();
-        AZ_Assert(m_sensorConfiguration.m_publishersConfigurations.size() == 1, "Invalid configuration of publishers for Odometry sensor");
+        AZ_Assert(m_sensorConfiguration.m_publishersConfigurations.size() == 1, "Invalid configuration of publishers for Odometry sensor")
 
-        const auto publisherConfig = m_sensorConfiguration.m_publishersConfigurations[Internal::kWheelOdometryMsgType];
+        const auto publisherConfig = m_sensorConfiguration.m_publishersConfigurations[WheelOdometryMsgType];
```
michalpelka, line 111:
```suggestion
        const auto& publisherConfig = m_sensorConfiguration.m_publishersConfigurations[WheelOdometryMsgType];
```

michalpelka (commented):
Nitpicking and questioned [[nodiscard]] for events.

adamdbrw:
I see there is also a conflict to rebase @PawelLiberadzki

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.h
PawelLiberadzki: Fair point, I removed this from public interface.

Gems/ROS2/Code/Source/Camera/ROS2CameraSensorComponent.h
PawelLiberadzki: This will not work here, because GetCameraNameFromFrame returns by value. I keep it as it is.

Gems/ROS2/Code/Source/Odometry/ROS2OdometrySensorComponent.cpp
PawelLiberadzki: We talked about this for the previous PR - then, I assumed that scene interface is retrieved from engine and it will not return nullptr. Do you think, that this is necessary/usefull anyway?

michalpelka (approved):

michalpelka:
Tested work as inteded. https://github.com/o3de/o3de-extras/issues/484 is still present

PR #529, opened 2023-09-22
Fix ROS2 remote gem compatibility

AMZN-alexpete:
The repo.json gem compatibility does not match what is in the actual ROS2 gem.json causing the ROS2 gem to fail compatibility checks before downloading in Project Manager. 

*Testing*
I am not currently setup to test this on Linux, but at the very least this non-code change shouldn't break things any worse than before.

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -123,7 +123,7 @@
                 "ROS2"
             ],
             "compatible_engines": [
-                "o3de-sdk==1.2.0",
+                "o3de-sdk>=1.2.0",
                 "o3de>=1.2.0"
             ],
             "icon_path": "preview.png",
--- a/repo.json
+++ b/repo.json
@@ -123,8 +123,8 @@
                 "ROS2"
             ],
             "compatible_engines": [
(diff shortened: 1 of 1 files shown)


spham-amzn (approved):

lumberyard-employee-dm (approved):

adamdbrw (approved):

michalpelka:
@AMZN-alexpete we do not support building o3de-extras/development against o3de/2305 (1.2.0).
I would like to set the minimum engine version for o3de-extras/stabilization/2301 and o3de-extras/development for 2.1.0 (current o3de/stabilization/2310).

AMZN-alexpete:
> @AMZN-alexpete we do not support building `o3de-extras/development` and `o3de-extras/stabilization/2310` against `o3de/2305 (1.2.0)`. I would like to set the minimum engine version for` o3de-extras/stabilization/2310` and `o3de-extras/development` for **2.1.0** (current `o3de/stabilization/2310`).

Good call, I've updated the minimum to be `2.1.0`

PR #540, opened 2023-09-26
Create asset Gem with OTTO robots

aniafafe:
Resolves: https://github.com/RobotecAI/ROSCon2023Demo/issues/123


Gems/OTTORobots/README.md
```diff
@@ -0,0 +1,17 @@
+# OTTO Robots Gem
+
+<!-- ![](docs/images/OttoRobots.png) -->
```
jhanca-robotecai, line 3: Please add the picture or remove the comment.

jhanca-robotecai (approved):
The gem builds correctly, and links with the project correctly. The prefabs look really good and work as expected. I did not check the script for animation, as it [crashes o3de](https://github.com/o3de/o3de/issues/16809).

Btw. it is always helpful to add a few words of description while creating a pull request; a short info on how it was tested and why it is there (in this case linking to the issue with the appropriate keyword to close the issue after PR is merged).

jhanca-robotecai (changes requested):
I checked the gem again while doing something else (I updated o3de in the meanwhile) and I noticed a hole in the place where ToF is typically located. I believe this should be fixed in this gem and in the original URDF.

My import results:
image

Picture found online
image

PR #544, opened 2023-09-27
Fix Mismatched Gem versioning for the ROS gem

spham-amzn:
The [recent update ](https://github.com/o3de/o3de-extras/pull/541)to fix the engine versioning for the ROS2 gem was not reflected in the ROS2 gem's `gem.json`.

This mismatch causes the following error:
```
[ERROR] o3de.download: SECURITY VIOLATION: Downloaded gem.json contents do not match the advertised manifest json contents.
[ERROR] o3de.download: Could not validate zip, deleting /home/o3de/.o3de/Cache/gems/ROS2/gem.zip
```

### DIFF
--- a/Gems/ROS2/gem.json
+++ b/Gems/ROS2/gem.json
@@ -18,8 +18,8 @@
         "ROS2"
     ],
     "compatible_engines": [
-        "o3de-sdk>=1.2.0",
-        "o3de>=1.2.0"
+        "o3de-sdk>=2.1.0",
+        "o3de>=2.1.0"
     ],
     "icon_path": "preview.png",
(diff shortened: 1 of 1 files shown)


cgalvan (approved):

AMZN-alexpete:
Hmm, I'm not sure why I didn't think of this earlier, but we will likely need to the .zip download for the gem with the correct gem.json and that will change the zip hash which means we will need to update the relevant repo.json files too.

lumberyard-employee-dm (approved):

AMZN-alexpete:
I will need to merge the changes from this PR to stab so we update the gem.json and the .zip and sha in repo.json https://github.com/o3de/o3de-extras/pull/545

AMZN-alexpete (approved):

PR #555, opened 2023-09-29
Convert <material> tags in SDF/URDF files

mbalfour-amzn:
This is the first pass at converting <material> tags in SDF/URDF files. It attempts to convert everything except <script>, <shader>, and <pbr> elements right now. There is some commented-out code that shows roughly how the PBR textures would be converted, but more work will be necessary to get the texture paths resolved, copied, and processed by the AP before handling it here. Hopefully, I'll get that working in a follow-up PR, but it's going to take some restructuring of how we detect and copy assets in the importer right now, so I wanted to keep those changes separate.

I also completely removed the MaterialMap. It was a leftover artifact from using urdfdom instead of sdfimport and was no longer needed or used.

Here are some before/after examples:

Demonstrates ambient, emissive, and specular conversions:
dcvviewer_8wiPXrkUrp
dcvviewer_6367PgyWfh

Demonstrates transparency and diffuse conversions:
dcvviewer_qOWPttMIRm
dcvviewer_MBoZLdB0hq

Demonstrates double-sided conversions (screenshot from inside the model):
dcvviewer_nyK4oqkp7i
dcvviewer_nV2NURfLEP


Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -49,38 +51,7 @@ namespace ROS2
...
-            auto VisitAllModels = [&GetVisualsFromModel](const sdf::Model& model, const Utils::ModelStack&) -> Utils::VisitModelResponse
-            {
-                GetVisualsFromModel(model);
-                // Continue to visit all models within the SDF document and query their <visual> tags
-                return Utils::VisitModelResponse::VisitNestedAndSiblings;
-            };
-            Utils::VisitModels(*m_root, VisitAllModels);
-
-            m_visualsMaker = VisualsMaker(AZStd::move(materialMap), urdfAssetsMapping);
+            m_visualsMaker = VisualsMaker(urdfAssetsMapping);
```
lumberyard-employee-dm, line 54: Actaully this entire if block can be removed.
We can pass in the urdfAssetMapping variable as part of the constructor initialization list for m_visualsMaker on line 43.

We had to make a new instance before to pass in the MaterialMap, but as we are no longer doing that, we can initialize it directly in the constructor.

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -225,36 +226,262 @@ namespace ROS2
...
-            AZStd::any(materialColor));
-        entity->Deactivate();
+        // First, force the model asset to get loaded into memory before adding the material component.
+        // This is required so that we can get the default material map that will be used to override properties for each material.
+        auto modelAsset = AZ::Data::AssetManager::Instance().GetAsset<AZ::RPI::ModelAsset>(assetId, AZ::Data::AssetLoadBehavior::Default);
+        modelAsset.BlockUntilLoadComplete();
+
+        // Initialize the material component configuration to contain all of the material mappings from the model.
+        AZ::Render::MaterialComponentConfig config;
+        config.m_materials = AZ::Render::GetDefautMaterialMapFromModelAsset(modelAsset);
```
lumberyard-employee-dm, line 471: @gadams3 Just fixed the name of thus function to be `GetDefaultMaterialMapFromModelAsset` and the old name is now deprecated.

lumberyard-employee-dm, line 471: We should check if the Model Asset is ready before proceding.

mbalfour-amzn: He fixed it in stabilization, so there's a timing issue where it isn't fixed in development yet.

mbalfour-amzn: I'm kinda mixed on that. Even if the model fails to load, we _could_ still continue and set the default materials without the slots to the override values, and create the material component. It would be reasonable to print an error if it failed to load though.

mbalfour-amzn: Added an error message.

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
mbalfour-amzn: Fixed!

lumberyard-employee-dm (approved):

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -171,10 +171,9 @@ namespace ROS2
                 const AZ::Vector3 scaleVector = URDF::TypeConversions::ConvertVector3(meshGeometry->Scale());
 
                 const auto asset = PrefabMakerUtils::GetAssetFromPath(*m_urdfAssetsMapping, AZStd::string(meshGeometry->Uri().c_str()));
-                AZ::Data::AssetId assetId;
+                assetId = Utils::GetModelProductAssetId(asset->m_sourceGuid);
```
adamdbrw, line 174: `asset` is accessed now before/regardless of it is valid (checked by `if (asset)`), not sure if this is intended.

adamdbrw (approved):

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
adamdbrw: Should we wait with merging this until the the aforementioned change is in o3de development? The alternative is to target stabilization2310 branch of o3de-extras now with this PR.

jhanca-robotecai: @adamdbrw the code builds with o3de `development` only due to changes in `SceneApi`, so targeting `stabilization/2310` in o3de-extras would be incorrect

mbalfour-amzn: The spelling change should get pulled into development today I think, so I'll wait to submit this until after it's got the correct spelling.

jhanca-robotecai (approved):
The code works as expected, i.e. the colors are correctly loaded. However, I had problems with building it. In particular, I had to add two includes to `VisualsMaker.cpp` :
```c++
#include <sdf/Material.hh>
#include <sdf/Pbr.hh>
```

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
mbalfour-amzn: Nope, not intended, fixed!

mbalfour-amzn:
> The code works as expected, i.e. the colors are correctly loaded. However, I had problems with building it. In particular, I had to add two includes to `VisualsMaker.cpp` :
> 
> ```c++
> #include <sdf/Material.hh>
> #include <sdf/Pbr.hh>
> ```

Ah, thanks! I build with Unity builds enabled so I didn't notice this. Added the missing includes.

michalpelka (approved):

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
mbalfour-amzn: Spelling change is in dev, so I've now changed it here and I'm running AR.

Issue #560, opened 2023-10-02
RobotImporter fails to import simple urdf


jhanca-robotecai:
RobotImporter fails to import a simple `urdf` file with two links and a fixed joint. It works correctly on `urdf` file with one link only. 
[test_models.zip](https://github.com/o3de/o3de-extras/files/12783750/test_models.zip)

It works correctly on 5185ef0.
Prefab creator stalls on empty screen on 6acba55.
![fail1](https://github.com/o3de/o3de-extras/assets/134940295/94ee89c6-8c97-44bb-a68f-3c294dfa53bd)

Prefab creator complains it cannot create a prefab on 1bba3b7.
![fail2](https://github.com/o3de/o3de-extras/assets/134940295/91849a89-fc03-4525-9b56-c4568a09561a)

Therefore the main suspects are #523 and #547

michalpelka:
I found some issue :
```
michalpelka@mpsimpc:~/github/o3de-extras$ git diff
diff --git a/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp b/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
index 87e5178f..3722cf5d 100644
--- a/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/RobotImporterWidget.cpp
@@ -439,10 +439,6 @@ namespace ROS2
         const AZ::IO::Path prefabPath(AZ::IO::Path(AZ::Utils::GetProjectPath()) / prefabPathRelative);
         bool fileExists = AZ::IO::FileIOBase::GetInstance()->Exists(prefabPath.c_str());
 
-        {
-            m_prefabMakerPage->setSuccess(false);
-            return;
-        }
         if (fileExists)
         {
             QMessageBox msgBox;
```
Now I am getting info on created entities, but still no luck with prefab:
```
[Error] (Prefab) - PrefabLoader::SaveTemplateToFile - Failed to save template 'Levels/www/www.prefab' to location '/home/michalpelka/github/ROSCon2023Demo/Project/Assets/Importer/004_husky.prefab'.Error: Relative path 'Assets/Importer/004_husky.prefab' for location didn't match template name.
```

michalpelka:
#523 gives empty entites list (as in first screenshot).
#547 fixing bug, allows prefab makere to continue, but gives error:
[...]

PR #561, opened 2023-10-02
SDF Nested Model Support Changes

lumberyard-employee-dm:
Added a `GetModelContainingModel` function that can return the parent model of a nested model

Updated the URDFPrefabMaker to create an Entity for which represents the URDF/SDF model itself.

The AddEntitiesForLink function has been updated to to pass in the model that the link is attached to and in order to allow the CollidersMaker, ArticulationsMaker and SensorsMaker have access to the Model

Updated the Check Asset Page to not add Assets with empty source paths or null Asset IDs to the list of resolved assets

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
@@ -105,7 +105,9 @@ namespace ROS2
         int i = m_table->rowCount();
         m_table->setRowCount(i + 1);
 
-        bool isOk = assetSourcePath.has_value() && resolvedSdfPath.has_value();
+        // The Asset ID GUID not be null(all zeros) and the asset source path must not be empty
+        bool isOk = (assetSourcePath.has_value() && !assetSourcePath->empty() && assetSourcePath != "not found")
+            && (resolvedSdfPath.has_value() && !resolvedSdfPath->empty()) && !assetUuid.IsNull();
         if (!isOk)
         {
             m_missingCount++;
@@ -154,8 +156,13 @@ namespace ROS2
             m_table->item(i, Columns::ResolvedMeshPath)->setIcon(m_failureIcon);
             m_table->setItem(i, Columns::ProductAsset, createCell(false, QString()));
         }
-        m_assetsPaths.push_back(assetSourcePath ? *assetSourcePath : AZStd::string());
-        m_assetsUuids.push_back(assetUuid);
+
+        // Don't add an empty asset path to the list of resolved assets
+        if (isOk)
+        {
+            m_assetsPaths.push_back(AZStd::move(*assetSourcePath));
+            m_assetsUuids.push_back(assetUuid);
+        }
     }
 
     void CheckAssetPage::StartWatchAsset()
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
@@ -105,7 +105,7 @@ namespace ROS2
         int i = m_table->rowCount();
         m_table->setRowCount(i + 1);
 
-        // The Asset ID GUID not be null(all zeros) and the asset source path must not be empty
+        // The Asset ID GUID must not be null(all zeros) and the asset source path must not be empty
         bool isOk = (assetSourcePath.has_value() && !assetSourcePath->empty() && assetSourcePath != "not found")
             && (resolvedSdfPath.has_value() && !resolvedSdfPath->empty()) && !assetUuid.IsNull();
         if (!isOk)
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
@@ -535,7 +535,7 @@ namespace ROS2
             return createEntityResult;
         }
 
-        // Get the model entity and update it's transform component with the pose information
+        // Get the model entity and update its transform component with the pose information
         AZ::EntityId entityId = createEntityResult.GetValue();
         AZStd::unique_ptr<AZ::Entity> entity(AzToolsFramework::GetEntityById(entityId));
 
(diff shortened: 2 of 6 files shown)


Gems/ROS2/Code/Source/RobotImporter/Pages/CheckAssetPage.cpp
```diff
@@ -96,30 +96,32 @@ namespace ROS2
...
         const AZStd::string assetSourcePath,
         const AZ::Crc32& crc32,
-        const AZStd::string resolvedSdfPath)
+        const AZStd::string resolvedAssetPath)
     {
         int i = m_table->rowCount();
         m_table->setRowCount(i + 1);
 
-        bool isOk = !assetSourcePath.empty();
+        // The Asset ID GUID not be null(all zeros) and the asset source path must not be empty
```
mbalfour-amzn, line 108:
```suggestion
        // The Asset ID GUID must not be null(all zeros) and the asset source path must not be empty
```

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -183,10 +184,43 @@ namespace ROS2
...
+                    if (AzToolsFramework::Prefab::PrefabEntityResult createModelEntityResult = CreateEntityForModel(*attachedModel);
+                        createModelEntityResult)
+                    {
+                        modelEntityId = createModelEntityResult.GetValue();
+                        // Add the model entity to the created entity list
+                        // so that it gets added to the prefab
+                        createdEntities.emplace_back(modelEntityId);
+
+                        std::string modelName = attachedModel->Name();
+                        AZStd::string azModelName(modelName.c_str(), modelName.size());
```
mbalfour-amzn, line 213:
```suggestion
                        AZStd::string azModelName(modelName.data(), modelName.size());
```

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -183,10 +184,43 @@ namespace ROS2
...
+                        AZStd::lock_guard<AZStd::mutex> lck(m_statusLock);
+                        m_status.emplace(azModelName, AZStd::string::format("created as: %s", modelEntityId.ToString().c_str()));
```
mbalfour-amzn, line 215: Huh, I'm surprised we never added a ToFixedString() for EntityId.

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
```diff
@@ -493,15 +527,91 @@ namespace ROS2
...
     {
-        if (!link)
+        auto createEntityResult = PrefabMakerUtils::CreateEntity(AZ::EntityId{}, model.Name().c_str());
+        if (!createEntityResult.IsSuccess())
         {
-            AZ::Failure(AZStd::string("Failed to create prefab entity - link is null"));
+            return createEntityResult;
+        }
+
+        // Get the model entity and update it's transform component with the pose information
```
mbalfour-amzn, line 538:
```suggestion
        // Get the model entity and update its transform component with the pose information
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -434,9 +434,10 @@ namespace ROS2::Utils
                 if (geometry->Type() == sdf::GeometryType::MESH)
                 {
                     auto pMesh = geometry->MeshShape();
-                    if (pMesh)
+                    std::string meshUri = pMesh->Uri();
+                    if (pMesh && !meshUri.empty())
                     {
-                        filenames.insert(AZStd::string(pMesh->Uri().c_str(), pMesh->Uri().size()));
+                        filenames.emplace(meshUri.c_str(), meshUri.size());
```
mbalfour-amzn, line 440:
```suggestion
                        filenames.emplace(meshUri.data(), meshUri.size());
```

mbalfour-amzn (approved):

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
lumberyard-employee-dm: There is no differences between `.data()` and `.c_str()` and neither one is better in this case, plus the C++ Best Practices Guide recommends using `.c_str()` over `.data()` when accessing the string internal buffer: https://github.com/o3de/sig-core/blob/main/governance/C%2B%2B-Best-Practices-Guide.md#string-buffer-access

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
lumberyard-employee-dm: I guess we never had a pass done over it.
Since it is a 64-bit value, it can fit within 20-digits, so an `AZStd::fixed_string<32>` would be the best data type for it, if we do add the overload.

Gems/ROS2/Code/Source/RobotImporter/URDF/URDFPrefabMaker.cpp
mbalfour-amzn: IMO the Best Practice doesn't apply in this case, since you're explicitly intentionally ignoring the null termination and creating a new string without a null terminator. By using c_str(), it's guaranteeing that a null-terminated string is getting generated if it doesn't already exist. data() allows you to create the new string without requiring the null terminator.

If you want to use c_str() and preserve the terminator, then shouldn't this use modelName.size() + 1?

lumberyard-employee-dm: `c_str()` and `data()` are literally the same implementation [per the C++ standard](https://en.cppreference.com/w/cpp/string/basic_string/data) post C++11
> The returned array is null-terminated, that is, data() and [c_str()](https://en.cppreference.com/w/cpp/string/basic_string/c_str) perform the same function.

mbalfour-amzn (approved):

michalpelka (commented):
LGTM, but due to #560 I cannot test it.
Edit: tested again, works as intended.

michalpelka (approved):

PR #566, opened 2023-10-04
Fully working SDF material conversions

mbalfour-amzn:
This PR finishes the work of supporting <material> conversions:
* <pbr> subsections in a <material> will now convert correctly.
* <script> subsections in a <material> will look for an asset of the same name (ex: "Gazebo/Wood" will look for "Gazebo/Wood.material") and use the asset reference if found.
* Most of the standard Gazebo materials have been added as material assets. These materials could use further refinement and textures (wood, grass, asphalt, etc), right now they are all solid color materials that are roughly similar to the equivalent standard Gazebo materials, but are neither exactly the same nor enhanced. I'm not an artist though, so hopefully someone else will come along and improve these. I tried to preserve diffuse color, emissive behavior, and transparency as best as possible when converting.

Some additional changes are in here as well:
* Added an "empty.sdf" file. This can be used with the URI prefix mapping system to replace an <include> file with an empty file. Useful for when you're trying to remove an <include> without modifying the source file.
* Removed BuildCollider() / BuildColliders(), they were dead code.
* Extended the asset copying and status tracking code to also handle any textures referenced by materials. These extensions should also make it much easier to add other asset references in the future, like heightmap images, geometry images, etc. The textures referenced by a <pbr> block will get copied over and reported on the asset status page.

(Converted https://github.com/azazdeaz/fields-ignition/tree/noetic/fields_ignition/generated_examples/tomato_field/tomato_0 )

Asset status page showing referenced textures that are copied:
image

SDF conversion using multiple <pbr> blocks to apply different textures to different parts of the tomato plant:
image
NOTE: This is also demonstrating some limitations in the conversion:
* <submesh> support doesn't exist yet, so multiple materials are trying to get applied to the entire mesh, leaving a somewhat random and chaotic result. 
* The material in the source SDF _ought_ to denote `<doubleSided>` but doesn't, so the leaves are invisible from one side. 
* There isn't a way for the source SDF to specify an opacity mode of Blended/Cutout, so the leaves ended up with visible opaque black fringes from the conversion which requires a manual edit to set the opacity mode to fix.

(Converted https://github.com/Daniella1/urdf_files_dataset/tree/main/urdf_files/ros-industrial/motoman/motoman_sia10d_support/urdf )

SDF conversion using "Gazebo/White" and "Gazebo/Blue" default materials:
image

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
@@ -64,147 +64,6 @@ namespace ROS2
         }
     }
 
-    void CollidersMaker::BuildColliders(const sdf::Link* link)
-    {
-        if (!link)
-        {
-            return;
-        }
-
-        for (uint64_t index = 0; index < link->CollisionCount(); index++)
-        {
-            BuildCollider(link->CollisionByIndex(index));
-        }
-    }
-
-    void CollidersMaker::BuildCollider(const sdf::Collision* collision)
-    {
-        if (!collision)
-        { // it is ok not to have collision in a link
-            return;
-        }
-
-        auto geometry = collision->Geom();
-        bool isPrimitiveShape = geometry->Type() != sdf::GeometryType::MESH;
-        if (!isPrimitiveShape)
-        {
-            auto meshGeometry = geometry->MeshShape();
-            if (!meshGeometry)
-            {
-                return;
-            }
-            const auto asset = PrefabMakerUtils::GetAssetFromPath(*m_urdfAssetsMapping, meshGeometry->Uri());
-            if (!asset)
-            {
-                return;
-            }
-            const auto& azMeshPath = asset->m_sourceAssetGlobalPath;
-
-            AZStd::shared_ptr<AZ::SceneAPI::Containers::Scene> scene;
-            AZ::SceneAPI::Events::SceneSerializationBus::BroadcastResult(
-                scene, &AZ::SceneAPI::Events::SceneSerialization::LoadScene, azMeshPath.c_str(), AZ::Uuid::CreateNull(), "");
-            if (!scene)
-            {
-                AZ_Error(
-                    Internal::CollidersMakerLoggingTag,
-                    false,
-                    "Error loading collider. Invalid scene: %s, URDF/SDF path: %s",
-                    azMeshPath.c_str(),
-                    meshGeometry->Uri().c_str());
-                return;
-            }
-
-            AZ::SceneAPI::Containers::SceneManifest& manifest = scene->GetManifest();
-            auto valueStorage = manifest.GetValueStorage();
-            if (valueStorage.empty())
-            {
-                AZ_Error(
-                    Internal::CollidersMakerLoggingTag, false, "Error loading collider. Invalid value storage: %s", azMeshPath.c_str());
-                return;
-            }
-
-            auto view = AZ::SceneAPI::Containers::MakeDerivedFilterView<AZ::SceneAPI::DataTypes::ISceneNodeGroup>(valueStorage);
-            if (view.empty())
-            {
-                AZ_Error(Internal::CollidersMakerLoggingTag, false, "Error loading collider. Invalid node views: %s", azMeshPath.c_str());
-                return;
-            }
-
-            // Select all nodes for both visual and collision nodes
-            for (AZ::SceneAPI::DataTypes::ISceneNodeGroup& mg : view)
-            {
-                AZ::SceneAPI::Utilities::SceneGraphSelector::SelectAll(scene->GetGraph(), mg.GetSceneNodeSelectionList());
-            }
-
-            // Update scene with all nodes selected
-            AZ::SceneAPI::Events::ProcessingResultCombiner result;
-            AZ::SceneAPI::Events::AssetImportRequestBus::BroadcastResult(
-                result,
-                &AZ::SceneAPI::Events::AssetImportRequest::UpdateManifest,
-                *scene,
-                AZ::SceneAPI::Events::AssetImportRequest::ManifestAction::Update,
-                AZ::SceneAPI::Events::AssetImportRequest::RequestingApplication::Editor);
-
-            if (result.GetResult() != AZ::SceneAPI::Events::ProcessingResult::Success)
-            {
-                AZ_Trace(Internal::CollidersMakerLoggingTag, "Scene updated\n");
-                return;
-            }
-
-            auto assetInfoFilePath = AZ::IO::Path{ azMeshPath };
-            assetInfoFilePath.Native() += ".assetinfo";
-            AZ_Printf(Internal::CollidersMakerLoggingTag, "Saving collider manifest to %s\n", assetInfoFilePath.c_str());
-            scene->GetManifest().SaveToFile(assetInfoFilePath.c_str());
-
-            // Set export method to convex mesh
-            auto readOutcome = AZ::JsonSerializationUtils::ReadJsonFile(assetInfoFilePath.c_str());
-            if (!readOutcome.IsSuccess())
-            {
-                AZ_Error(
-                    Internal::CollidersMakerLoggingTag,
-                    false,
-                    "Could not read %s with %s",
-                    assetInfoFilePath.c_str(),
-                    readOutcome.GetError().c_str());
-                return;
-            }
-            rapidjson::Document assetInfoJson = readOutcome.TakeValue();
-            auto manifestObject = assetInfoJson.GetObject();
-            auto valuesIterator = manifestObject.FindMember("values");
-            if (valuesIterator == manifestObject.MemberEnd())
-            {
-                AZ_Error(
-                    Internal::CollidersMakerLoggingTag, false, "Invalid json file: %s (Missing 'values' node)", assetInfoFilePath.c_str());
-                return;
-            }
-
-            constexpr AZStd::string_view physXMeshGroupType = "{5B03C8E6-8CEE-4DA0-A7FA-CD88689DD45B} MeshGroup";
-            auto valuesArray = valuesIterator->value.GetArray();
-            for (auto& value : valuesArray)
-            {
-                auto object = value.GetObject();
-
-                auto physXMeshGroupIterator = object.FindMember("$type");
-                if (AZ::StringFunc::Equal(physXMeshGroupIterator->value.GetString(), physXMeshGroupType))
-                {
-                    value.AddMember(rapidjson::StringRef("export method"), rapidjson::StringRef("1"), assetInfoJson.GetAllocator());
-                }
-            }
-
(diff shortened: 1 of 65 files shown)


Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -231,7 +232,54 @@ namespace ROS2
...
+        AZStd::string materialName(material->ScriptName().c_str(), material->ScriptName().size());
+        if (materialName.empty())
+        {
+            return;
+        }
+
+        // If a material has a <script> element we'll treat the name as a path and name to an O3DE material.
+        // For example, "Gazebo/Wood" will look for a product material in "<cache>/gazebo/wood.azmaterial"
+        AZ::IO::Path materialProductPath(materialName);
+        materialProductPath.ReplaceExtension(".azmaterial");
```
lumberyard-employee-dm, line 246: Shouldn't we lowercase the material name here?
```suggestion
        AZStd::string materialName(material->ScriptName().c_str(), material->ScriptName().size());
        if (materialName.empty())
        {
            return;
        }

        // Make sure the material name is lowercased before checking the path in the Asset Cache
        AZStd::to_lower(materialName);
        // If a material has a <script> element we'll treat the name as a path and name to an O3DE material.
        // For example, "Gazebo/Wood" will look for a product material in "<cache>/gazebo/wood.azmaterial"
        AZ::IO::Path materialProductPath(materialName);
        materialProductPath.ReplaceExtension(".azmaterial");
```

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -231,7 +232,54 @@ namespace ROS2
         entity->Deactivate();
     }
 
-    static void OverrideMaterialPbrSettings(const sdf::Material* material, [[maybe_unused]] AZ::Render::MaterialAssignmentMap& overrides)
+    static void OverrideScriptMaterial(const sdf::Material* material, [[maybe_unused]] AZ::Render::MaterialAssignmentMap& overrides)
```
lumberyard-employee-dm, line 235: The `overrides` parameter is no longer unused.
```suggestion
    static void OverrideScriptMaterial(const sdf::Material* material, AZ::Render::MaterialAssignmentMap& overrides)
```

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.h
```diff
@@ -76,37 +89,31 @@ namespace ROS2::Utils
...
+    //! @returns vector of relative paths to products, empty if no products are found
+    AZStd::vector<AZStd::string> GetProductAssets(const AZ::Uuid& sourceAssetUUID);
 
     //! Helper function that gives the desired product asset ID from source asset GUID
     //! @param sourceAssetUUID is source asset GUID
     //! @param typeId type of product asset
     //! @returns product asset id (invalid id if not found)
     AZ::Data::AssetId GetProductAssetId(const AZ::Uuid& sourceAssetUUID, const AZ::TypeId typeId);
 
+    //! Helper function that gives AZ::RPI::ImageAsset product asset from source asset GUID
```
lumberyard-employee-dm, line 112: This comment should be updated now that a AssetID is returned.

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
mbalfour-amzn: It's not _required_, since the asset cache lookup with also to_lower it, but i'm great with making it explicit here.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -364,36 +373,45 @@ namespace ROS2::Utils
...
-                    // create asset info at destination location using the temporary mesh file
-                    const bool assetInfoOk = CreateSceneManifest(targetPathAssetTmp, targetPathAssetInfo, needsCollider, needsVisual);
+                    const bool needsVisual = (assetReferenceType & ReferencedAssetType::VisualMesh) == ReferencedAssetType::VisualMesh;
+                    const bool needsCollider = (assetReferenceType & ReferencedAssetType::ColliderMesh) == ReferencedAssetType::ColliderMesh;
+                    const bool isMeshFile = (needsVisual || needsCollider);
+
+                    // if the asset is a mesh, create asset info at destination location using the temporary mesh file
+                    const bool assetInfoOk = isMeshFile
+                        ? CreateSceneManifest(targetPathAssetTmp, targetPathAssetInfo, needsCollider, needsVisual)
+                        : true;
```
lumberyard-employee-dm, line 383: A ternary statement isn't needed here.
A logical-or can be used.
```suggestion
                    // if the asset is a mesh, create asset info at destination location using the temporary mesh file
                    const bool assetInfoOk = !isMeshFile ||
                        CreateSceneManifest(targetPathAssetTmp, targetPathAssetInfo, needsCollider, needsVisual);
```

mbalfour-amzn: Personally I find the ternary more straightforward to read, I'd prefer to keep it that way.

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
```diff
@@ -424,39 +424,96 @@ namespace ROS2::Utils
         VisitModelsForNestedModels(sdfRoot);
     }
 
-    AZStd::unordered_set<AZStd::string> GetMeshesFilenames(const sdf::Root& root, bool visual, bool colliders)
+    AssetFilenameReferences GetReferencedAssetFilenames(const sdf::Root& root)
     {
-        AZStd::unordered_set<AZStd::string> filenames;
-        auto GetMeshesFromModel = [&filenames, visual, colliders](const sdf::Model& model, const ModelStack&) -> VisitModelResponse
+        AssetFilenameReferences filenames;
+        auto GetMeshesFromModel = [&filenames](const sdf::Model& model, const ModelStack&) -> VisitModelResponse
```
lumberyard-employee-dm, line 430: I guess we can rename this to "GetAssetsFromModel"?

mbalfour-amzn: Ooops, missed that one, I'll fix.

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -231,7 +232,56 @@ namespace ROS2
...
+            {
+                override.m_defaultMaterialAsset = materialAsset;
+            }
+            else
+            {
+                override.m_materialAsset = materialAsset;
+            }
+        }
+
+        AZ_Info("AddMaterial", "Added product material %s", materialProductPath.String().c_str());
```
lumberyard-employee-dm, line 281: This can use `.c_str()`.
Also we probably want a newline between info messages as `AZ_Info` will not add newlines when the content is written out to the log file.
```suggestion
        AZ_Info("AddMaterial", "Added product material %s\n", materialProductPath.c_str());
```

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
```diff
@@ -231,7 +232,56 @@ namespace ROS2
...
+            assetId,
+            &AZ::Data::AssetCatalogRequestBus::Events::GetAssetIdByPath,
+            materialProductPath.String().c_str(),
+            azrtti_typeid<AZ::RPI::MaterialAsset>(),
+            AutoGenerateId);
+
+        // No asset was found, we can't convert the material script.
+        if (!assetId.IsValid())
+        {
+            AZ_Warning("AddMaterial", false, "Failed to find product material for %s", materialProductPath.String().c_str());
```
lumberyard-employee-dm, line 263: Calling `.String()` is not needed, since the `AZ::IO::Path` contains an `AZStd::string` internally and therefore doesn't need to make a  copy
```suggestion
            AZ_Warning("AddMaterial", false, "Failed to find product material for %s", materialProductPath.c_str());
```

Gems/ROS2/Code/Source/RobotImporter/Utils/RobotImporterUtils.cpp
mbalfour-amzn: Fixed!

lumberyard-employee-dm (approved):

Gems/ROS2/Code/Source/RobotImporter/URDF/VisualsMaker.cpp
mbalfour-amzn: Wait, that's not true. AZ_Info adds newlines.
https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Debug/Trace.h#L325-L329
https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Debug/Trace.h#L114

lumberyard-employee-dm: Yeah, that is not the Trace function that runs when running an application like the Editor or Launcher.
That is the interface class that only runs before an `AZ::Debug::Trace` instance is created.

There real function is the [Trace::Printf](https://github.com/o3de/o3de/blob/development/Code/Framework/AzCore/AzCore/Debug/Trace.cpp#L536-L559) function that runs.
That does not add a newline.

lumberyard-employee-dm: Those Trace functions are only used in the O3DE Kernel to log messages before main has run
https://github.com/o3de/o3de/commit/5cec7f916c95e00123d59e145de68b4c25fbbe6e

mbalfour-amzn: Agh, ty. I put back the newline.

Gems/ROS2/Code/Source/RobotImporter/URDF/CollidersMaker.cpp
```diff
@@ -64,147 +64,6 @@ namespace ROS2
         }
     }
 
-    void CollidersMaker::BuildColliders(const sdf::Link* link)
```
michalpelka, line 67: For other reviewers, the removed code a was dead code. Now creating colliders' product assets is done in "    CreateSceneManifest(sourceAssetPath,  assetInfoFile,  collider, visual)"

michalpelka (approved):
Tested for regression in urdf, tested SDF pointed by Mike.
Works as intended.

adamdbrw (approved):

PR #587, opened 2023-10-18
RobotImporter: lidar name fix

jhanca-robotecai:
After importing lidar, the name is not set correctly. This is a WIP (a working quickfix) for the issue. It will be further investigated if the change of the private field is required.

Tested on: 
[model_lidar.zip](https://github.com/o3de/o3de-extras/files/13031619/model_lidar.zip)


jhanca-robotecai:
This issue was fixed in #590, this quickfix should not be merged.

PR #599, opened 2023-11-06
Simple API to listen for ros2 system component state change

MateuszWasilewski:
This PR allows SystemComponents in Editor to get ros2Node without actively querying ROS2SystemComponent for its state.

Decouples ROS2 SystemComponent state from assumptions based on external events. 
Useful when creating components that publish data originating from other system components.

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/ROS2Bus.h
+++ b/Gems/ROS2/Code/Include/ROS2/ROS2Bus.h
@@ -8,10 +8,11 @@
 #pragma once
 
 #include <AzCore/EBus/EBus.h>
+#include <AzCore/EBus/Event.h>
 #include <AzCore/Interface/Interface.h>
+#include <ROS2/Clock/SimulationClock.h>
 #include <builtin_interfaces/msg/time.hpp>
 #include <geometry_msgs/msg/transform_stamped.hpp>
-#include <ROS2/Clock/SimulationClock.h>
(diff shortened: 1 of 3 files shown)


Gems/ROS2/Code/Include/ROS2/ROS2Bus.h
```diff
@@ -33,6 +34,12 @@ namespace ROS2
         //! @note Alternatively, you can use your own node along with an executor.
         virtual std::shared_ptr<rclcpp::Node> GetNode() const = 0;
 
+        //! Attach handler to ROS2 node change events.
+        //! You can use this method to correctly initialize SystemComponents that depend on ROS2Node.
+        //! @note callback is active as long as handler is not destroyed.
+        using NodeChanged = AZ::Event<std::shared_ptr<rclcpp::Node>>;
```
PawelLiberadzki, line 40: You can consider placing alias at the top of class definition - this is easier to read than placing it between doc and method signature.
+
@param is missing in documentation for this method.

PawelLiberadzki: I would optionally consider adding "...Event" to this alias name, if current form is not the part of convention in O3DE.

adamdbrw (approved):

michalpelka (approved):

PR #608, opened 2023-11-10
OpenXR Fixes

galibzon:
This PR improves XrSpaces location calculation.

    Previously, XrSpaces were being located each frame
    during OpenXRVkInput::PollActions() which was called
    before OpenXRVk::Device::BeginInternal(). The problem
    with that approach is that xrLocateSpace needs
     a "predicted display time" for the CURRENT frame.
    But the "predicted display time" is calculated during
      OpenXRVk::Device::BeginInternal(), this means that all XrPoses
      for the CURRENT frame were calculated with the "predicted
    display time" of the PREVIOUS frame. This was causing the rendered
    frames to look jittery.

    In this PR, the XrSpaces are now located during
     OpenXRVk::Device::BeginInternal() this guarantees that
     when xrLocateSpace is called it is using the correct
     "predicted display time" for the CURRENT frame. This improves
     the jitter because the camera View Srg transforms now feed the correct
     transform value for each frame improving the stability of the rendered
     views.

    Another improvement is that the Base Space used in xrLocateSpace
    is the LOCAL space instead of the hard coded VIEW space. Also APIS were
    added to change the Base Space for both Visualized spaces and Controller (Joysticks)
    spaces.

    Also the engine is now notified via AZ::RPI::XRSpaceNotificationBus::OnXRSpaceLocationsChanged
    at the right time so the engine can update the ViewSrg for each eye
    at the correct time.

    Another improvement is that the AZ::RPI::PoseData reported to the engine
    is now given according to the O3DE convention: X+ Right, Y+ forward, Z+ Up.
    Also added equivalent APIs that report the data as AZ::Transform.

    Fixed bug when the Xr Poses would be corrupted
    if the Proximity Sensor was enabled.

    Now the XrSpaces are reset each time a XR_SESSION_STATE_READY
    event is received (typically because the Proximity Sensor is disabled)

### DIFF
--- a/Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkSession.h
+++ b/Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkSession.h
@@ -40,6 +40,21 @@ namespace OpenXRVk
         //! Return the Xrspace related to the SpaceType enum
         XrSpace GetXrSpace(SpaceType spaceType) const;
 
+        ////////////////////////////////////////////////////////////////////////////////////////////
+        //! Called by a Device when the predicted display time has been updated (typically
+        //! the device updates the predicted display time during BeginFrame).
+        //! See OpenXRVkInput.h UpdateXrSpaceLocations(...) for more details.
+        void UpdateXrSpaceLocations(const OpenXRVk::Device& device, XrTime predictedDisplayTime, AZStd::vector<XrView>& xrViews);
+
+        //! Setters and Getters for the base spaces that will be used
+        //! when calling xrLocateSpace().
+        //! By default, the base space for visualization is SpaceType::Local
+        //! and the base space for Joysticks/controllers is SpaceType::View (aka the Head)
+        void SetBaseSpaceTypeForVisualization(SpaceType spaceType);
+        void SetBaseSpaceTypeForControllers(SpaceType spaceType);
+        SpaceType GetBaseSpaceTypeForVisualization() const;
+        SpaceType GetBaseSpaceTypeForControllers() const;
+
         //////////////////////////////////////////////////////////////////////////
         // XR::Session overrides
         AZ::RHI::ResultCode InitInternal(AZ::RHI::XRSessionDescriptor* descriptor) override;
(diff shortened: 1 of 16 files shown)


Gems/OpenXRVk/Code/Source/OpenXRVkInput.cpp
```diff
@@ -345,51 +371,78 @@ namespace OpenXRVk
...
+            const auto headToLeftEyeTm = AzTransformFromXrPose(xrViews[0].pose);
+            const auto headToRightEyeTm = AzTransformFromXrPose(xrViews[1].pose);
+
+            AZ::RPI::XRSpaceNotificationBus::Broadcast(&AZ::RPI::XRSpaceNotifications::OnXRSpaceLocationsChanged,
+                baseToHeadTm, headToLeftEyeTm, headToRightEyeTm);
+        }
 
-        rawControllerData.m_leftPositionState = convertVector3(m_handSpaceLocation[static_cast<AZ::u32>(XR::Side::Left)].pose.position);
-        rawControllerData.m_rightPositionState = convertVector3(m_handSpaceLocation[static_cast<AZ::u32>(XR::Side::Right)].pose.position);
+        return false;
```
AMZN-alexpete, line 407: It looks like every return path is false - what does the return value mean?

AMZN-alexpete (approved):
Just some minor comments - could you also bump the gem.json version minor version unless there are breaking changes in which case please increment the major version.

Gems/OpenXRVk/Code/Source/OpenXRVkInput.cpp
galibzon: Nice catch. Was missing a return true;. Fixed. Also  bumped the minor version of the OpenXRVK Gem. The new version is 1.0.1.

Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkSession.h
```diff
@@ -68,12 +84,25 @@ namespace OpenXRVk
...
+        // when the proximity sensor is ON.
+        void ResetSpaces();
 
         XrSession m_session = XR_NULL_HANDLE;
         XrSessionState m_sessionState = XR_SESSION_STATE_UNKNOWN;
         XrEventDataBuffer m_eventDataBuffer;
         XrInstance m_xrInstance = XR_NULL_HANDLE;
         XrGraphicsBindingVulkan2KHR m_graphicsBinding{ XR_TYPE_GRAPHICS_BINDING_VULKAN_KHR };
+        
+        // Application defined base space that will used to calculate
```
moudgils, line 97: that will **be** used

Gems/OpenXRVk/Code/Include/OpenXRVk/OpenXRVkSession.h
```diff
@@ -68,12 +84,25 @@ namespace OpenXRVk
...
+        // the relative pose of all other spaces.
+        // Typically SpaceType::Local or SpaceType::Stage.
+        SpaceType m_baseSpaceTypeForVisualization = SpaceType::Local;
+
+        // Application defined base space that will use to calculate
```
moudgils, line 102: will **be used** to

moudgils (approved):
Nice Work

ds5678:
> bump the gem.json version minor version

From what I can tell, this didn't get released.

PR #611, opened 2023-11-15
Defer response in spawn entity service

### LINKED ISSUE #142: Validate if the robot spawn was successful
ROS2Spawner responses with success without checking if the spawn was actually successful. It would be profitable to check it in some way. Probably using the m_completionCallback.

arturkamieniecki:
Resolves #142.

I've changed the spawn_entity service such that the response after adding the entity to the spawn queue is not imminently sent. Now the completion callback sends the response after spawning the entity.

### DIFF
--- a/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -40,9 +40,12 @@ namespace ROS2
 
         m_spawnService = ros2Node->create_service<gazebo_msgs::srv::SpawnEntity>(
             "spawn_entity",
-            [this](const SpawnEntityRequest request, SpawnEntityResponse response)
+            [this](
+                const SpawnEntityServiceHandle service_handle,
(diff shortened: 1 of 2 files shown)


Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -146,9 +154,14 @@ namespace ROS2
             PreSpawn(id, view, transform, spawnableName, spawnableNamespace);
         };
 
-        spawner->SpawnAllEntities(m_tickets.at(spawnableName), optionalArgs);
+        optionalArgs.m_completionCallback = [service_handle, header](auto id, auto view)
```
adamdbrw, line 157: Is this guaranteed to be called in each possible outcome of spawning? Is there an option to add a timeout?

adamdbrw (commented):
Good change, please take a look at my comment and investigate / answer

adamdbrw (approved):

michalpelka (approved):

PR #613, opened 2023-11-17
Spawnable validation

### LINKED ISSUE #136: Validate spawner's spawnable field
It should be checked if what was passed to the spawnable field of the spawner is really a spawnable.

arturkamieniecki:
Depends on #611.
Resolves #136.

I've added a check before the spawning of the Entity. This checks if the asset was loaded without error and returns a negative response when the error is detected.

Additionally in the o3de/development branch, the new DPE feature shows when a spawnable is improperly loaded.
Screenshot from 2023-11-17 11-49-55
Screenshot from 2023-11-17 11-50-04
There is no possibility to input a non spawnable in the above menu as the field is `AZ::Data::Asset<AzFramework::Spawnable>`.

### DIFF
--- a/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -136,8 +136,7 @@ namespace ROS2
 
         if (spawnable->second->IsLoading())
         {
-            // This check is done during the simulation. All assets in the build of the simulation are loaded and processed.
-            // This will only run when the simulation is started before all assets are processed.
+            // This is an Editor only situation. All assets during game mode are fully loaded.
             response.success = false;
             response.status_message = "Asset for spawnable " + request->name + " has not yet loaded.";
             service_handle->send_response(*header, response);
@@ -147,7 +146,7 @@ namespace ROS2
         if (spawnable->second->IsError())
         {
             response.success = false;
-            response.status_message = "Spawnable " + request->name + " loaded with an error";
+            response.status_message = "Spawnable " + request->name + " loaded with an error.";
             service_handle->send_response(*header, response);
             return;
         }
--- a/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -147,7 +147,7 @@ namespace ROS2
         if (spawnable->second->IsError())
         {
             response.success = false;
-            response.status_message = "Spawnable loaded with an error";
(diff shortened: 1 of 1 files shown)


michalpelka (approved):

Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -120,6 +120,25 @@ namespace ROS2
...
+            response.success = false;
+            response.status_message = "Asset for spawnable " + request->name + " has not yet loaded.";
+            service_handle->send_response(*header, response);
+            return;
+        }
+
+        if (spawnable->IsError())
+        {
+            response.success = false;
+            response.status_message = "Spawnable loaded with an error";
```
jhanca-robotecai, line 137: Nit: it could be an added value to add `request->name` in status message asabove

jhanca-robotecai (approved):
Approve

michalpelka:
Please rebase.

Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -120,6 +120,25 @@ namespace ROS2
...
+            response.success = false;
+            response.status_message = "Asset for spawnable " + request->name + " has not yet loaded.";
+            service_handle->send_response(*header, response);
+            return;
+        }
+
+        if (spawnable->IsError())
+        {
+            response.success = false;
+            response.status_message = "Spawnable " + request->name + " loaded with an error";
```
adamdbrw, line 137: Is it possible to get and add the error message to response here?

Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -120,6 +120,25 @@ namespace ROS2
             return;
         }
 
+        auto& spawnable = m_controller.GetSpawnables().find(spawnableName)->second;
+        if (spawnable->IsLoading())
+        {
+            // This check is done during the simulation. All assets in the build of the simulation are loaded and processed.
```
adamdbrw, line 126: This is Editor-only situation, correct? Please rephrase the comment for clarity.

adamdbrw (commented):
Good added checks and giving the caller a feedback!

Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
arturkamieniecki: It is possible to get the whole log of the build process. This log would be long and would decrease message readability.

adamdbrw (approved):

PR #616, opened 2023-11-21
FlushIO after copying files in Robot Importer

### LINKED ISSUE #607: Bug: does not import existing meshes
**Describe the bug**
Robot importer cannot find existing meshes once in a while. When the same import is repeated, it finds it correctly.

**Assets required**
Tested with [ANYmal C urdf file](https://github.com/ANYbotics/anymal_c_simple_description).

**Steps to reproduce**
1. Open Robot importer
2. Select file
3. Unmark `Preserve URDF fixed joints` option.
4. Next, next
5. See error - mesh listed as non-existing does exist

**Expected behavior**
All meshes should be detected correctly.

**Actual behavior**
Mesh listed as non-existing

**Screenshots/Video**
Screenshot from 2023-11-10 12-24-56

after repeating the operation, mesh was found:

Screenshot from 2023-11-10 12-26-57

and the model was fixed:
Screenshot from 2023-11-10 12-27-12

**Found in Branch**
`development`

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) and [o3de/o3de-extras](https://github.com/o3de/o3de-extras) repositories**
o3de: [o3de:point-release/23101](https://github.com/o3de/o3de/tree/point-release/23101) commit cfa481a235
o3de-extras: 707f4314

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Ubuntu

 - Version 22.04
 - CPU 12700H
 - GPU RTX 3070 Mobile
 - Memory 32 GB

**Additional context**
This error was already fixed https://github.com/o3de/o3de-extras/pull/400 and https://github.com/o3de/o3de-extras/pull/469

arturkamieniecki:
Resolves #607.

I've added a flushIOOfAsset function that flushes the assets to make the copied files available to the Asset Processor before further processing. 

This addition slows down the importing process. After clicking next in the page before CheckAssetsPage the user needs to wait and the window does not reflect that anything is being done in the background (the files are being copied). I've proposed a UX change in #617.

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
@@ -431,7 +431,7 @@ namespace ROS2::Utils
                             targetPathAssetDst.c_str(),
                             outcomeMoveDst.GetResultCode());
 
-                        // call FlushIOOfAsset to ensure the asset processor is aware of the new file
+                        // call FlushIOOfAsset to ensure the asset processor is aware of the move operation.
                         FlushIOOfAsset(targetPathAssetDst);
 
                         if (outcomeMoveDst)
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
@@ -374,6 +374,14 @@ namespace ROS2::Utils
 
                 if (outcomeCopyTmp)
                 {
+                    // An I/O flush is required here because we need to load the Scene file into memory from the temporary directory
+                    // to generate a proper manifest for it in the final directory before the Scene file itself is moved into the final
+                    // directory and processed. This ensures that when the Scene file is detected by the Asset Processor in the final
+                    // location, it will already have the desired export settings in the manifest to cause it to be exported correctly.
+                    // If we didn't flush the I/O here, the Scene file load can fail because the load contains a call to 
+                    // AssetSystemComponent::GetSourceInfoBySourcePath that will fail if the Asset Processor hasn't detected the
+                    // existence of the file yet. With the flush, everything works correctly.
+
                     FlushIOOfAsset(targetPathAssetTmp);
 
                     const bool needsVisual = (assetReferenceType & ReferencedAssetType::VisualMesh) == ReferencedAssetType::VisualMesh;
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
@@ -371,8 +371,11 @@ namespace ROS2::Utils
                     resolvedPath.c_str(),
                     targetPathAssetTmp.c_str(),
                     outcomeCopyTmp.GetResultCode());
+
                 if (outcomeCopyTmp)
                 {
+                    FlushIOOfAsset(targetPathAssetTmp);
+
                     const bool needsVisual = (assetReferenceType & ReferencedAssetType::VisualMesh) == ReferencedAssetType::VisualMesh;
                     const bool needsCollider = (assetReferenceType & ReferencedAssetType::ColliderMesh) == ReferencedAssetType::ColliderMesh;
                     const bool isMeshFile = (needsVisual || needsCollider);
@@ -420,18 +423,8 @@ namespace ROS2::Utils
                             targetPathAssetDst.c_str(),
                             outcomeMoveDst.GetResultCode());
 
-                        // call GetAssetStatus_FlushIO to ensure the asset processor is aware of the new file
-                        AzFramework::AssetSystem::AssetStatus copiedAssetStatus =
-                            AzFramework::AssetSystem::AssetStatus::AssetStatus_Unknown;
-                        AzFramework::AssetSystemRequestBus::BroadcastResult(
-                            copiedAssetStatus,
-                            &AzFramework::AssetSystem::AssetSystemRequests::GetAssetStatus_FlushIO,
-                            targetPathAssetDst.c_str());
-                        AZ_Warning(
-                            "CopyAssetForURDF",
-                            copiedAssetStatus != AzFramework::AssetSystem::AssetStatus::AssetStatus_Unknown,
-                            "Asset processor did not recognize the new file %s.",
-                            targetPathAssetDst.c_str());
+                        // call FlushIOOfAsset to ensure the asset processor is aware of the new file
+                        FlushIOOfAsset(targetPathAssetDst);
 
                         if (outcomeMoveDst)
                         {
@@ -660,5 +653,19 @@ namespace ROS2::Utils
             }
         }
         return assetsFilepaths;
+    }
+
+    AzFramework::AssetSystem::AssetStatus FlushIOOfAsset(const AZ::IO::Path& path)
+    {
+        AzFramework::AssetSystem::AssetStatus assetStatus = AzFramework::AssetSystem::AssetStatus::AssetStatus_Unknown;
+        AzFramework::AssetSystemRequestBus::BroadcastResult(
+            assetStatus, &AzFramework::AssetSystem::AssetSystemRequests::GetAssetStatus_FlushIO, path.c_str());
+        AZ_Warning(
+            "CopyAssetForURDF",
+            assetStatus != AzFramework::AssetSystem::AssetStatus::AssetStatus_Unknown,
+            "Asset processor did not recognize the new file %s.",
+            path.c_str());
+
+        return assetStatus;
     }
 } // namespace ROS2::Utils
--- a/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.h
+++ b/Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.h
@@ -17,6 +17,7 @@
 #include <AzCore/Math/Crc.h>
 #include <AzCore/std/containers/unordered_map.h>
 #include <AzCore/std/containers/unordered_set.h>
+#include <AzFramework/Asset/AssetSystemBus.h>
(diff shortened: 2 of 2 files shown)


jhanca-robotecai (approved):
Which o3de branch/commit did you use to recreate the issue? My problem is that I was getting #607 very often while importing robots on `main` 0d490a2b4b and `point-release`  cfa481a235 but I have problems recreating the issue on the current `point-release` branch 85b67a2edd. 

The code looks correct though, but I am not sure if still needed.

@mbalfour-amzn would you be so kind to verify it?

arturkamieniecki:
I've recreated this issue on the development branch of the engine. I'll check further if this issue still persists.

arturkamieniecki:
Tested on point-release `85b67a2edd64fcee53d6caf56ea800c6c2e7c01e `. The issue still persists.

arturkamieniecki:
The error is visible in the AssetProcessor logs.
Screenshot from 2023-11-24 13-07-48

mbalfour-amzn:
The logic of the change doesn't quite make sense to me here. The change looks like it's telling the AP to flush I/O for a file that's been copied to a temporary location that's ignored by the AP, and then copies it a second time to a location processed by the AP, where it's flushed again. I don't know why the newly-added flush would affect the AP if it's happening in a directory the AP is ignoring.

I suspect this is probably accidentally working around the problem by slowing things down enough that the real race condition isn't getting hit.

mbalfour-amzn:
In tracking down the cause of the error, is it clear on which line the failure point is occurring? For example, is the problem that the call to GetAssetStatus in here is returning invalid results?

The one kind of wild theory I've got for why the extra flush _might_ work is that since we're copying to a location then renaming to the final destination, maybe flush doesn't work correctly on a renamed file and it needs to be done on the original copy? That seems pretty weird though, I don't have a plausible reason why that would be the case.

lumberyard-employee-dm:
> The issue is caused by the early return of `AzToolsFramework::AssetSystemRequestBus::Events::GetSourceInfoBySourcePath` during `SceneSerializationHandler::LoadScene` for the file in the Temp directory. `GetSourceInfoBySourcePath` under the hood calls `AssetSystemComponent::GetSourceInfoBySourcePath` that produces a request to AssetProcessor.
> 
> I do not exactly understand how the Asset Processor process is structured, hence I do not have an entry point for the above-mentioned request form Editor. However, I found that it finally fails in `FileStateCache::Exists` on that file in question is missing from `m_fileInfoMap`. So the whole ordeal finishes because `FileStateCache::m_fileInfoMap` does not contain an entry for our file.
> 
> From my small investigation (with a number of prints and patience to reproduce) I can see two things:
> 
>     * `FileStateCache::m_fileInfoMap` contains the files in the temp directory e.g., `/ROSCon2023Demo/Project/Assets/UrdfImporter/$tmp_698984667.tmp/remote.dae`
> 
>     * For the situation of missing file like in @arturkamieniecki screenshots, this file is missing from `FileStateCache::m_fileInfoMap`.
> 
>     * It is a race between `FileStateCache::AddOrUpdateFileInternal` and `FileStateCache::Exists` for the file in question.
> 
> 
> Another thing that I found in asset processor's logs that can be related, but will not be fixed by this PR is :
> 
> ```
> 11/27/23 5:29 PM - FileWatcher - inotify_add_watch failed for path /home/michalpelka/github/ROSCon2023Demo/Project/Assets/UrdfImporter/$tmp_698984667.tmp with error 28: No space left on device (try increasing fs.inotify.max_user_watches with sysctl)
> ```

The `inotify_add_watch` is an issue with running out of filesystem watches to detect when a file in a folder has been modified.

I know the default limit on some linux kernels are 8192 inotify watches and we have added error messages to indicate when that cap has been hit: https://github.com/o3de/o3de/commit/8fa5dc21e29da8844c35f865341ed42dc7afb3a0

michalpelka:
Thanks @lumberyard-employee-dm  -  issue `inotify_add_watch` is a known issue with the environment. 
I at this moment I am sure that `FlushIO` causes to `FileStateCache::AddOrUpdateFileInternal` to move before  `FileStateCache::Exists`.

michalpelka (approved):

michalpelka:
I've accepted the change, but I would be more than happy if @mbalfour-amzn could take a look into my investigation and give his opinion :+1: .

mbalfour-amzn:
> I've accepted the change, but I would be more than happy if @mbalfour-amzn could take a look into my investigation and give his opinion 👍 .

Thanks for the detailed explanation! It makes perfect sense to me now, it's just really unfortunate. Longer term, it would be nice to find a way to generate the manifest without requiring the LoadScene call in the first place, since that can be a pretty expensive operation to do even without also requiring the flush here. It's also possible that we need to expose some alternate scene loading APIs that would allow it to be loaded from anywhere, not just from somewhere the AP knows about.

For now, my suggestion would be to put a VERY detailed comment above that first flush containing the results of the investigation so that all of this context is preserved. Maybe something like this:
```
// An I/O flush is required here because we need to load the Scene file into memory from the temporary directory 
// to generate a proper manifest for it in the final directory before the Scene file itself is moved into the final
// directory and processed. This ensures that when the Scene file is detected by the Asset Processor in the final
// location, it will already have the desired export settings in the manifest to cause it to be exported correctly.
// If we didn't flush the I/O here, the Scene file load can fail because the load contains a call to 
// AssetSystemComponent::GetSourceInfoBySourcePath that will fail if the Asset Processor hasn't detected the
// existence of the file yet. With the flush, everything works correctly.
```

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -420,18 +423,8 @@ namespace ROS2::Utils
...
-                        AzFramework::AssetSystemRequestBus::BroadcastResult(
-                            copiedAssetStatus,
-                            &AzFramework::AssetSystem::AssetSystemRequests::GetAssetStatus_FlushIO,
-                            targetPathAssetDst.c_str());
-                        AZ_Warning(
-                            "CopyAssetForURDF",
-                            copiedAssetStatus != AzFramework::AssetSystem::AssetStatus::AssetStatus_Unknown,
-                            "Asset processor did not recognize the new file %s.",
-                            targetPathAssetDst.c_str());
+                        // call FlushIOOfAsset to ensure the asset processor is aware of the new file
```
michalpelka, line 426:
```suggestion
                        // An I/O flush is required here because we need to load the Scene file into memory from the temporary directory 
                        // to generate a proper manifest for it in the final directory before the Scene file itself is moved into the final
                        // directory and processed. This ensures that when the Scene file is detected by the Asset Processor in the final
                        // location, it will already have the desired export settings in the manifest to cause it to be exported correctly.
                        // If we didn't flush the I/O here, the Scene file load can fail because the load contains a call to 
                        // AssetSystemComponent::GetSourceInfoBySourcePath that will fail if the Asset Processor hasn't detected the
                        // existence of the file yet. With the flush, everything works correctly.

```

michalpelka, line 426:
```suggestion
                      // An I/O flush is required here because we need to load the Scene file into memory from the temporary directory 
                        // to generate a proper manifest for it in the final directory before the Scene file itself is moved into the final
                        // directory and processed. This ensures that when the Scene file is detected by the Asset Processor in the final
                        // location, it will already have the desired export settings in the manifest to cause it to be exported correctly.
                        // If we didn't flush the I/O here, the Scene file load can fail because the load contains a call to 
                        // AssetSystemComponent::GetSourceInfoBySourcePath that will fail if the Asset Processor hasn't detected the
                        // existence of the file yet. With the flush, everything works correctly.

```

mbalfour-amzn: I'd actually suggest that this comment get added above the call to FlushIOOfAsset on line 377 as opposed to here, since that's the one that's needed for the LoadScene() call.

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -371,8 +371,11 @@ namespace ROS2::Utils
                     resolvedPath.c_str(),
                     targetPathAssetTmp.c_str(),
                     outcomeCopyTmp.GetResultCode());
+
                 if (outcomeCopyTmp)
                 {
+                    FlushIOOfAsset(targetPathAssetTmp);
```
michalpelka, line 377:
```suggestion
                    // An I/O flush is required here because we need to load the Scene file into memory from the temporary directory
                    // to generate a proper manifest for it in the final directory before the Scene file itself is moved into the final
                    // directory and processed. This ensures that when the Scene file is detected by the Asset Processor in the final
                    // location, it will already have the desired export settings in the manifest to cause it to be exported correctly.
                    // If we didn't flush the I/O here, the Scene file load can fail because the load contains a call to 
                    // AssetSystemComponent::GetSourceInfoBySourcePath that will fail if the Asset Processor hasn't detected the
                    // existence of the file yet. With the flush, everything works correctly.

                    FlushIOOfAsset(targetPathAssetTmp);
```

Gems/ROS2/Code/Source/RobotImporter/Utils/SourceAssetsStorage.cpp
```diff
@@ -420,18 +431,8 @@ namespace ROS2::Utils
...
-                        AzFramework::AssetSystemRequestBus::BroadcastResult(
-                            copiedAssetStatus,
-                            &AzFramework::AssetSystem::AssetSystemRequests::GetAssetStatus_FlushIO,
-                            targetPathAssetDst.c_str());
-                        AZ_Warning(
-                            "CopyAssetForURDF",
-                            copiedAssetStatus != AzFramework::AssetSystem::AssetStatus::AssetStatus_Unknown,
-                            "Asset processor did not recognize the new file %s.",
-                            targetPathAssetDst.c_str());
+                        // call FlushIOOfAsset to ensure the asset processor is aware of the new file
```
michalpelka, line 434:
```suggestion
                        // call FlushIOOfAsset to ensure the asset processor is aware of the move operation.
```

michalpelka:
rebased to try resolve CI

PR #622, opened 2023-11-28
Add delete entity service

### LINKED ISSUE #618: Add a possibility to despawn entities
**Is your feature request related to a problem? Please describe.**
I would like to have a way to de-spawn entities that were spawned during the simulation when they are no longer relevant. This is especially important for cases where entities are complex e.g. when they are publishing multiple topics under different namespaces and or when new entities are periodically spawned during the simulation run.

**Describe the solution you'd like**
Two scenarios I have in mind are a de-spawn after preconfigured time or after reaching special area marked as a de-spawn point.

**Describe alternatives you've considered**
Possible alternative would be to move entity away from the scene/deactivate it or even maintain the list of entities for future use. But this logic seem more complex than removing entities when no longer needed.

arturkamieniecki:
I've added a service that allows to despawn a spawned entity.

I've modified the `spawn_entity` service to include the name of the ticket (this will become the name of the robot used to remove it later) in the `status_message`.

By calling the `gazebo_msgs/srv/DeleteEntity` service with the `name` parameter set to the response of the spawn service the entity will be deleted.

Resolves #618

### DIFF
--- a/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -10,6 +10,8 @@
 #include "Spawner/ROS2SpawnerComponentController.h"
 #include <AzCore/Serialization/EditContext.h>
 #include <AzCore/Serialization/SerializeContext.h>
+#include <AzCore/std/string/conversions.h>
+#include <AzCore/std/string/string.h>
 #include <AzFramework/Spawnable/Spawnable.h>
 #include <ROS2/Frame/ROS2FrameComponent.h>
 #include <ROS2/ROS2Bus.h>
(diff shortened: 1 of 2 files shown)


Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -48,6 +50,14 @@ namespace ROS2
                 SpawnEntity(service_handle, header, request);
             });
 
+        m_deleteService = ros2Node->create_service<gazebo_msgs::srv::DeleteEntity>(
+            "delete_entity",
+            [this](
+                const DeleteEntityServiceHandle service_handle, const std::shared_ptr<rmw_request_id_t> header, DeleteEntityRequest request)
```
adamdbrw, line 56: What is the reason for using this callback type as opposed to request/response one?

Is it about returning response earlier before Spawn / Despawn entities call is finished?
It would be valid to assume service can be blocked until items are fully spawned / despawned to avoid the need for extra checks on the client side.

michalpelka (approved):

arturkamieniecki:
Rebased to try to fix AR.

PR #637, opened 2023-12-19
Add AudioEngineWwise (25de560e)

AMZN-alexpete:
### Changes
- copies `AudioEngineWwise` gem from main `o3de` repo 25de560e into `o3de-extras`
- updates the version from `0.0.0` to `1.0.0`
- updates repo.json to include the gem for testing as a remote gem.
- adds wwise project, banks and ATL libs and copies them into projects that are missing them (https://github.com/o3de/o3de-extras/pull/637/commits/313083b04597e7bc1bd719286fd2cfaf68fc14ae)
- changes how bank_info_parser.py is run because it can no longer depend on the AudioEngineWwise gem being inside the engine like it did before. https://github.com/o3de/o3de-extras/pull/637/commits/34c582f61e820625f522f0424064587d58538539

### Description
All users must compile AudioEngineWwise from source because O3DE cannot distribute the binaries, so it make sense for the AudioEngineWwise gem to be moved here and added as a remote Gem that can be downloaded through `canonical.o3de.org` and not included in the base install of the O3DE SDK.

AudioEngineWwise will not be removed from the main O3DE Repo until AudioEngineWwise has been successfully been tested as a remote Gem that can be downloaded from `canonical.o3de.org` and used in projects.

Command used to update repo.json
```bash
o3de.bat edit-repo-properties --repo-path . --add-gem d:/git/o3de-extras/Gems/AudioEngineWwise --release-archive-path d:\git\o3de-extras\releases --download-prefix https://github.com/o3de/o3de-extras/releases/download/2.0
```

### Testing
Added this remote gem via the O3DE SDK and source-built ProjectManager, built project with gem and then opened the Editor, verify there were no errors.  Verified I could play a sound from the audio controls editor window and that it played in game.


spham-amzn:
Wont having it in o3de and o3de-extras cause problems with duplicate gems?

AMZN-alexpete:
> Wont having it in o3de and o3de-extras cause problems with duplicate gems?

It shouldn't because they are different gem versions, however @galibzon mentioned he was getting CMake errors about duplicate targets which IS an error you get when you have duplicates, so I'm looking into that now.  My initial testing was done with the O3DE SDK, and I'm testing with source next.

AMZN-alexpete:
I've added 

> Wont having it in o3de and o3de-extras cause problems with duplicate gems?

[I've updated the CMakeLists.txt to use a different name](https://github.com/o3de/o3de/pull/17140) for the targets in case you are using a source engine that has AudioEngineWwise (version 0.0.0) even if the project doesn't use that version.

I tested by adding `AudioEngineWwise==1.0.0` to AutomatedTesting and making sure it could compile and run and play audio.  In the Visual Studio Solution you see targets for `AudioEngineWwise` and `Wwise` because all the gems in the engine appear in the Solution even though not all are used (a different issue to fix).

Gems/AudioEngineWwise/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+    endmacro()
+endif()
+
+o3de_gem_setup("AudioEngineWwise")
+
+if(TARGET AudioEngineWwise)
+    # For backwards compatibility with engines that include AudioEngineWwise
+    # use Wwise for the targets if AudioEngineWwise is already defined
+    set(gem_name "Wwise")
+endif()
```
lumberyard-employee-dm, line 42: Actually the logic should check if the `@gemroot:AudioEngineWwise@` property is set and if so, call o3de_gem_setup on "Wwise" otherwise it can just use "AudioEngineWwise"

Though this backwards compability code will probably cause more issues than it solves by making users update the name of any references to `Gem::AudioEngineWwise` they have as a dependency of another Gem.

Gems/AudioEngineWwise/Code/CMakeLists.txt
```diff
@@ -0,0 +1,335 @@
...
+            file(COPY ${sounds_path_source} DESTINATION ${project_path})
+        endif()
+
+        cmake_path(SET audio_atl_path  "${project_path}/libs/gameaudio/wwise")
+        if (NOT EXISTS ${audio_atl_path})
+            message(STATUS "Copying default ATL libraries to ${audio_atl_path}")
+            cmake_path(SET audio_atl_path_source "${CMAKE_CURRENT_SOURCE_DIR}/../External/libs")
+            file(COPY ${audio_atl_path_source} DESTINATION ${project_path})
+        endif()
+    endforeach()
```
lumberyard-employee-dm, line 251: This should use a stamp file that can be used to track whether a file in the Wwise 3rdParty `External/Sounds` or External/libs` folder has been updated and needs to be recopied to the project.

You can use the logic the Python 3rdParty library uses to detect if the pip requirements files have been updated
https://github.com/o3de/o3de/blob/07122d2fb666e4da2aba809751bb40f8c4df4754/cmake/LYPython.cmake#L77-L91
https://github.com/o3de/o3de/blob/07122d2fb666e4da2aba809751bb40f8c4df4754/cmake/LYPython.cmake#L129-L131

AMZN-alexpete: This code only performs the initial population of the audio banks and other misc required files.  After they are added they should only be updated through the Wwise application's 'Generate AudioBanks' commands.

Gems/AudioEngineWwise/CMakeLists.txt
AMZN-alexpete: The way `o3de_gem_setup()` works is to pull the name from the `gem.json` so I'll need to override the property.

AMZN-alexpete: I tried using the `@GEMROOT:AudioEngineWwise@` method, but it is set based on code in `Subdirectores.cmake` which resolves it to the correct remote gem folder, but doesn't prevent the engine's AudioEngineWwise target from being added which is why I'm checking if the target exists.

Gems/AudioEngineWwise/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+if(NOT COMMAND o3de_gem_setup)
+    # Backwards compatibility for older engines
```
spham-amzn, line 10: Is a backwards compatibility check necessary if this gem is still in the engine repo and not o3de-extras?

AMZN-alexpete: This particular line is for engines that do not have the `o3de_gem_setup` function.  The development branch has this function but the latest point release doesn't.

spham-amzn: Yeah, but isn't this PR being merged into development? This line would make sense if this change is going to be cherry picked into the point release I guess

AMZN-alexpete: I was actually considering bypassing merging it to the point release, and instead getting it included in the next major release, and getting it added to the canonical repo.json so that it is available publicly after a brief period of testing.  Ideally, for the game jam, though that's coming up fast.

spham-amzn (approved):

lumberyard-employee-dm (approved):

PR #641, opened 2023-12-21
Update RobotImporter wizard

jhanca-robotecai:
This PR adds multiple changes to _RobotImporter_ wizard:
1. A bugfix for `Create prefab` button from the last wizard page - the button was never gone when the _back_ button was pressed
2. The page notifying about URDF parsing was redesigned not to include the modified source code anymore - this code can be either displayed in a separate window or saved to the file to use later. This change is caused by the fact that the source code can be long. Moreover, it might happen some other warnings are printed after to code. I believe it is hard to read for the most.
3. The page displaying prefab creating logs was redesigned to print _Markdown_ and the logs were split into categories. This part will be later used to display logs from SDFormat sensor and plugin importer.

I also renamed one file as it is used by URDF/SDFormat/Xacro, but the name suggests URDF only.

I am open for comments on the UI in 2.

Ad 2.
Before on the left, after on the right
image

Ad 3.
Before on the left, after on the right
image

### DIFF
--- a/Gems/ROS2/Code/Source/RobotImporter/Pages/ModifiedURDFWindow.cpp
+++ b/Gems/ROS2/Code/Source/RobotImporter/Pages/ModifiedURDFWindow.cpp
@@ -6,14 +6,14 @@
  *
  */
 
-#include "ModifiedUrdfWindow.h"
+#include "ModifiedURDFWindow.h"
 #include <AzCore/Utils/Utils.h>
 #include <QCloseEvent>
 #include <QVBoxLayout>
 
 namespace ROS2
 {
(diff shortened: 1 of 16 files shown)


jhanca-robotecai:
Added one more change:
4. Print information about articulation links - previously there was no information about joints in the prefab maker summary when `use articulations` flag was set to true

Before on the left, after on the right
image

michalpelka (approved):

adamdbrw (approved):

jhanca-robotecai:
Rebased and force-pushed on top of `development` to rerun and fix Jenkins output.

PR #609, opened 2023-11-13
Add SDFormat model plugins

jhanca-robotecai:
- Closes #610 
- Add hook template for SDFormat model plugins
- Add pipeline for processing hooks
- Add a sample hook for `libgazebo_ros_skid_steer_drive.so` and `libgazebo_ros_diff_drive`

The code was tested using modified _ROSBot_ urdf files. The only modification was adding the plugin itself based on [libgazebo_ros_skid_steer_drive docs](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Skid-Steer-drive) and [libgazebo_ros_diff_drive docs](https://github.com/ros-simulation/gazebo_ros_pkgs/wiki/ROS-2-Migration:-Diff-drive). Please find the robots [attached](https://github.com/o3de/o3de-extras/files/13447959/test_models.zip).

~~The attached robots [test_models.zip](https://github.com/o3de/o3de-extras/files/13447959/test_models.zip) require one change after import: setting "Use motor" toggle to true in all used wheels. This part will be fixed in the next PR.~~

Please note, that ROS topic names are not parsed at this point, therefore the robots can be controlled via `/base_link/default_topic` topic. This will be fixed in another PR (topic names are not parsed in Sensor importer either).

PR #642, opened 2024-01-02
Fix SkidSteeringDriveModel

### LINKED ISSUE #615: Fix skid steering drive model parameters
Wheelbase and track
[Wheelbase definition](https://en.wikipedia.org/wiki/Wheelbase)

The speed of the wheels depends primarily on the distance between the wheels in the skid steering drive model. In o3de-extras implementation, the skid steering drive model uses _wheelbase_ parameter, but it does not use _track_ parameter.
https://github.com/o3de/o3de-extras/blob/153d83cbf09695c2e141e5016ba7a3ba35ef5709/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp#L162

Therefore I believe the terms are used incorrectly for this model. The configuration is shared with Ackermann implementation, in which terms are used correctly.  Please keep in mind, that changing the name in the code might influence all available prefabs using this drive model.

jhanca-robotecai:
## What does this PR do?

This PR fixes some issues in _SkidSteeringDriveModel_ implementation. Both the method for applying the rotation to wheels based on the requested model movement and the method for calculating the model's movement based on the wheels' rotation were refactored.
- The track parameter is used instead of the wheelbase parameter for calculation (Closes #615), the naming is also unified and fixed.
- The same distance between wheels is used in the codebase (it was two times bigger while [getting the speed](https://github.com/o3de/o3de-extras/blob/d9fc445a72a30ed683cf6cb18dfa8b3a3e762f6a/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp#L60) than while [applying the speed](https://github.com/o3de/o3de-extras/blob/d9fc445a72a30ed683cf6cb18dfa8b3a3e762f6a/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp#L162). This part was working correctly only for robots with two wheels per axle, now the code is more readable and works for any number of wheels.
- One buffer instead of [two separate ones](https://github.com/o3de/o3de-extras/blob/d9fc445a72a30ed683cf6cb18dfa8b3a3e762f6a/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.h#L55) is used to cache data; there is no recalculation of some of it in [every step](https://github.com/o3de/o3de-extras/blob/d9fc445a72a30ed683cf6cb18dfa8b3a3e762f6a/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp#L160).
- The correct number of axles is used in the calculation (non-drivable axles were taken into consideration [previously](https://github.com/o3de/o3de-extras/blob/d9fc445a72a30ed683cf6cb18dfa8b3a3e762f6a/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp#L83) ).
- Segfault triggered when the [number of wheels == 1](https://github.com/o3de/o3de-extras/blob/153d83cbf09695c2e141e5016ba7a3ba35ef5709/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp#L161) is removed.
- The names such as `axis` and `axle` are unified along the codebase and used correctly.

## How was this PR tested?

- create an empty level
- import robot with SkidSteering (ROSbot in my case)
- add `ROS2 Wheel Odometry Sensor`
- send commands to move the robot via ROS2 and echo the odometry in another terminal - the values should be approximately the same
```bash
ros2 topic pub -r 10 /cmd_vel geometry_msgs/msg/Twist "{linear: {x: 0.5, y: 0.0, z: 0.0}, angular: {x: 0.0, y: 0.0, z: 1.0}}" 
ros2 topic echo /odom
```

### DIFF
--- a/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp
+++ b/Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp
@@ -43,11 +43,11 @@ namespace ROS2::VehicleDynamics
     void SkidSteeringDriveModel::Activate(const VehicleConfiguration& vehicleConfig)
     {
         m_config = vehicleConfig;
-        m_wheelsCache.clear();
+        m_wheelsData.clear();
         m_initialized = false;
     }
 
-    void SkidSteeringDriveModel::InitializeCache()
+    void SkidSteeringDriveModel::ComputeWheelsData()
     {
         // Compute number of drive axles to scale the contribution of each wheel to vehicle's velocity
         int driveAxlesCount = 0;
@@ -76,10 +76,10 @@ namespace ROS2::VehicleDynamics
             }
             for (size_t wheelId = 0; wheelId < wheelCount; wheelId++)
             {
-                const auto column = ProduceWheelColumn(wheelId, axle, driveAxlesCount);
-                if (column.wheelControllerComponentPtr != nullptr)
+                const auto data = ComputeSingleWheelData(wheelId, axle, driveAxlesCount);
+                if (data.wheelControllerComponentPtr != nullptr)
                 {
-                    m_wheelsCache.emplace_back(AZStd::move(column));
+                    m_wheelsData.emplace_back(AZStd::move(data));
                 }
             }
         }
@@ -87,7 +87,7 @@ namespace ROS2::VehicleDynamics
         m_initialized = true;
     }
 
-    SkidSteeringDriveModel::SkidSteeringWheelData SkidSteeringDriveModel::ProduceWheelColumn(
+    SkidSteeringDriveModel::SkidSteeringWheelData SkidSteeringDriveModel::ComputeSingleWheelData(
         const int wheelId, const AxleConfiguration& axle, const int axlesCount) const
     {
         SkidSteeringDriveModel::SkidSteeringWheelData out;
@@ -118,13 +118,13 @@ namespace ROS2::VehicleDynamics
     {
         if (!m_initialized)
         {
-            InitializeCache();
+            ComputeWheelsData();
         }
 
         float dX = 0; // accumulated contribution to vehicle's linear movements of every wheel
         float dPhi = 0; // accumulated contribution to vehicle's rotational movements of every wheel
 
-        for (const auto& wheel : m_wheelsCache)
+        for (const auto& wheel : m_wheelsData)
         {
             const float omega = wheel.axis.Dot(wheel.wheelControllerComponentPtr->GetRotationVelocity());
             dX += omega * wheel.dX;
@@ -143,7 +143,7 @@ namespace ROS2::VehicleDynamics
 
         if (!m_initialized)
         {
-            InitializeCache();
+            ComputeWheelsData();
         }
 
         const float angularTargetSpeed = inputs.m_angularRates.GetZ();
@@ -158,7 +158,7 @@ namespace ROS2::VehicleDynamics
         m_currentLinearVelocity =
             Utilities::ComputeRampVelocity(linearTargetSpeed, m_currentLinearVelocity, deltaTimeNs, linearAcceleration, maxLinearVelocity);
 
-        for (const auto& wheel : m_wheelsCache)
+        for (const auto& wheel : m_wheelsData)
         {
             const float wheelRate =
                 (m_currentLinearVelocity + m_currentAngularVelocity * wheel.wheelPosition) / wheel.wheelData.m_wheelRadius;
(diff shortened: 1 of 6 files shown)


adamkrawczyk (approved):

Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp
```diff
@@ -38,72 +38,95 @@ namespace ROS2::VehicleDynamics
...
+                {
+                    m_wheelsCache.emplace_back(AZStd::move(column));
+                }
+            }
+        }
+
+        m_initialized = true;
+    }
+
+    SkidSteeringDriveModel::SkidSteeringWheelData SkidSteeringDriveModel::ProduceWheelColumn(
```
adamkrawczyk, line 85: Consider move semantic for `SkidSteeringDriveModel` to optimize return. But since this is not executed periodically it's not crucial. Nice to have.

Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.h
```diff
@@ -37,25 +37,43 @@ namespace ROS2::VehicleDynamics
         AZStd::pair<AZ::Vector3, AZ::Vector3> GetVelocityFromModel() override;
 
     private:
+        // Data collected once and cached for the reuse
```
adamdbrw, line 40: This comment should be next to the member (m_wheelsCache instead). 
Instead, describe what the structure is for here.

Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.h
```diff
@@ -37,25 +37,43 @@ namespace ROS2::VehicleDynamics
...
+        //! It can be thought of as a column of the Jacobian matrix of the mechanical system. Jacobian matrix for this model is a matrix
+        //! of size 2 x number of wheels. This function returns elements of column that corresponds to the given wheel and cache
+        //! necessary data to find the wheel's rotation as a scalar.
+        //! @param wheelId - id of the currently processed wheel in the axle vector
         //! @param axle - the wheel's axle configuration
-        //! @param axisCount - number of axles in the vehicle
-        //! @returns A tuple containing of :
-        //!  - pointer to WheelControllerComponent (API to query for wheel's ration speed),
+        //! @param axlesCount - number of axles in the vehicle
+        //! @returns A structure containing of:
```
adamdbrw, line 64: This documentation should link to the structure documentation, here just link to the structure.

Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.h
```diff
@@ -37,25 +37,43 @@ namespace ROS2::VehicleDynamics
...
+            float dX{ 0.0f }; // wheel's contribution to vehicle's linear movement
+            float dPhi{ 0.0f }; // wheel's contribution to vehicle's rotational movement
+            AZ::Vector3 axis{ AZ::Vector3::CreateZero() }; // rotation axis of the wheel
+        };
+        AZStd::vector<SkidSteeringWheelData> m_wheelsCache;
+        bool m_initialized = false;
+
+        //! Collect all necessary data to compute the impact of wheels on the vehicle's velocity.
+        //! The data is stored in cache and reused.
+        void InitializeCache();
```
adamdbrw, line 55: Better call it ComputeWheelVelocityImpact or similar - focus on semantics (What happens when this function is called) and not implementation (Cache). Cache is a bit misleading here.

Gems/ROS2/Code/Source/VehicleDynamics/DriveModels/SkidSteeringDriveModel.cpp
jhanca-robotecai: Please correct me if I am wrong, but the temporary in `SkidSteeringDriveModel::ProduceWheelColumn` is implicitly treated as an rvalue in the return statement by the compiler's RVO.

adamkrawczyk: You are right! I have checked and according to the docs this is a perfect candidate for rvo.

michalpelka (approved):

jhanca-robotecai:
Rebased and force-pushed after #609 was merged to resolve conflicts - only conflicts in one comment found, hence no need for another review.

adamdbrw (approved):
LGTM, good work!

PR #643, opened 2024-01-04
Ros2 camera image encoding 

MateuszWasilewski:
## What does this PR do?

This PR modifies `CameraPostProcessingRequestBus` to allow for multiple ordered handlers.

In principle this will allow for composable usage
1. PostProcessImage
2. PostProcessImageInOtherWay
3. Convert image type

This is a similar behavior as o3de Pass System but is supports ros2 Image type.
`SupportsFormat` function was removed because it's semantic would be unclear when many handlers are present. Additionally, it's behavior can be achieved by ignoring Post Processing requests on the Handler side.

Provided `ROS2ImageEncodingConvertComponent` implementation leaves room for further additions of other encoding conversions.

## How was this PR tested?

After inserting `ros2 camera component` on empty level
```
ros2 topic echo <rgb_camera_topic>
```
shows `rgba8` image encoding.

After adding `ROS2ImageEncodingConvertComponent` to the same Entity, and setting conversion `rgba8 -> rgb8`
```
ros2 topic echo <rgb_camera_topic>
```
shows `rgb8` image encoding.

Additionally testing with 2 `CameraPostProcessingRequestBus` also yields expected results.

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
@@ -40,8 +40,9 @@ namespace ROS2
 
         //! Apply post-processing function, if any implementations to the bus are in the entity.
         //! @param image standard image message passed as a reference. It will be changed through post-processing.
-        //! @note Handler should always handle function call, even if it does not support the format of the image.
-        //! in this case it should not change the image.
+        //! @note Handler should always handle function call,
+        //! if handler does not support the format of the image, it should not change the image.
+        //! @note Every image is modified in place. Only one handler has access to the image at the time.
         virtual void ApplyPostProcessing(sensor_msgs::msg::Image& image) = 0;
 
         //! Get priority of the post-processing bus.
--- a/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
@@ -47,7 +47,7 @@ namespace ROS2
         //! Get priority of the post-processing bus.
         //! @return priority of the bus.
         //! @note higher priority buses will be processed first.
-        virtual AZ::u8 GetPriority() const;
+        virtual AZ::u8 GetPriority() const = 0;
 
         //! Compare two post-processing buses.
         //! @param other bus to compare to.
--- a/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
@@ -5,10 +5,11 @@
  * SPDX-License-Identifier: Apache-2.0 OR MIT
  *
  */
+
 #pragma once
 
-#include "AzCore/EBus/EBusSharedDispatchTraits.h"
 #include <AzCore/Component/EntityId.h>
+#include <AzCore/EBus/EBusSharedDispatchTraits.h>
 #include <AzCore/Interface/Interface.h>
 #include <AzCore/std/string/string.h>
 #include <sensor_msgs/msg/image.hpp>
@@ -24,21 +25,37 @@ namespace ROS2
     class CameraPostProcessingRequests : public AZ::EBusSharedDispatchTraits<CameraPostProcessingRequests>
     {
     public:
+        //! Each camera sensor component has its own post-processing bus.
         using BusIdType = AZ::EntityId;
         static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::ById;
-        static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Multiple;
+        //! Multiple post-processing functions can be registered to the bus.
+        //! They will be executed in the order
+        static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::MultipleAndOrdered;
+
+        //! Priority of the post-processing bus.
+        //! @note higher priority buses will be processed first.
+        constexpr static AZ::u8 MIN_PRIORITY = 0;
+        constexpr static AZ::u8 MAX_PRIORITY = 255;
+        constexpr static AZ::u8 DEFAULT_PRIORITY = 127;
 
         //! Apply post-processing function, if any implementations to the bus are in the entity.
         //! @param image standard image message passed as a reference. It will be changed through post-processing.
-        //! @note you should check whether the encoding format is supported first with SupportsFormat.
+        //! @note Handler should always handle function call, even if it does not support the format of the image.
+        //! in this case it should not change the image.
         virtual void ApplyPostProcessing(sensor_msgs::msg::Image& image) = 0;
 
-        //! Query whether a particular image encoding is supported by registered postprocessing.
-        //! @param encodingFormat name of the format.
-        virtual bool SupportsFormat(const AZStd::string& encodingFormat) = 0;
+        //! Get priority of the post-processing bus.
+        //! @return priority of the bus.
+        //! @note higher priority buses will be processed first.
+        virtual AZ::u8 GetPriority() const;
 
-    protected:
-        ~CameraPostProcessingRequests() = default;
+        //! Compare two post-processing buses.
+        //! @param other bus to compare to.
+        //! @return true if this bus should be processed before the other.
+        inline bool Compare(const CameraPostProcessingRequests* other) const
+        {
+            return GetPriority() > other->GetPriority();
+        }
     };
 
     using CameraPostProcessingRequestBus = AZ::EBus<CameraPostProcessingRequests>;
--- a/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
@@ -34,9 +34,9 @@ namespace ROS2
 
         //! Priority of the post-processing bus.
         //! @note higher priority buses will be processed first.
-        constexpr static AZ::u8 MIN_PRIORITY = 0;
-        constexpr static AZ::u8 MAX_PRIORITY = 255;
-        constexpr static AZ::u8 DEFAULT_PRIORITY = 127;
+        static constexpr AZ::u8 MIN_PRIORITY = 0;
+        static constexpr AZ::u8 MAX_PRIORITY = 255;
+        static constexpr AZ::u8 DEFAULT_PRIORITY = 127;
 
         //! Apply post-processing function, if any implementations to the bus are in the entity.
         //! @param image standard image message passed as a reference. It will be changed through post-processing.
@@ -50,9 +50,9 @@ namespace ROS2
         //! @note higher priority buses will be processed first.
         virtual AZ::u8 GetPriority() const = 0;
 
-        //! Compare two post-processing buses.
-        //! @param other bus to compare to.
-        //! @return true if this bus should be processed before the other.
+        //! Compare function required by BusHandlerOrderCompare = BusHandlerCompareDefault
+        //! \param[in] other Another instance of the class to compare
+        //! @return True if this bus should be processed before the other.
         inline bool Compare(const CameraPostProcessingRequests* other) const
         {
             return GetPriority() > other->GetPriority();
--- a/Gems/ROS2/Code/Source/Camera/CameraSensor.h
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensor.h
@@ -7,13 +7,10 @@
  */
 #pragma once
 
+#include "CameraPublishers.h"
 #include <Atom/Feature/Utils/FrameCaptureBus.h>
 #include <AzCore/std/containers/span.h>
-
-#include "CameraPublishers.h"
 #include <ROS2/ROS2GemUtilities.h>
-
-#include <chrono>
 #include <rclcpp/publisher.hpp>
 #include <sensor_msgs/msg/camera_info.hpp>
 #include <sensor_msgs/msg/image.hpp>
--- a/Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
+++ b/Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
@@ -6,9 +6,8 @@
  *
  */
 
-#include "ROS2ImageFormatConvertComponent.h"
-#include "AzCore/RTTI/RTTIMacros.h"
-#include "AzCore/Serialization/EditContextConstants.inl"
+#include "ROS2ImageEncodingConversionComponent.h"
+#include <AzCore/RTTI/RTTIMacros.h>
 #include <AzCore/Serialization/EditContext.h>
 #include <AzCore/Serialization/SerializeContext.h>
 #include <sensor_msgs/msg/detail/image__struct.hpp>
@@ -16,9 +15,9 @@
 namespace AZStd
 {
     template<>
-    struct hash<ROS2::EncodingConvertData>
+    struct hash<ROS2::EncodingConversion>
     {
-        size_t operator()(const ROS2::EncodingConvertData& data) const
+        size_t operator()(const ROS2::EncodingConversion& data) const
         {
             return (static_cast<AZ::u16>(data.encodingIn) << 8) | static_cast<AZ::u16>(data.encodingOut);
         }
@@ -35,6 +34,7 @@ namespace ROS2
             AZ_Assert(image.step == image.width * 4, "Image step is not width * 4");
             AZ_Assert(image.data.size() == image.step * image.height, "Image data size is not step * height");
 
+            // Perform conversion in place
             for (size_t pixelId = 0; pixelId < image.width * image.height; ++pixelId)
             {
                 size_t pixelOffsetIn = pixelId * 4;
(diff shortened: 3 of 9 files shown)


Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConvertComponent.h
```diff
@@ -0,0 +1,64 @@
...
+#include <ROS2/Camera/CameraPostProcessingRequestBus.h>
+
+namespace ROS2
+{
+    enum class ImageEncoding : AZ::u8
+    {
+        rgba8,
+        rgb8,
+        mono8,
+        mono16,
```
jhanca-robotecai, line 22: Enums in o3de should be [UpperCamelCase](https://github.com/o3de/sig-core/blob/main/governance/Coding-Standards-and-Style-Guide.md#enums).

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConvertComponent.h
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    struct EncodingConverion
```
jhanca-robotecai, line 25: Typo
```suggestion
    struct EncodingConversion
```

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConvertComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+                    ->EnumAttribute(ImageEncoding::rgba8, "rgba8")
+                    ->EnumAttribute(ImageEncoding::rgb8, "rgb8")
+                    ->EnumAttribute(ImageEncoding::mono8, "mono8")
+                    ->EnumAttribute(ImageEncoding::mono16, "mono16")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::ComboBox, &EncodingConverion::encodingOut, "Encoding Out", "Encoding of the output image")
+                    ->EnumAttribute(ImageEncoding::rgba8, "rgba8")
+                    ->EnumAttribute(ImageEncoding::rgb8, "rgb8")
+                    ->EnumAttribute(ImageEncoding::mono8, "mono8")
+                    ->EnumAttribute(ImageEncoding::mono16, "mono16");
```
jhanca-robotecai, line 92: Don't you want to validate at this point if the input format is different from the output format? You could do it in `ROS2ImageEncodingConvertComponent::ValidateEncodingConversion` as well.

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConvertComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+    };
+} // namespace AZStd
+
+namespace ROS2
+{
+    namespace
+    {
+        void Rgba8ToRgb8(sensor_msgs::msg::Image& image)
+        {
+            AZ_Assert(image.encoding == "rgba8", "Image encoding is not rgba8");
```
jhanca-robotecai, line 33: You defined `ImageEncodingNames`, I believe it makes sense to use it in case it changes one day (e.g. to capital letters).

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConvertComponent.cpp
```diff
@@ -0,0 +1,185 @@
...
+            // Perform conversion in place
+            for (size_t pixelId = 0; pixelId < image.width * image.height; ++pixelId)
+            {
+                size_t pixelOffsetIn = pixelId * 4;
+                size_t pixelOffsetOut = pixelId * 3;
+                image.data[pixelOffsetOut] = image.data[pixelOffsetIn];
+                image.data[pixelOffsetOut + 1] = image.data[pixelOffsetIn + 1];
+                image.data[pixelOffsetOut + 2] = image.data[pixelOffsetIn + 2];
+            }
+            image.encoding = "rgb8";
```
jhanca-robotecai, line 46: You defined `ImageEncodingNames`, I believe it makes sense to use it in case it changes one day (e.g. to capital letters).

Gems/ROS2/Code/Source/Camera/CameraSensor.h
```diff
@@ -7,13 +7,10 @@
  */
 #pragma once
 
+#include "CameraPublishers.h"
```
jhanca-robotecai, line 10: Is this one needed?

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+        {
+            serializeContext->Class<EncodingConversion>()
+                ->Version(0)
+                ->Field("EncodingIn", &EncodingConversion::encodingIn)
+                ->Field("EncodingOut", &EncodingConversion::encodingOut);
+
+            if (AZ::EditContext* ec = serializeContext->GetEditContext())
+            {
+                ec->Class<EncodingConversion>("Encoding Conversion", "Specifies encoding conversion")
+                    ->DataElement(
```
michalpelka, line 101: Please correct me if I am wrong. Does the user need to put encoding of image coming from Atom?

jhanca-robotecai (approved):
All my concerns were addressed, the code works as expected. The only thing that could be somewhat changed is the name of the component - my intuition was to type "camera" in a filter box and, as expected, I did not find it. I don't have a better idea though.

Without conversion:
```
height: 2
width: 2
encoding: rgba8
is_bigendian: 0
step: 8
data:
- 135
- 135
- 138
- 255
- 98
- 95
- 91
- 255
- 147
- 152
- 158
- 255

```

With:
```
height: 2
width: 2
encoding: rgb8
is_bigendian: 0
step: 6
data:
- 135
- 135
- 138
- 98
- 95
- 91
- 147
- 152
- 158

```

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
MateuszWasilewski: Yes, user needs to put encoding from Atom, this allows for having separate conversions  from
- rgba8
- mono8
to any other encoding if desired, It allows for piping encoding conversions with any other operations.

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    void ROS2ImageEncodingConversionComponent::Deactivate()
+    {
+        CameraPostProcessingRequestBus::Handler::BusDisconnect();
+    }
+
+    void ROS2ImageEncodingConversionComponent::ApplyPostProcessing(sensor_msgs::msg::Image& image)
+    {
+        auto nameIter = ImageEncodingFromName.find(image.encoding.c_str());
+        if (nameIter == ImageEncodingFromName.end())
```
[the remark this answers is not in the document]
MateuszWasilewski: Assert would forbid using Encoding Conversion with RGBD camera. Current solution acts as a pass through when encoding does not match. Example usecase:
```
Camera -> RGB conversion -> output
   \
   Depth conversion -> output
```
Which in data flow may look like:
```
Camera -> RGB conversion -> Depth conversion -> output

**Comment by MateuszWasilewski** [comment]
@jhanca-robotecai potencial solution would be naming component `ROS2CameraSensorEncodingConversionComponent` or similar but it is a needlessly complex name.

File: Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h [hunk 9]
```diff
@@ -24,21 +25,38 @@ namespace ROS2
...
-        static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Multiple;
+        //! Multiple post-processing functions can be registered to the bus.
+        //! They will be executed in the order
+        static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::MultipleAndOrdered;
+
+        //! Priority of the post-processing bus.
+        //! @note higher priority buses will be processed first.
+        constexpr static AZ::u8 MIN_PRIORITY = 0;
+        constexpr static AZ::u8 MAX_PRIORITY = 255;
+        constexpr static AZ::u8 DEFAULT_PRIORITY = 127;
```
adamkrawczyk, line 39: For consistency, the static should be first

Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
```diff
@@ -24,21 +25,38 @@ namespace ROS2
...
+        //! @return priority of the bus.
+        //! @note higher priority buses will be processed first.
+        virtual AZ::u8 GetPriority() const = 0;
 
-    protected:
-        ~CameraPostProcessingRequests() = default;
+        //! Compare two post-processing buses.
+        //! @param other bus to compare to.
+        //! @return true if this bus should be processed before the other.
+        inline bool Compare(const CameraPostProcessingRequests* other) const
```
spham-amzn, line 56: The name of this function is a little misleading to me. Its comparing priorities and returning of the current bus has a higher priority than the input, maybe rename to `IsHigherPriority` or `ComparePriority` ?

spham-amzn, line 56: Could also make the argument pass by reference to prevent accidental null-pointers.
```suggestion
        inline bool Compare(const CameraPostProcessingRequests& other) const
```

Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
```diff
@@ -24,21 +25,38 @@ namespace ROS2
...
+        {
+            return GetPriority() > other->GetPriority();
```
spham-amzn, line 58: Based on the pass by reference comment above.

```suggestion
            return GetPriority() > other.GetPriority();
```

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+            { "mono8", ImageEncoding::Mono8 },
+            { "mono16", ImageEncoding::Mono16 },
+        };
+
+        void Rgba8ToRgb8(sensor_msgs::msg::Image& image)
+        {
+            const std::string inputEncoding = ImageEncodingNames.at(ImageEncoding::RGBA8);
+            const std::string outputEncoding = ImageEncodingNames.at(ImageEncoding::RGB8);
+            AZ_Assert(image.encoding == inputEncoding, "Image encoding is %s, expected %s", image.encoding.c_str(), inputEncoding.c_str());
+            AZ_Assert(image.step == image.width * 4, "Image step is not width * 4");
```
spham-amzn, line 50:
```suggestion
            AZ_Assert(image.step == image.width * 4, "Image step is not width * 4 (%d != %d)", image.step, image.width * 4);
```

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+            AZ_Assert(image.data.size() == image.step * image.height, "Image data size is not step * height");
```
spham-amzn, line 51:
```suggestion
            AZ_Assert(image.data.size() == image.step * image.height, "Image data size (%d) is not step * height (%d)", image.data.size(), image.step * image.height);
```

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+            {
+                size_t pixelOffsetIn = pixelId * 4;
+                size_t pixelOffsetOut = pixelId * 3;
+                image.data[pixelOffsetOut] = image.data[pixelOffsetIn];
+                image.data[pixelOffsetOut + 1] = image.data[pixelOffsetIn + 1];
+                image.data[pixelOffsetOut + 2] = image.data[pixelOffsetIn + 2];
+            }
+            image.encoding = outputEncoding;
+            image.step = image.width * 3;
+            image.data.resize(image.step * image.height);
```
spham-amzn, line 64: Shouldn't the resize of the image data vector occur before the loop that populates it above?

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+        const AZStd::unordered_map<EncodingConversion, const AZStd::function<void(sensor_msgs::msg::Image&)>> supportedFormatChange = {
+            { { ImageEncoding::RGBA8, ImageEncoding::RGB8 }, Rgba8ToRgb8 },
+        };
+
+        AZ::Outcome<void, AZStd::string> ValidateEncodingConversion(EncodingConversion newConversion)
+        {
+            if (newConversion.encodingIn == newConversion.encodingOut)
+            {
+                return AZ::Failure(AZStd::string("Conversion to same type if forbidden"));
```
spham-amzn, line 75:
```suggestion
                return AZ::Failure(AZStd::string("Conversion to same type is forbidden"));
```

spham-amzn: Is it forbidden? Or can be a no-op?

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+                    ->EnumAttribute(ImageEncoding::Mono8, "mono8")
+                    ->EnumAttribute(ImageEncoding::Mono16, "mono16")
+                    ->Attribute(AZ::Edit::Attributes::ChangeValidate, &EncodingConversion::ValidateOutputEncoding);
+            }
+        }
+    }
+
+    AZ::Outcome<void, AZStd::string> EncodingConversion::ValidateInputEncoding(void* newValue, const AZ::Uuid& valueType)
+    {
+        AZ_TracePrintf("EncodingConversion", "ValidateInputEncodingConversion");
```
spham-amzn, line 121: `AZ_TracePrintf` is deprecated, use AZ_Trace or AZ_Info instead (see https://github.com/o3de/o3de/issues/14838 )

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
```diff
@@ -0,0 +1,206 @@
...
+        ImageEncoding* newEncoding = static_cast<ImageEncoding*>(newValue);
+        return ValidateEncodingConversion({ *newEncoding, encodingOut });
+    }
+
+    AZ::Outcome<void, AZStd::string> EncodingConversion::ValidateOutputEncoding(void* newValue, const AZ::Uuid& valueType)
+    {
+        AZ_TracePrintf("EncodingConversion", "ValidateOutputEncodingConversion");
```
spham-amzn, line 128: See https://github.com/o3de/o3de/issues/14838

Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
MateuszWasilewski: Function with this convention is required for correct EBus ordering
According to `EBus.h`
```cpp
        /**
         * Sorting function for EBus event handlers.
         * Used only when the #HandlerPolicy is AZ::EBusHandlerPolicy::MultipleAndOrdered.
         * This function determines the order in which handlers at an address receive an event.
         *
         * By default, the function requires the handler to implement the following comparison
         * function.
         * @code{.cpp}
         * // Returns whether 'this' should precede 'other'.
         * bool Compare(const Interface* other) const;
         * @endcode
         */
        using BusHandlerOrderCompare = BusHandlerCompareDefault;

```

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
MateuszWasilewski: In principle it could be a no-op but banning it simplifies workflow for the user who gets immediate feedback (rather then debugging why format doesn't change)

Gems/ROS2/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.cpp
MateuszWasilewski: during population loop we read from [0, oldSize) and write to [0, newSize].
According to [`cppreference`](https://en.cppreference.com/w/cpp/container/vector/resize) `Vector capacity is never reduced when resizing to smaller size ...`
resizing before the loop is correct but more confusing (reading from outsize the vector size)

michalpelka (approved):

Gems/ROS2/Code/Include/ROS2/Camera/CameraPostProcessingRequestBus.h
spham-amzn: Thanks for the clarification

spham-amzn (approved):

PR #593, opened 2023-10-30
ROS2 frame editor component

arturkamieniecki:
Edit from @michalpelka to mainteners:
It is ***BREAKING CHANGE*** since ***every ROS 2*** prefab will be **affected**. Please merge it with caution.

@arturkamieniecki message:
I've created a ROS2FrameEditorComponent and ROS2FrameSystemComponent. This change allows during editing to see the effective namespaces of the frames.

The new ROS2FrameNotificationBus allows components to get notified about changes in the frame configuration.

ROS2FrameSystemComponent keeps track of all frames in the level and applies changes to the frame components when needed. (During movement of the frame entities and changes in the components configuration).

I've also updated all templates and prefabs associated with this change.

Acompaniamenting PR that change projects to the ROS2FrameEditorComponent:
https://github.com/RobotecAI/ROSCon2023Demo/pull/263
https://github.com/o3de/RobotVacuumSample/pull/33
https://github.com/o3de/ROSConDemo/pull/268

Resolves https://github.com/o3de/o3de-extras/issues/131

PR #650, opened 2024-01-15
Update SDK version

michalpelka:
## What does this PR do?

Since some braking changes are introduced to ROS 2 Gem (e.g. #593), we need to update major version of ROS 2 Gem.

## How was this PR tested?

Build against o3de dev branch `496f6130575367a56d99f54038d00818b1800523`


michalpelka:
According to a discussion in #593 - I will update this PR #593  to the minimum required version of ROS2 gem in project templates.

arturkamieniecki:
I've bumped the ROS2 gem versions and fixed broken tests.

Gems/ROS2/gem.json
```diff
@@ -18,8 +18,8 @@
         "ROS2"
     ],
     "compatible_engines": [
-        "o3de-sdk>=2.1.0",
-        "o3de>=2.1.0"
+        "o3de-sdk>=4.1.0",
+        "o3de>=4.1.0"
```
spham-amzn, line 22: Shouldn't this be '3'?

Gems/ROS2/gem.json
```diff
@@ -18,8 +18,8 @@
         "ROS2"
     ],
     "compatible_engines": [
-        "o3de-sdk>=2.1.0",
-        "o3de>=2.1.0"
+        "o3de-sdk>=4.1.0",
```
michalpelka, line 21:
```suggestion
        "o3de-sdk>=3.1.0",
        "o3de>=3.1.0"
```

Gems/ROS2/gem.json
michalpelka: Set to 3.1.0. Thanks!

spham-amzn (approved):

Gems/ROS2/gem.json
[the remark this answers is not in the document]
spham-amzn: I just checked with development, the current engine version is "4.0.0"

spham-amzn: Sorry, I just checked o3de/development and the engine version is [4.0.0](https://github.com/o3de/o3de/blob/57c811f7fb5505600b081f14bf197b3ce1c26183/engine.json#L7), sorry about that.

jhanca-robotecai (commented):
I tested ROS2ProjectTemplate with extra Proteus robot added - works as expected.

I am not sure about the `o3de` version though - based on @spham-amzn check:
> I just checked with development, the current engine version is "4.0.0"

I believe we should target `>=4.0.0`, as the current `development` branch in `extras` will not build correctly with `2310.x` due to linking errors after changes in module names in cmake.

adamdbrw (approved):

PR #655, opened 2024-01-19
Add feature to prevent camera from tilting

michalpelka:
## What does this PR do?

Following camera allows to switch camera to many different location, but it can tilt (horizon tilts with robot).
With this feature, the rotation matrix is decomposed in such manner that pitch and yaw is removed.

## How was this PR tested?
Manual test with internal project.

### DIFF
--- a/Gems/ROS2/Code/Source/SimulationUtils/FollowingCameraComponent.cpp
+++ b/Gems/ROS2/Code/Source/SimulationUtils/FollowingCameraComponent.cpp
@@ -29,6 +29,8 @@ namespace ROS2
     {
     }
 
+
+
     void FollowingCameraComponent::Reflect(AZ::ReflectContext* reflection)
     {
         FollowingCameraConfiguration::Reflect(reflection);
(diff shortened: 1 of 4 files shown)


spham-amzn:
Is it possible to set the rotation of the transform (`SetRotation`) with a quaternion as well?

spham-amzn (approved):

Gems/ROS2/Code/Source/SimulationUtils/FollowingCameraComponent.cpp
```diff
@@ -129,7 +164,7 @@ namespace ROS2
         // get parent's transform
         const AZ::Transform parent_transform = target_world_transform * target_local_transform.GetInverse();
 
-        CacheTransform(parent_transform, deltaTime);
+        CacheTransform(m_configuration.m_lockZAxis?RemoveTiltFromTransform(parent_transform):parent_transform , deltaTime);
```
jhanca-robotecai, line 167: Nit: it would be more readable if formatted.

jhanca-robotecai (approved):
I tested the code driving an Ackermann model on a bumpy ground, the results are great!

PR #665, opened 2024-02-01
Update PhysX5 references for robots related files

spham-amzn:
## What does this PR do?
This updates all the robot related project templates and gems (ROS2) to reference 'PhysX5' instead of 'PhysX' for simulation, now that both gems will be available at the same time (without the need of a cmake flag when creating the project from source). (Refer to https://github.com/o3de/o3de/pull/17436 ).
Update templates that relate to ROS2 to tag version 3.1.0

## How was this PR tested?
From a source build,  created projects from the following templates:

* Ros2FleetRobotTemplate
* Ros2ProjectTemplate
* Ros2RoboticManipulationTemplate

From the installer (SDK) build performed the same project creation workflows above

Built and launch rVIZ for each of the templates above.

RFC: https://github.com/o3de/sig-simulation/issues/85


michalpelka (approved):

Templates/Ros2FleetRobotTemplate/Template/Registry/physxdebugconfiguration.setreg
```diff
@@ -1,7 +1,7 @@
 {
     "Amazon": {
         "Gems": {
-            "PhysX": {
+            "PhysX5": {
```
lumberyard-employee-dm, line 4: The name of the Gem doesn't affect the Settings Registry settings path for the PhysX5 Gem,

That is controlled through the [PhysXSettingsRegistryManager](https://github.com/o3de/o3de/blob/36ad79e819835e71dc2a2d41430894917b48b7e4/Gems/PhysX/Core/Code/Source/Configuration/PhysXSettingsRegistryManager.cpp#L19-L21) which has not been change by the [PR](https://github.com/o3de/o3de/pull/17436/files) to split PhysX 4 and 5 into separate gems.
The settings path should still be "/Amazon/Gems/PhysX"

lumberyard-employee-dm (changes requested):

Gems/ROS2/gem.json
```diff
@@ -29,7 +29,7 @@
         "Atom_Feature_Common",
         "Atom_Component_DebugCamera",
         "CommonFeaturesAtom",
-        "PhysX",
+        "PhysX5",
```
AMZN-alexpete, line 32: We should probably update the `compatible_engines` in this gem.json to reflect the versions of O3DE that will have the PhysX5 gem.

Gems/WarehouseAutomation/gem.json
```diff
@@ -22,10 +22,10 @@
         "Linux"
     ],
     "icon_path": "preview.png",
-    "requirements": "This gem requires ROS 2 gem and PhysX gem.",
+    "requirements": "This gem requires ROS 2 gem and PhysX5 gem.",
     "dependencies": [
         "ROS2==3.0.0",
-        "PhysX"
+        "PhysX5"
```
AMZN-alexpete, line 28: ditto here, we should update `compatible_engines` as that will let users see in the Project Manager that this gem is not compatible with older versions of the engine

Templates/Ros2ProjectTemplate/Template/Gem/enabled_gems.cmake
```diff
@@ -18,7 +18,7 @@ set(ENABLED_GEMS
     ImGui
     LandscapeCanvas
     LyShine
-    PhysX
```
AMZN-alexpete, line 21: yikes, we should eventually update this template to not have an `enabled_gems.cmake` file any more.  All these gems should be [moved to the `project.json` like the default template](https://github.com/o3de/o3de/blob/development/Templates/DefaultProject/Template/project.json) and several other changes like updating the [EngineFinder.cmake](https://github.com/o3de/o3de/blob/development/Templates/DefaultProject/Template/cmake/EngineFinder.cmake) and whatever [`CMakeLists.txt`](https://github.com/o3de/o3de/blob/development/Templates/DefaultProject/Template/CMakeLists.txt) changes are needed.

repo.json
```diff
@@ -130,7 +130,7 @@
                 "Atom_Feature_Common",
                 "Atom_Component_DebugCamera",
                 "CommonFeaturesAtom",
-                "PhysX",
+                "PhysX5",
```
AMZN-alexpete, line 133: changing the gem data manually will invalidate the hash for the remote object.  If you use the o3de cli to update the repo.json it should update the hash as well, also we should bump the gem version so this change actually appears in the `versions_data` section that way we won't break older versions of the gem.

Gems/WarehouseAutomation/gem.json
AMZN-alexpete: also we should bump the gem version to `1.1.0`

Gems/ROS2/gem.json
AMZN-alexpete: Also we should bump the gem version to `3.1.0`

AMZN-alexpete (commented):
I think the repo.json changes will break existing older versions of the ros2 gem. I left some comments related to how to update it.  Mostly stuff like bumping gem versions, using the cli.  Also, it looks like both those templates are using the older (now removed) `enabled_gems.cmake` file.  Not sure if you want to upgrade those templates in this PR though.

Templates/Ros2FleetRobotTemplate/Template/Registry/physxdebugconfiguration.setreg
spham-amzn: Updated in the main [PR](https://github.com/o3de/o3de/pull/17436/commits/5e02f5ba744a9577e7fa9f256b3faf159bfa13e8)

lumberyard-employee-dm (approved):

repo.json
```diff
@@ -198,7 +247,7 @@
             "display_name": "WarehouseAssets",
             "license": "Apache-2.0 or MIT",
             "license_url": "https://opensource.org/licenses/Apache-2.0",
-            "origin": "The name of the originator or creator",
+            "origin": "RobotecAI",
```
AMZN-alexpete, line 250: Changing this here, instead of in `versions_data` will likely break older versions of the Gem because the O3DE CLI will run a sha check and compare all the gem fields and error if it finds any inconsistencies between the repo.json and what appears in gem.json

AMZN-alexpete (approved):

PR #646, opened 2024-01-09
Add optional subscription of position control to JointsManipulationComponent

przemekpjski:
## What does this PR do?

Solves #442.

Introduces position control interface in `JointsManipulationComponent`.
Uses the `std_msgs/msg/Float64MultiArray` message type.
The number of values in the `Float64MultiArray` must match the number of joints in `JointsManipulationComponent`. Values are mapped to joints by order in which they are stored in the component (ahhh, thus TODO I need to introduce some fixed order).

## How was this PR tested?

Tested using the `ros2 topic pub` command.
Example:
```
ros2 topic pub --once /position_controller/commands  std_msgs/msg/Float64MultiArray "{data: [1.4, 0.4, 0, -1.2, -1.6, -0.1]}"
```

PR #675, opened 2024-03-07
Optional subscription of position control to JointsManipulationComponent

jhanca-robotecai:
## What does this PR do?

This is another implementation of #646, which is currently held and, I believe, the author does not intend to continue. 

The main difference between #646 and this contribution:
- [Refactor ControlSubscriptionHandler to be more generic](https://github.com/o3de/o3de-extras/pull/646/commits/24ddb72a40aeae1c72f90d8f969e28a370f34ff0) commit is reverted as unrelated to this PR.
- `[JointNamePositionPair](https://github.com/o3de/o3de-extras/pull/646/commits/52e5fbfd473968d56441750b55cb3f14aa3ac99c#diff-86d9f81f8394f66b5cddd8ccb95b976d7a97dbb0e55041963952f551a513efdf)` class is removed (`AZStd::pair` is used in this contribution instead)
- The prefabs are updated in this contribution instead of adding a script to update prefabs manually.
- An additional method for parsing the entity and finding all joints is added in this contribution.

What could be still added:
- while some of the prefabs are updated in this contribution, there could be a serialization added to automatically parse the old format - this part is implemented as a PoC

## How was this PR tested?

Level `RoboticManipulation` of `Ros2RoboticManipulationTemplate`:
```
ros2 topic pub --once /position_controller/commands std_msgs/msg/Float64MultiArray "{data: [1.4, 0.4, 0, 1.2, 0.6, 0.5, 0.8, 0.1, 0.5]}"
```


Gems/ROS2/Code/Source/Manipulation/JointsManipulationComponent.cpp
```diff
@@ -449,4 +470,31 @@ namespace ROS2
         }
         MoveToSetPositions(deltaTime);
     }
+
+    void JointsManipulationComponent::ProcessPositionControlMessage(const std_msgs::msg::Float64MultiArray& message)
+    {
+        if (message.data.size() != m_initialPositions.size())
```
michalpelka, line 476: I am not fan of this design. 
Let imagine that we have an articulated contraption with four legs. 
And only part of the system is controlled by control subscription.
In that situation  :
- we can have multiple topics for different segments of our contraption.
- we will have different number of initial joints position than subscription.
IHMO we are creating here a lot inflexibility.
Consider:
moving subscription to separate component with own topic and joint name configuration. I think it would be more flexible.

jhanca-robotecai:
> I would consider to separate subscription to sidecar component for flexibility. Especially underlying API totally allows that.

Subscription is moved to a separate component, following your suggestion. The initial position can be still set by the default component, while the new component responds to the topic calls and uses the bus to move the joint accordingly. 

The change was tested on the same level, with the new component added to `panda_link0` entity of `PandaRobot`:
Screenshot from 2024-03-08 14-54-58

The system responds to the call:
```
ros2 topic pub --once /position_controller/commands std_msgs/msg/Float64MultiArray "{data: [1]}"
```

jhanca-robotecai:
Button label was changed from `Find joints` to `Find all joints`

Gems/ROS2/Code/Source/Manipulation/JointsPositionsComponent.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    void JointsPositionsComponent::GetRequiredServices(AZ::ComponentDescriptor::DependencyArrayType& required)
+    {
+        required.push_back(AZ_CRC_CE("ROS2Frame"));
+        required.push_back(AZ_CRC_CE("JointsControllerService"));
+    }
+
+    void JointsPositionsComponent::Activate()
+    {
+        if (m_jointPositionsSubscriberConfiguration.m_subscribe)
```
michalpelka, line 37: Since it is a different component I see no need for this flag

michalpelka (approved):

pawelbudziszewski (approved):

PR #697, opened 2024-05-08
Fix issue with position control index processing

nbbrooks:
## What does this PR do?

This fixes an issue where the joints positions command at the first index is sent to all joints, instead of their respective commands.

I do not know the preferred convention for dual range loop iteration in C++17 (which I believe is what O3DE targets) so feel free to make changes!

## How was this PR tested?
1. Build an engine from the `92df6b` PhysX Gem Split
2. Build a Ros2RoboticManipulationTemplate project from this PR
3. Added a JointsPositionsEditorComponent with J1-7  to panda_link0
image
4. Command `ros2 topic pub --once /position_controller/commands std_msgs/msg/Float64MultiArray "{data: [0.1, -0.69, 0.1, -2.26, 0.1, 1.47, 0.89]}"` (this is 0.1 rad from the default starting position) from another terminal with the simulator running

If you build the project from upstream, the arm will attempt to move all joints to `0.1`

## Other notes
As a side note, servoing the arm using relative positions (e.g. with moveit_servo) has some issues with the default Articulation link settings for the Panda. They are quite aggressive motions, which ends up with a sort of bang-bang feedback loop that makes servoing unstable.

### DIFF
--- a/Gems/ROS2/Code/Source/Manipulation/JointsPositionsComponent.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/JointsPositionsComponent.cpp
@@ -94,19 +94,19 @@ namespace ROS2
             return;
         }
 
-        auto commandIter = message.data.cbegin();
-        for (const auto& jointName : m_jointNames)
+        for(int i=0; i<message.data.size(); i++)
         {
             AZ::Outcome<void, AZStd::string> result;
             JointsManipulationRequestBus::EventResult(
-                result, m_rootOfArticulation, &JointsManipulationRequests::MoveJointToPosition, jointName, *commandIter);
+                result, m_rootOfArticulation, &JointsManipulationRequests::MoveJointToPosition, m_jointNames[i], message.data[i]);
             if (!result.IsSuccess())
             {
                 AZ_Error(
                     "JointsPositionsComponent",
                     result,
-                    "PositionController: command failed for joint %s: ",
-                    jointName.c_str(),
+                    "PositionController: failed for joint %s and command %d: ",
+                    m_jointNames[i].c_str(),
+                    message.data[i],
                     result.GetError().c_str());
             }
         }


adamdbrw (approved):

michalpelka (approved):

michalpelka (approved):

michalpelka:
Thank you for changing @nbbrooks. I would like to merge, but yo need to setup DCO https://github.com/o3de/o3de-extras/pull/697/checks?check_run_id=24734677415
please add a signoff message with your email address to your commit and force push.
It should look like that:
```
Signed-off-by: Michał Pełka <michal.pelka@robotec.ai>
```

nbbrooks:
Whoops, should be addressed now!

spham-amzn (approved):

Gems/ROS2/Code/Source/Manipulation/JointsPositionsComponent.cpp
```diff
@@ -94,19 +94,19 @@ namespace ROS2
...
         }
 
-        auto commandIter = message.data.cbegin();
-        for (const auto& jointName : m_jointNames)
+        for(int i=0; i<message.data.size(); i++)
         {
             AZ::Outcome<void, AZStd::string> result;
             JointsManipulationRequestBus::EventResult(
-                result, m_rootOfArticulation, &JointsManipulationRequests::MoveJointToPosition, jointName, *commandIter);
+                result, m_rootOfArticulation, &JointsManipulationRequests::MoveJointToPosition, m_jointNames[i], message.data[i]);
```
spham-amzn, line 101: Actually, it looks like the error is in the original logic. `commandIter` needs to be incremented as well during the traversal of the `m_jointNames` list. (from the original line)
```
JointsManipulationRequestBus::EventResult(
                result, m_rootOfArticulation, &JointsManipulationRequests::MoveJointToPosition, jointName, *commandIter);
                commandIter++;
```

Gems/ROS2/Code/Source/Manipulation/JointsPositionsComponent.cpp
```diff
@@ -94,19 +94,19 @@ namespace ROS2
             return;
         }
 
-        auto commandIter = message.data.cbegin();
-        for (const auto& jointName : m_jointNames)
+        for(int i=0; i<message.data.size(); i++)
```
spham-amzn, line 97: Actually, it looks like the error is in the original logic. commandIter needs to be incremented as well during the traversal of the m_jointNames list. (from the original line). You could also do
```
for (auto jointNameIter = m_jointNames.cbegin(), auto commandIter = message.data.cbegin();
       jointNameIter != m_jointNames..cend() && commandIter != message.data..cend();
       ++jointNameIter, ++commandIter)
```
and update the `m_jointNames[i]` with *jointNameIter and `message.data[i]` with `*commandIter`

PR #698, opened 2024-05-20
Redo Ros2ProjectTemplate

knicked:
## What does this PR do?

It replaces the assets used in Ros2ProjectTemplate with those from the WarehouseAssets gem.

preview
*new template's preview*

Additionally:
- it fixes an issue where the template's project gem was not enabled, causing components not to load.
- it fixes some include issues that caused templates to fail to build if the project's name contained characters that would be sanitized in C++ code.

These fixes have been applied to all templates.

## How was this PR tested?

Projects were created using modified templates, built and ran successfully.


michalpelka:
Project works as expected, but Ros2ProjectTemplate should be simple and light. This is not a case here.

jhanca-robotecai:
Cherry-picked the crucial changes to #723

PR #700, opened 2024-05-29
repo.json updates for Ros2RoboticManipulationTemplate

spham-amzn:
## What does this PR do?

Recent updates to the `Ros2RoboticManipulationTemplate` was not reflected in repo.json, which causes an error trying to download the template from Project Manager that is built from the development branch. This update fixes that using the `o3de.sh edit-repo-properties` command line for the template.

## How was this PR tested?
Tested locally

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -3691,10 +3691,6 @@
                     "isTemplated": "false"
                 },
                 {
-                    "file": "Assets/Factory/ConveyorLine/ConveyorBelt_effector_short.fbx.assetinfo",
-                    "isTemplated": "false"
-                },
-                {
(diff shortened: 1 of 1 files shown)


amzn-changml:
Seems reasonable. We also have to update this repo file with the same change: https://github.com/o3de/canonical.o3de.org/blob/main/src/repo.json

amzn-changml (approved):

spham-amzn:
> Seems reasonable. We also have to update this repo file with the same change: https://github.com/o3de/canonical.o3de.org/blob/main/src/repo.json

Yep, that is the next step

Issue #705, opened 2024-06-13
Editor with ROS2 gem enabled crash on simulate


yosagi:
**Describe the bug**
Editor with ROS2 gem enabled crash on simulate.

**Assets required**
Any project with ROS2 gem added.

**Steps to reproduce**
1. Create a project with default template.
2. Add ROS2 gem in "configure gems".
3. Build the project.
4. Open the project in a Editor.
5. Create or open any level.
6. Press simulate button on top right corner on the Editor.
7. The editor crashes with SIGSEGV.

**Expected behavior**
No crash. 

**Actual behavior**
Crash.

**Screenshots/Video**
The following is a first few lines of backtrace on gdb session.
```
Module: Success!
[New Thread 0x7ffdb07f8640 (LWP 295626)]
[New Thread 0x7ffdb05f7640 (LWP 295627)]
[Detaching after fork from child process 295628]
[New Thread 0x7ffd7ae4b640 (LWP 295630)]

Thread 1 "Editor" received signal SIGSEGV, Segmentation fault.
ROS2::ROS2SystemComponent::Deactivate (this=0x555558d40250) at /mnt/data/yos/work/upstream/o3de-extras/Gems/ROS2/Code/Source/ROS2SystemComponent.cpp:172
warning: Source file is more recent than executable.
172             m_simulationClock->Deactivate();
(gdb) bt
#0  ROS2::ROS2SystemComponent::Deactivate (this=0x555558d40250) at /mnt/data/yos/work/upstream/o3de-extras/Gems/ROS2/Code/Source/ROS2SystemComponent.cpp:172
#1  0x00007fff010184ba in non-virtual thunk to ROS2::ROS2EditorSystemComponent::OnStopPlayInEditor() ()
    at /mnt/data/yos/work/upstream/o3de-extras/Gems/ROS2/Code/Source/ROS2EditorSystemComponent.cpp:66
[...]

PR #708, opened 2024-06-17
Fix ROS 2 System component to work with Simulate in Editor

michalpelka:
## What does this PR do?

This PR prevents null pointer dereference when the ROS2 System component is deactivated at the start of the simulation in the editor.

#705 
## How was this PR tested?

Run against manipulation template with o3de/stabilization/2409

Note:
This is PR is crash-fix. ROS 2 functionality is partially usable in simulation mode. The ROS 2, sensors, and others are paused on the Editor's window loses focus. However, I think it is unrelated to ROS 2 and is some feature or bug in o3de.


adamdbrw (approved):
Good fix!

spham-amzn (approved):

PR #715, opened 2024-06-25
Update ROS 2 sensors base implementation

michalpelka:
- Add and tested `SensorConfigurationRequestBus`
- Added helper functions to discover ROS 2 Sensors in runtime
- Moved ROS 2 Sensors types to `ROS2SensorTypesIds.h`

## What does this PR do?

It is a fix of an existing feature - a bus `SensorConfigurationRequestBus` that allows adjusting the behavior of ROS 2 sensors.
It uses the ComponentEntityId pair as the key.

Also, a number of things were added or adjusted:
- Add and tested `SensorConfigurationRequestBus`
- Added helper functions to discover ROS 2 Sensors in runtime
- Moved ROS 2 Sensors types to `ROS2SensorTypesIds.h`

## How was this PR tested?

It is only API, that 3rd party needs to integrate. As test, I've implemented a Gem to test it:
https://github.com/RobotecAI/robotec-o3de-tools/pull/38
image

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
+++ b/Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
@@ -197,7 +197,7 @@ namespace ROS2
         typename EventSourceT::AdaptedEventType m_sensorAdaptedEvent{};
 
         float m_adaptedFrequency{ 30.0f }; ///< Adapted frequency value.
-        float m_lastDeltaTime{ 0.0f }; ///< Effective frequency value.
+        float m_lastDeltaTime{ 0.0f }; ///< Last difference in time between adapted events, used to compute effective frequency value.
         float m_adaptedDeltaTime{ 0.0f }; ///< Accumulator for calculating adapted delta time.
         int m_tickCounter{ 0 }; ///< Internal counter for controlling adapter frequency.
     };
--- a/Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
+++ b/Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
@@ -107,6 +107,7 @@ namespace ROS2
                         return;
                     }
 
+                    m_lastDeltaTime =  m_adaptedDeltaTime;
                     m_sensorAdaptedEvent.Signal(m_adaptedDeltaTime, AZStd::forward<decltype(args)>(args)...);
                     m_adaptedDeltaTime = 0.0f;
                 });
@@ -132,6 +133,17 @@ namespace ROS2
         void SetFrequency(float adaptedFrequency)
         {
             m_adaptedFrequency = adaptedFrequency;
+        }
+
+        //! Gets adapter working frequency, based on the last obtained delta time between adapted events.
+        //! If the adapter has not been started yet, the frequency will be zero.
+        [[nodiscard]] float GetEffectiveFrequency() const
+        {
+            if (m_lastDeltaTime == 0.0f)
+            {
+                return 0.0f;
+            }
+            return 1.0f / m_lastDeltaTime;
         }
 
         //! Connects given event handler to source event (ROS2::SensorEventSource). That event is signalled regardless of adapted frequency
@@ -185,6 +197,7 @@ namespace ROS2
         typename EventSourceT::AdaptedEventType m_sensorAdaptedEvent{};
 
         float m_adaptedFrequency{ 30.0f }; ///< Adapted frequency value.
+        float m_lastDeltaTime{ 0.0f }; ///< Effective frequency value.
         float m_adaptedDeltaTime{ 0.0f }; ///< Accumulator for calculating adapted delta time.
         int m_tickCounter{ 0 }; ///< Internal counter for controlling adapter frequency.
     };
--- a/Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
+++ b/Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
@@ -69,7 +69,7 @@ namespace ROS2
             return m_sensorConfiguration;
         }
 
-        void EnableSensor(bool sensorEnabled) override
+        void SetSensorEnabled(bool sensorEnabled) override
         {
             if (sensorEnabled)
             {
@@ -81,12 +81,12 @@ namespace ROS2
             }
         }
 
-        void EnablePublishing(bool publishingEnabled) override
+        void SetPublishingEnabled(bool publishingEnabled) override
         {
             m_sensorConfiguration.m_publishingEnabled = publishingEnabled;
         }
 
-        void EnableVisualize(bool visualizeEnabled) override
+        void SetVisualizeEnabled(bool visualizeEnabled) override
         {
             m_sensorConfiguration.m_visualize = visualizeEnabled;
         }
--- a/Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
+++ b/Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
@@ -12,6 +12,7 @@
 #include <AzCore/Serialization/EditContext.h>
 #include <ROS2/Frame/ROS2FrameComponent.h>
 #include <ROS2/ROS2GemUtilities.h>
+#include <ROS2/ROS2SensorTypesIds.h>
 #include <ROS2/Sensor/Events/EventSourceAdapter.h>
 #include <ROS2/Sensor/SensorConfiguration.h>
 #include <ROS2/Sensor/SensorConfigurationRequestBus.h>
@@ -68,19 +69,50 @@ namespace ROS2
             return m_sensorConfiguration;
         }
 
+        void EnableSensor(bool sensorEnabled) override
+        {
+            if (sensorEnabled)
+            {
+                m_eventSourceAdapter.Start();
+            }
+            else
+            {
+                m_eventSourceAdapter.Stop();
+            }
+        }
+
         void EnablePublishing(bool publishingEnabled) override
         {
             m_sensorConfiguration.m_publishingEnabled = publishingEnabled;
+        }
+
+        void EnableVisualize(bool visualizeEnabled) override
+        {
+            m_sensorConfiguration.m_visualize = visualizeEnabled;
+        }
+
+        float GetEffectiveFrequency() const override
+        {
+            return m_eventSourceAdapter.GetEffectiveFrequency();
+        }
+
+        void SetDesiredFrequency(float frequency) override
+        {
+            m_sensorConfiguration.m_frequency = frequency;
+            m_eventSourceAdapter.SetFrequency(frequency);
         }
 
         virtual ~ROS2SensorComponentBase() = default;
 
         void Activate() override
         {
+            AZ::EntityComponentIdPair entityComponentIdPair(GetEntityId(), GetId());
+            SensorConfigurationRequestBus::Handler::BusConnect(entityComponentIdPair);
         }
 
         void Deactivate() override
         {
+            SensorConfigurationRequestBus::Handler::BusDisconnect();
         }
 
     protected:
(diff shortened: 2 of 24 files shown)


Gems/ROS2/Code/Source/Sensor/SensorHelpers.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace ROS2
+{
+
+    AZStd::vector<AZ::EntityComponentIdPair> GetSensorsForEntity(const AZ::EntityId& entityId)
+    {
+        AZStd::vector<AZ::EntityComponentIdPair> sensors;
+        AZ::Entity* entity = AZ::Interface<AZ::ComponentApplicationRequests>::Get()->FindEntity(entityId);
+        AZ_Assert(entity, "Entity not found");
+        if (entity)
```
MateuszWasilewski, line 22:
```suggestion
        if (AZ::Entity* entity = AZ::Interface<AZ::ComponentApplicationRequests>::Get()->FindEntity(entityId))
```
I think in case of invalid `entityId` we shouldn't both assert and check, only check and treat it as valid input

Gems/ROS2/Code/Include/ROS2/Sensor/SensorHelper.h
```diff
@@ -0,0 +1,30 @@
...
+{
+    //! Function to check if the component is a sensor component
+    //! @param component Component to check
+    //! @return True if the component is a sensor component
+    bool IsComponentROS2Sensor(const AZ::Component* component);
+
+    //! Function to get Ids of all sensors attached to the entity
+    //! @param entityId Id of the entity
+    //! @return Vector of EntityComponentIdPair of all sensors attached to the entity
+    AZStd::vector<AZ::EntityComponentIdPair> GetSensorsForEntity(const AZ::EntityId& entityId);
```
MateuszWasilewski, line 22: I think `AZStd::vector<AZ::ComponentId>` would be better return type from this API, because each component will belong to the same entity, whose entityId is already known by the caller

Gems/ROS2/Code/Include/ROS2/Sensor/SensorHelper.h
```diff
@@ -0,0 +1,30 @@
...
+
+    //! Function to get Ids of all sensors attached to the entity
+    //! @param entityId Id of the entity
+    //! @param sensorType Type of the sensor, see @file ROS2SensorTypesIds.h
+    //! @return Vector of EntityComponentIdPair of all sensors attached to the entity
+    AZStd::vector<AZ::EntityComponentIdPair> GetSensorsOfType(const AZ::EntityId& entityId, const AZ::Uuid& sensorType);
```
MateuszWasilewski, line 28: The same as above.

Gems/ROS2/Code/Include/ROS2/Sensor/SensorConfigurationRequestBus.h
```diff
@@ -7,22 +7,40 @@
  */
 #pragma once
 
+#include <AzCore/Component/ComponentBus.h>
 #include <AzCore/Component/EntityId.h>
 #include <AzCore/EBus/EBus.h>
 #include <AzCore/Interface/Interface.h>
 #include <ROS2/Sensor/SensorConfiguration.h>
-
```
MateuszWasilewski, line 14: nit: is this line removal intended?

Gems/ROS2/Code/Source/Sensor/SensorHelpers.cpp
```diff
@@ -0,0 +1,71 @@
...
+        {
+            auto components = entity->GetComponents();
+            for (auto component : components)
+            {
+                AZ_Assert(component, "Component not found");
+                const auto ptr = component->RTTI_IsTypeOf(sensorType);
+                if (ptr)
```
MateuszWasilewski, line 48:
```suggestion
                if (component->RTTI_IsTypeOf(sensorType))
```
RTTI_IsTypeOf returns a bool, not a ptr

Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
```diff
@@ -68,19 +69,50 @@ namespace ROS2
             return m_sensorConfiguration;
         }
 
+        void EnableSensor(bool sensorEnabled) override
```
MateuszWasilewski, line 72: I suggest holding the same convention for all function in this header
Ex.
GetSensorConfiguration
SetSensorEnabled
SetPublishingEnabled
SetVisualizeEnabled
GetEffectiveFrequency
SetDesiredFrequency

MateuszWasilewski (commented):
Code looks really great and brings additional features to the gem.
Left minor suggestions.

Gems/ROS2/Code/Source/Sensor/SensorHelpers.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    AZStd::vector<AZ::EntityComponentIdPair> GetSensorsForEntity(const AZ::EntityId& entityId)
+    {
+        AZStd::vector<AZ::EntityComponentIdPair> sensors;
+        AZ::Entity* entity = AZ::Interface<AZ::ComponentApplicationRequests>::Get()->FindEntity(entityId);
+        AZ_Assert(entity, "Entity not found");
+        if (entity)
+        {
+            auto components = entity->GetComponents();
+            for (auto component : components)
```
patrykantosz, line 25:
```suggestion
            for (const auto* component : components)
```

Gems/ROS2/Code/Source/Sensor/SensorHelpers.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    AZStd::vector<AZ::EntityComponentIdPair> GetSensorsOfType(const AZ::EntityId& entityId, const AZ::Uuid& sensorType)
+    {
+        AZStd::vector<AZ::EntityComponentIdPair> sensors;
+        AZ::Entity* entity = AZ::Interface<AZ::ComponentApplicationRequests>::Get()->FindEntity(entityId);
+        AZ_Assert(entity, "Entity not found");
+        if (entity)
+        {
+            auto components = entity->GetComponents();
+            for (auto component : components)
```
patrykantosz, line 44:
```suggestion
            for (const auto* component : components)
```

patrykantosz (commented):
Great work! Just minor suggestions from my side.

Gems/ROS2/Code/Source/Sensor/SensorHelpers.cpp
michalpelka: I think both should stay.
- Assert causes breakpoint during debug
- if - prevents crash in the product.
Using only assert is not a great idea IHMO.

Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
```diff
@@ -68,19 +69,50 @@ namespace ROS2
...
+
+        float GetEffectiveFrequency() const override
+        {
+            return m_eventSourceAdapter.GetEffectiveFrequency();
+        }
+
+        void SetDesiredFrequency(float frequency) override
+        {
+            m_sensorConfiguration.m_frequency = frequency;
+            m_eventSourceAdapter.SetFrequency(frequency);
```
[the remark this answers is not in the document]
michalpelka: I want to prevent division by zero only.

Gems/ROS2/Code/Include/ROS2/Sensor/Events/EventSourceAdapter.h
```diff
@@ -134,6 +135,17 @@ namespace ROS2
             m_adaptedFrequency = adaptedFrequency;
         }
 
+        //! Gets adapter working frequency, based on the last obtained delta time between adapted events.
+        //! If the adapter has not been started yet, the frequency will be zero.
+        [[nodiscard]] float GetEffectiveFrequency() const
+        {
+            if (m_lastDeltaTime == 0.0f)
```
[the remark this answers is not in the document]
michalpelka: I want to prevent division by zero only.

Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
michalpelka: Agree,

Gems/ROS2/Code/Include/ROS2/Sensor/SensorHelper.h
michalpelka: Disagree, since I want to have a type that can be passed directly to SensorConfigurationRequestBus. Using only ComponentId would cause some boilerplate code to be needed by API consumer.

Gems/ROS2/Code/Include/ROS2/Sensor/SensorHelper.h
michalpelka: Disagree, since I want to have a type that can be passed directly to SensorConfigurationRequestBus. Using only ComponentId would cause some boilerplate code to be needed by API consumer.

michalpelka (commented):
Thanks for review @pijaro, @MateuszWasilewski @patrykantosz 
I gave argumentation against some suggestions - please @MateuszWasilewski, @pijaro  close if you agree or continue argumnent.

Gems/ROS2/Code/Include/ROS2/Sensor/ROS2SensorComponentBase.h
michalpelka: @patrykantosz could you refactor my PR following @MateuszWasilewski suggestion?

patrykantosz: @michalpelka Sure.

Gems/ROS2/Code/Source/Sensor/SensorHelpers.cpp
MateuszWasilewski: I think the best solution would be both `AZ_Warning` and `if` for user can set breakpoint for debug. If we put `AZ_Assert` here, we can't skip it when debugging something else.

patrykantosz:
@pijaro @MateuszWasilewski Could you please check if now everything is as it should be?

adamdbrw:
Ping me tomorrow to review this please

Gems/ROS2/Code/Source/Sensor/SensorHelpers.cpp
michalpelka: Ok. Agree

Gems/ROS2/Code/Include/ROS2/Sensor/SensorHelper.h
jhanca-robotecai: I second @MateuszWasilewski in this matter. 
I understand your point in making interfaces easier to use with the bus, but it makes the API weird. When asking for the list of sensors for the given entity, I expect nothing but the list of sensors. I know the _id_ of the entity at this point and I don't need to get it back duplicated with every sensor _id_.

patrykantosz:
@adamdbrw Ping. As you wished.

pijaro:
Tested with o3de development, profile config, project-centric, and with https://github.com/RobotecAI/robotec-o3de-tools/pull/38 - works as expected.

michalpelka:
Since I am author, I cannot accept officially, but I support changes by @patrykantosz, thanks. For me it can be merged.

pijaro (approved):

Gems/ROS2/Code/Include/ROS2/Sensor/SensorHelper.h
pijaro: I agree that a raw Component vector would be a better fit, but in my opinion, the proposed approach is also acceptable. 

A vector of components is more semantically correct, but this one has some valid quality-of-life improvement.

Gems/ROS2/Code/Include/ROS2/Sensor/SensorHelper.h
adamdbrw: I think returning components only is correct, plus a utility function to convert these to vector of entitycomponentidpair - what you mean as boilerplate would amount to replacing x with f(x) to pass to the bus.

adamdbrw (approved):
Suggestions to apply, mostly small typo fixes etc.
I don't want the comment about return vector to stop the merge, it can be applied in a separate PR

amzn-changml:
Confirmed AR passed. Merging this now

michalpelka:
Thanks @amzn-changml for merging. Kudos to @patrykantosz for applying helpful comments from @adamdbrw and @pijaro !

PR #720, opened 2024-07-08
Lidar: Internal storage changed from Euler to quaternion

### LINKED ISSUE #488: Lidars keeps rotation data as Euler angles
Lidars keeps data about raycasts as Euler angles, only converting them into quaternions when making raycast requests. Euler angles should only be used for used input, so the data should be converted much earlier.

c00kiepreferences:
## What does this PR do?

Closes #488 

Now Lidar converts data about raycast angles right after receiving them as input, instead of converting them only during performing raycast requests.

## How was this PR tested?

Tested on basic ROS2 project template using rviz2. Outputs correspond with previous version both for 2D and 3D Lidar components.

### DIFF
--- a/Gems/ROS2/Code/Source/Lidar/LidarTemplateUtils.h
+++ b/Gems/ROS2/Code/Source/Lidar/LidarTemplateUtils.h
@@ -43,6 +43,7 @@ namespace ROS2
         //! @param rotations Rotations as Euler angles in radians to compute directions from.
         //! @param rootRotation Root rotation as Euler angles in radians.
         //! @return Ray directions constructed by transforming an X axis unit vector by the provided rotations.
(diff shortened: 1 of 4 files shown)


Gems/ROS2/Code/Source/Lidar/LidarTemplateUtils.h
```diff
@@ -43,6 +43,7 @@ namespace ROS2
         //! @param rotations Rotations as Euler angles in radians to compute directions from.
         //! @param rootRotation Root rotation as Euler angles in radians.
```
jhanca-robotecai, line 44: You changed the parameter, but you missed the change in the documentation, please fix it.

c00kiepreferences: Done.

jhanca-robotecai (approved):
Thanks for the contribution!

adamdbrw (approved):

michalpelka (approved):

PR #724, opened 2024-07-08
Lidar: Fixed data acquisition while publishing disabled

### LINKED ISSUE #462: Sensor visualisation doesn't work with publishing disabled
Currently, disabling publishing completely disables data acquisition in the default implementations in ROS2SensorComponent. However, some sensors need this data to draw their visualisations (eg. lidar). This causes the sensors to draw them improperly, or not at all.
I suggest splitting FrequencyTick into 2 separate methods, "acquire data" and "publish". AcquireData would be called either every frame, in the case visualisation in enabled, or every time we need to publish, in the case it's not.

c00kiepreferences:
## What does this PR do?

Closes #462

Now Lidar acquires data about raycast results even when publishing is disabled, making it possible to visualise results independently from publishing.

Even though issue #462 suggests otherwise, problem with visualisation seems to only apply to Lidar components and further corrections are unnecessary.

## How was this PR tested?

Tested on basic ROS2 project template using rviz2. Visualisation works independently from publishing (and vice-versa).

### DIFF
--- a/Gems/ROS2/Code/Source/Lidar/ROS2Lidar2DSensorComponent.cpp
+++ b/Gems/ROS2/Code/Source/Lidar/ROS2Lidar2DSensorComponent.cpp
@@ -79,10 +79,6 @@ namespace ROS2
             m_sensorConfiguration.m_frequency,
             [this]([[maybe_unused]] auto&&... args)
             {
-                if (!m_sensorConfiguration.m_publishingEnabled)
-                {
-                    return;
-                }
                 FrequencyTick();
             },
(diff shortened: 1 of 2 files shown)


adamdbrw (approved):
Looks good and seems to fix two issues. Good work!

jhanca-robotecai (approved):
Fixes the issue!

amzn-changml:
This [passed AR](https://jenkins.build.o3de.org/blue/organizations/jenkins/o3de-extras/detail/PR-724/1/pipeline/), so will manually merge this now.

Issue #736, opened 2024-07-25
[ROS2] Feature: Improved Lidar Result retrieval and publishing


alek-kam-robotec-ai:
**Is your feature request related to a problem? Please describe.**
This issue relates to the growing complexity of the Lidar result data. The current implementation of both lidar result retrieval and publishing is outdated and broken (see https://github.com/o3de/o3de-extras/issues/719). It also makes it hard to add new, optional features (see https://github.com/o3de/o3de-extras/pull/713 and https://github.com/o3de/o3de-extras/pull/737).

**Describe the solution you'd like**
A new design of the raycast results handling should be designed and implemented.

Issue #719, opened 2024-07-05
[ROS 2]  Lidar Sensor does not allow for raycaster-side publication. 


alek-kam-robotec-ai:
**Describe the bug**
The pointcloud is always published in the ROS2 Lidar Sensor regardless of the configuration.
This is due to the fact that the raycasters are queried about their capability to publish before their configuration is finalized, namely
the `ConfigurePointCloudPublisher` is used after a call to `CanHandlePublishing` (methods of the LidarRequests interface) which requires the lidar to be already configured.

**Expected behavior**
The pointcloud publisher should be determined based on whether or not the Lidar implementation supports publishing under current lidar configuration.

**Found in Branch**
All branches. Was introduced by [commit](https://github.com/o3de/o3de-extras/commit/b5ce30ecd0900daa9c0d7914d47da844fe4d4912).

alek-kam-robotec-ai:
@adamdbrw @michalpelka 
I've spoken with one of the RGL team members and we suggest that this is patched after the instance segmentation (#713) and intensity (under development) features are merged.

jhanca-robotecai:
Isn't this one related to: https://github.com/o3de/o3de-extras/issues/462
and being fixed in https://github.com/o3de/o3de-extras/pull/724
?

alek-kam-robotec-ai:
No, it is not. As I've described, it is related to the raycaster configuration and query order.

alek-kam-robotec-ai:
Additionally, the same pull request hard-sets requested result types to both points and distances irrespective of the Lidar Component, which is not the intended functionality.

PR #737, opened 2024-07-25
[ROS2] Add support for lidar intensity

alek-kam-robotec-ai:
## What does this PR do?

This pr introduces lidar intensity data support for the ROS2 Lidar Sensor Component (does not implement this feature for the ROS2 Lidar 2D Sensor Component) . It also modifies point-cloud publishing as intensity is an optional feature and may not be provided by every lidar implementation (eg. Scene queries). For more information about the intensity capabilities and component support see the [RGL gem intensity support pull request](https://github.com/RobotecAI/o3de-rgl-gem/pull/33).

_**Note:**_ _Due to https://github.com/o3de/o3de-extras/issues/719 and future [instance segmentation](https://github.com/o3de/o3de-extras/pull/713) integration this part of the ROS2 Lidar Sensor will soon have to be updated (see https://github.com/o3de/o3de-extras/issues/736)._

## How was this PR tested?

This feature was tested on Mesh Components with and without the Material Components and on Terrain Components. Since this feature is only supported by the [RGL gem](https://github.com/RobotecAI/o3de-rgl-gem), it was tested with a custom RGL gem test project (and a different closed-source project).

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
@@ -88,6 +88,7 @@ namespace ROS2
     {
         Points = (1 << 0), //!< return 3D point coordinates
         Ranges = (1 << 1), //!< return array of distances
+        Intensity = (1 << 2), //!< return intensity data
     };
 
     //! Bitwise operators for RaycastResultFlags
@@ -97,6 +98,7 @@ namespace ROS2
     {
         AZStd::vector<AZ::Vector3> m_points;
         AZStd::vector<float> m_ranges;
+        AZStd::vector<float> m_intensity;
     };
 
     //! Interface class that allows for communication with a single Lidar instance.
--- a/Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
+++ b/Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
@@ -88,7 +88,7 @@ namespace ROS2
     {
         Points = (1 << 0), //!< return 3D point coordinates
         Ranges = (1 << 1), //!< return array of distances
-        Intensity = (1 << 2), //!< return intensity data
+        Intensities = (1 << 2), //!< return intensity data
     };
 
     //! Bitwise operators for RaycastResultFlags
@@ -98,7 +98,7 @@ namespace ROS2
     {
         AZStd::vector<AZ::Vector3> m_points;
         AZStd::vector<float> m_ranges;
-        AZStd::vector<float> m_intensity;
+        AZStd::vector<float> m_intensities;
     };
 
     //! Interface class that allows for communication with a single Lidar instance.
--- a/Gems/ROS2/Code/Source/Lidar/LidarCore.cpp
+++ b/Gems/ROS2/Code/Source/Lidar/LidarCore.cpp
@@ -77,7 +77,7 @@ namespace ROS2
         RaycastResultFlags requestedFlags = RaycastResultFlags::Ranges | RaycastResultFlags::Points;
         if (m_lidarConfiguration.m_lidarSystemFeatures & LidarSystemFeatures::Intensity)
         {
-            requestedFlags |= RaycastResultFlags::Intensity;
+            requestedFlags |= RaycastResultFlags::Intensities;
         }
 
         LidarRaycasterRequestBus::Event(m_lidarRaycasterId, &LidarRaycasterRequestBus::Events::ConfigureRaycastResultFlags, requestedFlags);
--- a/Gems/ROS2/Code/Source/Lidar/LidarCore.cpp
+++ b/Gems/ROS2/Code/Source/Lidar/LidarCore.cpp
@@ -75,6 +75,10 @@ namespace ROS2
         }
 
         RaycastResultFlags requestedFlags = RaycastResultFlags::Ranges | RaycastResultFlags::Points;
+        if (m_lidarConfiguration.m_lidarSystemFeatures & LidarSystemFeatures::Intensity)
+        {
+            requestedFlags |= RaycastResultFlags::Intensity;
+        }
 
         LidarRaycasterRequestBus::Event(m_lidarRaycasterId, &LidarRaycasterRequestBus::Events::ConfigureRaycastResultFlags, requestedFlags);
 
@@ -161,8 +165,10 @@ namespace ROS2
         return m_lidarRaycasterId;
     }
 
-    RaycastResult LidarCore::PerformRaycast()
+    const RaycastResult& LidarCore::PerformRaycast()
     {
+        static const RaycastResult EmptyResults{};
+
         AZ::Entity* entity = nullptr;
         AZ::ComponentApplicationBus::BroadcastResult(entity, &AZ::ComponentApplicationRequests::FindEntity, m_entityId);
         const auto entityTransform = entity->FindComponent<AzFramework::TransformComponent>();
@@ -172,7 +178,7 @@ namespace ROS2
         if (m_lastScanResults.m_points.empty())
         {
             AZ_TracePrintf("Lidar Sensor Component", "No results from raycast\n");
-            return RaycastResult();
+            return EmptyResults;
         }
         return m_lastScanResults;
     }
--- a/Gems/ROS2/Code/Source/Lidar/LidarCore.h
+++ b/Gems/ROS2/Code/Source/Lidar/LidarCore.h
@@ -37,7 +37,7 @@ namespace ROS2
 
         //! Perform a raycast.
         //! @return Results of the raycast.
-        RaycastResult PerformRaycast();
+        const RaycastResult& PerformRaycast();
         //! Visualize the results of the last performed raycast.
         void VisualizeResults() const;
 
--- a/Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
+++ b/Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
@@ -129,8 +129,6 @@ namespace ROS2
 
     void ROS2LidarSensorComponent::FrequencyTick()
     {
-        auto entityTransform = GetEntity()->FindComponent<AzFramework::TransformComponent>();
-
         if (m_canRaycasterPublish && m_sensorConfiguration.m_publishingEnabled)
         {
             const builtin_interfaces::msg::Time timestamp = ROS2Interface::Get()->GetROSTimestamp();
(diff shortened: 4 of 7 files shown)


Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +138,73 @@ namespace ROS2
...
+        for (int i = 0; i < numPointFields; i++)
         {
             sensor_msgs::msg::PointField pf;
             pf.name = point_field_names[i];
-            pf.offset = i * 4;
+            pf.offset = i * sizeof(float);
+            if (i == point_field_names.size() - 1U)
+            { // AZ::Vector3 may have padding.
+                pf.offset = sizeof(AZ::Vector3);
+            }
```
PawelLiberadzki, line 174: `sizeof(AZ::Vector3)` returns 16 so its true. However, I would just pack these coordinates without additional space (so sizeof(float) here) and then copy them accordingly. Please take a look at #713.

Note that you would have to change `message.point_step` respectively - line 161 above.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +138,73 @@ namespace ROS2
...
+            {
+                float intensity = results.m_intensity[pointIdx];
+                const auto intensityBytes = reinterpret_cast<const uint8_t*>(&intensity);
+                m_rosMessageData.insert(m_rosMessageData.end(), intensityBytes, intensityBytes + sizeof(float));
+            }
+        }
+
         AZ_Assert(message.row_step * message.height == sizeInBytes, "Inconsistency in the size of point cloud data");
-        memcpy(message.data.data(), lastScanResults.m_points.data(), sizeInBytes);
+        memcpy(message.data.data(), m_rosMessageData.data(), sizeInBytes);
```
PawelLiberadzki, line 207: This looks like unnecessary step - you copy data to `m_rosMessageData` and then to [message.data](url) instead of omitting this step. What is the reason behind this? I mean, instead of insertions above you can just memcpy part by part into `message.data`.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +138,73 @@ namespace ROS2
...
+                pf.offset = sizeof(AZ::Vector3);
+            }
+
             pf.datatype = sensor_msgs::msg::PointField::FLOAT32;
             pf.count = 1;
             message.fields.push_back(pf);
         }
 
-        size_t sizeInBytes = lastScanResults.m_points.size() * sizeof(AZ::Vector3);
+        m_rosMessageData.clear();
```
PawelLiberadzki, line 181: I we will decide to keep `m_rosMessageData` despite my comment below - you do not have to clear, if you change reserve to resize and inserts below to pointers / indices. Technically it sounds like a thing saving some allocations.

Gems/ROS2/Code/Include/ROS2/Lidar/LidarRaycasterBus.h
```diff
@@ -88,6 +88,7 @@ namespace ROS2
     {
         Points = (1 << 0), //!< return 3D point coordinates
         Ranges = (1 << 1), //!< return array of distances
+        Intensity = (1 << 2), //!< return intensity data
```
PawelLiberadzki, line 91: Seems that plural is used here -> Point**s**, Range**s**. Same below, in line 101.

Gems/ROS2/Code/Source/Lidar/LidarCore.h
```diff
@@ -37,7 +37,7 @@ namespace ROS2
 
         //! Perform a raycast.
         //! @return Results of the raycast.
-        RaycastResult PerformRaycast();
+        const RaycastResult& PerformRaycast();
```
michalpelka, line 40: If possible, consider
```suggestion
        const RaycastResult& PerformRaycast() const ;
```

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +138,73 @@ namespace ROS2
...
+        PublishRaycastResults(lastScanResults);
+    }
+
+    void ROS2LidarSensorComponent::PublishRaycastResults(const RaycastResult& results)
+    {
+        const bool isIntensityEnabled = m_lidarCore.m_lidarConfiguration.m_lidarSystemFeatures & LidarSystemFeatures::Intensity;
 
         auto* ros2Frame = GetEntity()->FindComponent<ROS2FrameComponent>();
         auto message = sensor_msgs::msg::PointCloud2();
         message.header.frame_id = ros2Frame->GetFrameID().data();
```
michalpelka, line 157: Why we are not using this API to build messages:
https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/include/sensor_msgs/point_cloud2_iterator.hpp ?
It seems to be much cleaner than throwing data like a hot potato.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
michalpelka: I would consider to use https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/include/sensor_msgs/point_cloud2_iterator.hpp instead of this sizeofs.
If not please at least make a lot of AZ_Asserts to catch a bug.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +138,73 @@ namespace ROS2
...
+        AZ_Assert(
+            !isIntensityEnabled || (results.m_points.size() == results.m_intensity.size()),
+            "Inconsistency in the size of raycast result attributes.");
+
+        const auto entityTransform = GetEntity()->FindComponent<AzFramework::TransformComponent>();
+        const auto inverseLidarTM = entityTransform->GetWorldTM().GetInverse();
+        for (size_t pointIdx = 0; pointIdx < results.m_points.size(); ++pointIdx)
+        {
+            const AZ::Vector3 globalPoint = inverseLidarTM.TransformPoint(results.m_points[pointIdx]);
+            const auto positionBytes = reinterpret_cast<const uint8_t*>(&globalPoint);
```
michalpelka, line 195: Please use memcpy instead of reinterpret_cast, or reconsider point_cloud2_iterator.hpp.

michalpelka (commented):
As I mentioned multiple times in review, please consider using point_cloud2_iterator.hpp to build ROS 2 message.

Gems/ROS2/Code/Source/Lidar/LidarCore.cpp
```diff
@@ -161,8 +165,10 @@ namespace ROS2
         return m_lidarRaycasterId;
     }
 
-    RaycastResult LidarCore::PerformRaycast()
+    const RaycastResult& LidarCore::PerformRaycast()
```
spham-amzn, line 168: Any reason to change this to a reference other than to not need a copy constructor? Its a little risky to return a reference to a member variable because of its potential to change after this call is done by another function.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +138,73 @@ namespace ROS2
...
-        message.point_step = sizeof(AZ::Vector3);
+        message.width = results.m_points.size();
+        message.point_step = sizeof(AZ::Vector3) + (isIntensityEnabled ? sizeof(float) : 0U);
         message.row_step = message.width * message.point_step;
 
-        AZStd::array<const char*, 3> point_field_names = { "x", "y", "z" };
-        for (int i = 0; i < point_field_names.size(); i++)
+        AZStd::array point_field_names = { "x", "y", "z", "intensity" };
+        const size_t numPointFields = isIntensityEnabled ? point_field_names.size() : point_field_names.size() - 1U;
+        for (int i = 0; i < numPointFields; i++)
```
spham-amzn, line 166: There is signed/unsigned arithmetic on line 171, so suggest to make `i` unsigned as well.

```suggestion
        for (size_t i = 0; i < numPointFields; i++)
```

Gems/ROS2/Code/Source/Lidar/LidarCore.h
spham-amzn: > If possible, consider

Probably can't make this `const` because it is returning a reference to a member function now as part of this change. The original version where a copy instance of `RaycastResult` can be `const` however.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
alek-kam-robotec-ai: @michalpelka Nice suggestion, i'll look into it.

Gems/ROS2/Code/Source/Lidar/LidarCore.h
alek-kam-robotec-ai: This method cannot be made const due to modification of the `m_lastScanResults`.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
alek-kam-robotec-ai: I'm all for it.

Gems/ROS2/Code/Source/Lidar/LidarCore.cpp
alek-kam-robotec-ai: The reason for this is to ommit copying of the raycast results which are quite large. It is highly unlikely that this reference will get invalidated since it is only used after each raycast. Until this reference is processed and discarded no further raycasts (for this object at least) will be scheduled.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
alek-kam-robotec-ai: iterators solve this beautifully.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +139,87 @@ namespace ROS2
...
         message.height = 1;
-        message.width = lastScanResults.m_points.size();
-        message.point_step = sizeof(AZ::Vector3);
-        message.row_step = message.width * message.point_step;
+        message.width = results.m_points.size();
+
+        sensor_msgs::PointCloud2Modifier modifier(message);
+        if (isIntensityEnabled)
+        {
+            modifier.setPointCloud2Fields(
```
michalpelka, line 166: Love this approach, much cleaner!

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +139,87 @@ namespace ROS2
                 aznumeric_cast<AZ::u64>(timestamp.sec) * aznumeric_cast<AZ::u64>(1.0e9f) + timestamp.nanosec);
         }
 
-        auto lastScanResults = m_lidarCore.PerformRaycast();
+        const RaycastResult& lastScanResults = m_lidarCore.PerformRaycast();
```
PawelLiberadzki, line 142: LidarCore is the same, so I guess such thing can be made in Lidar2D component.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +139,87 @@ namespace ROS2
...
         if (m_canRaycasterPublish || !m_sensorConfiguration.m_publishingEnabled)
         { // Skip publishing when it is disabled or can be handled by the raycaster.
             return;
         }
 
-        const auto inverseLidarTM = entityTransform->GetWorldTM().GetInverse();
-        for (auto& point : lastScanResults.m_points)
-        {
-            point = inverseLidarTM.TransformPoint(point);
-        }
```
PawelLiberadzki, line 154: This does not seem to be replaced by anything or any transformation in other place - was it intended or missed?

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
```diff
@@ -140,43 +139,87 @@ namespace ROS2
...
-            message.fields.push_back(pf);
+            AZ::Vector3 position = results.m_points[i];
+            *xIt = position.GetX();
+            *yIt = position.GetY();
+            *zIt = position.GetZ();
+
+            if (isIntensityEnabled)
+            {
+                *intensityIt.value() = results.m_intensities[i];
+                ++intensityIt.value();
```
PawelLiberadzki, line 217: This is probably an error - you increment .value() instead of operator itself, like you do with e.g. xIt below.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
alek-kam-robotec-ai: Right, my bad. Thanks for noticing.

Gems/ROS2/Code/Source/Lidar/ROS2LidarSensorComponent.cpp
alek-kam-robotec-ai: Note that `intenistyIt` is a pointcloud iterator wrapped inside of the optional.

PawelLiberadzki (approved):
Looks fine! I ran this in a simple scene with terrain and it seems to work!

michalpelka (approved):

jhanca-robotecai (approved):
The code works as expected, only two nitpicks.

PR #756, opened 2024-09-18
Fix templates for ROS 2 Jazzy for the upcoming release

jhanca-robotecai:
## What does this PR do?

Update readme, prefabs and example ROS 2 navigation scripts in `ROS2ProjectTemplate` and `RoboticManipulationTemplate` to match to the next release and run correctly with Ubuntu 24.04 + ROS 2 Jazzy combination

I am aware this is a late call, but it is not changing any source of o3de.

## How was this PR tested?

Templates were build and run on Ubuntu 22.04 with Humble and Ubuntu 24.04 with Jazzy.

Tested with 
o3de `stabilization/2409` fd8cfb0b
o3de-extras `stabilization/2409` a7eec48e


jhanca-robotecai:
Added one more fix for the `Dockerfile` - tested on Ubuntu 2404 and Ubuntu 2204. I modified the code to use `stabilization/2409` for both o3de and o3de-extras for tests.

pawelbudziszewski (approved):

michalpelka (approved):
Tested all templates against debian-installed package

pijaro (approved):

PR #759, opened 2024-09-25
Fix overrides for frame component

arturkamieniecki:
This PR was tested with #758, but it should not require it.

## What does this PR do?
This PR changes how custom namespaces are handled. Before if the namespace wasn't custom it was serialized. This value was changed when importing a URDF creating a lot of misleading overrides. Now only the custom namespace is serialized.
I've also added initialization functions to make sure that the proper namespace is created before component activation (both for editor and game frame components).

## How was this PR tested?

This PR was tested by importing the turtlebot4 urdf and changing namespace values using overrides.


Gems/ROS2/Code/Include/ROS2/Frame/NamespaceConfiguration.h
```diff
@@ -58,8 +58,13 @@ namespace ROS2
...
         void SetParentNamespace(const AZStd::string& parentNamespace);
 
+        //! Initializes the namespace configuration.
+        //! This function should be called in the Init functions of frame components.
+        void Init();
+
     private:
         AZStd::string m_namespace;
+        AZStd::string m_customNamespace = "";
         AZStd::string m_parentNamespace;
```
jhanca-robotecai, line 68: 1. Please initialize all three values
2. Please add a comment to clearly differentiate between `m_namespace` and `m_customNamespace` as it might be misleading for people who work less often with ROS 2

jhanca-robotecai (approved):
This PR fixes the issue; only minor remarks.

I am adding a picture of HOW IT WAS for the reference - every element with a `ROS2 Frame` component was marked as override:
image
image

Gems/ROS2/Code/Include/ROS2/Frame/NamespaceConfiguration.h
jhanca-robotecai: Pushed the fix.

pawelbudziszewski (approved):

PR #760, opened 2024-09-26
Add check for zero quaterion

arturkamieniecki:
## What does this PR do?

This PR adds a check for a zero quaternion in the spawner.

## How was this PR tested?

By spawning entities with zero quaternions and non-zero quaternions.

Resolves #757.

PR #762, opened 2024-09-26
Add check for zero quaterion

jhanca-robotecai:
## What does this PR do?

Cherry-pick of #760 to stabilization.

It fixes the issue within ROS 2 Spawner: O3DE crashes when spawning with undefined rotation. It used to be the user's responsibility to give correct data, but the check is straightforward and it could save from the crash.

## How was this PR tested?

Spawning with incorrect rotation was performed.


michalpelka (approved):

pawelbudziszewski (approved):

adamdbrw:
Warn plus default for better UX

adamdbrw (approved):

PR #766, opened 2024-10-08
repo.json updates for release 2409

spham-amzn:
## What does this PR do?
Updates the repo.json for o3de-extras for the 2409 release

## How was this PR tested?
Locally tested

### DIFF
--- a/Gems/ROS2/gem.json
+++ b/Gems/ROS2/gem.json
@@ -36,6 +36,5 @@
     ],
     "restricted": "ROS2",
     "repo_uri": "https://raw.githubusercontent.com/o3de/o3de-extras/development",
-    "download_source_uri": "https://github.com/o3de/o3de-extras/releases/download/2.0/ros2-3.2.1-gem.zip"
+    "download_source_uri": "https://github.com/o3de/o3de-extras/releases/download/2.0//ros2-3.2.1-gem.zip"
 }
-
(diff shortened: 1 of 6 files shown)


nick-l-o3de (approved):

moudgils (approved):

amzn-changml (approved):

Gems/ROS2/gem.json
```diff
@@ -1,6 +1,6 @@
 {
     "gem_name": "ROS2",
-    "version": "3.1.0",
+    "version": "3.2.1",
```
jhanca-robotecai, line 3: Why was this version changed? It is now ahead of the `development` branch. It also means that we bumped the version number from `2.0.0` up to `3.2.1` before the two stable releases, which I do not understand.

spham-amzn: The current release package for the ROS2 gem was version 3.2.0. For some reason, this version update wasn't cherry picked back to development (which was at 3.1.0), and we need the minor version bump to update the released gem zip file. 
As for 2.0.0 from the previous stable version, this was aa major version bump to support the PhysX split due to the gem dependency changes for related gems and templates (It needed to change PhysX -> PhysX5) so all of their versions needed a bump as well. (see https://github.com/o3de/ROSConDemo/pull/276 )

PR #772, opened 2024-10-11
Extend functionality of FrameConversion.py to handle overrides

zakmat:
## What does this PR do?

FrameConversion.py script is a nice script that automatically converts existing prefabs to work with new version of ROS2FrameEditorComponent supporting the migration of existing projects into the latest 2409 release.

There is however one detail that was missing - namely overrides were not handled. Therefore if the prefab overriden the ROS2FrameComponent down in the hierarchy the conversion was not applied there. The patch was no longer valid with the new serialization schema and was ignored as a result. This could be seen as warnings in Prefab's AssetProcessor logs. 

A typical scenario is a multi sensor robot that overrides frame names for multiple instances of the sensor prefab.

This PR tries to tackle the issue. Since the ROS2FrameComponent is not present in the patch's path I opted to instead look for patterns with ROS2FrameComponent field names:
```
                    "m_template/Namespace Configuration",
                    "m_template/Frame Name",
                    "m_template/Publish Transform",
                    "m_template/Joint Name"
```
**Warning: This can potentially break 3rd party components.** As far as I know those names are unique across the Canonical Gems but some other 3rd-party Components might be affected. Please consider it during review.

Context: https://github.com/o3de/o3de-extras/pull/593

## How was this PR tested?

After the migration I noticed the issue and fixed manually on separate branch. Once confirmed that frames were set correctly I have applied the automatic conversion and diffed the results

### DIFF
--- a/Gems/ROS2/Code/Source/Frame/Conversions/FrameConversion.py
+++ b/Gems/ROS2/Code/Source/Frame/Conversions/FrameConversion.py
@@ -12,7 +12,7 @@ import decimal
 
 def find_and_replace(data):
 
-    def search_for_components(item, foundComponents):
+    def search_for_components(item, foundComponents, insidePatches=False):
         if isinstance(item, dict):
             if "Components" in item:
                 foundComponents = True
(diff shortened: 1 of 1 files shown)


arturkamieniecki (approved):
Looks good.

jhanca-robotecai:
Rebased and force-pushed the code on top of the `development` branch - each PR in the repo should target the `development` branch and should be cherry-picked if necessary.

jhanca-robotecai (approved):
Works as expected; I assume third party components should be ignored in such case.

PR #777, opened 2024-10-28
Allow ROS2 Frame Component to be present on Level Entity

MateuszWasilewski:
## What does this PR do?

This PR allows adding ros2 frame to Level Entity, it allows for simplified level design when adding robots during simulation run. 

## How was this PR tested?

On 2409

1. Created new Level
2. Added ROS2 Frame on level entity
3. added robot to the level
4. Set robot top level tf to `publish tf on`
5. Verified tf tree in rqt-tf-tree


adamdbrw:
I would like to review this change (should find time this week)

MateuszWasilewski:
> What are the consequences when such component is added at the top?

Assuming that we add tf frame `tf0` at the level entity and 2 robots
- namespace `name1` top tf `tf1`
- namespace `name2` top tf `tf2` 

This will result in tf tree that looks similarly to
```
     -> name1/tf1 -> ...
   /
tf1
   \
     -> name2/tf2 -> ...
```
which conceptually corresponds to Entity tree in o3de
```
LevelEntity
├── Robot1
└── Robot2
```

adamdbrw (approved):

michalpelka (approved):

jhanca-robotecai:
Rebased on top of the `development` branch so we can successfully run tests

PR #789, opened 2024-11-08
Added if statement for 'remove' operations in FrameConversion script

patrykantosz:
## What does this PR do?

The FrameConversion.py script is very helpful if you need to migrate your project to the 2409 release. However one thing is still missing, even with PR #772 - modifying `remove` operations in the `Patches` section is not handled properly. 
```
                {
                    "op": "remove",
                    "path": "/Entities/Entity_[8678293627970]/Components/ROS2FrameComponent"
                },
 ```

This PR extends the script capabilities - I think that now all cases are now covered.

## How was this PR tested?

By preparing nested prefab with the O3DE 2310, converted prefab with a script and used it with the O3DE 2409.


jhanca-robotecai (approved):
Created some prefabs in 2309, imported to 2409; remove override works correctly.

PR #799, opened 2024-11-22
Fixes following o3de #18405

jhanca-robotecai:
## What does this PR do?

PR https://github.com/o3de/o3de/pull/18405 changed target names in O3DE Gems, which caused build failures in o3de-extras. This is fixed in this PR.

Additionally:
- _ROS 2 Gem_ version was changed to newer than the last release (new features were added, non-breaking change, hence minor)
- _ROS 2 Gem_ was set to require _development_ version of o3de (it will not build otherwise)
- _Ros2ProjectTemplate_ version was changed to reflect changes in the cmake

## How was this PR tested?

A sample project was built with o3de from `development` branch, 0efe207e34

### DIFF
--- a/Gems/ROS2/gem.json
+++ b/Gems/ROS2/gem.json
@@ -1,6 +1,6 @@
 {
     "gem_name": "ROS2",
-    "version": "3.2.1",
+    "version": "3.3.0",
     "platforms": [
         "Linux"
     ],
@@ -19,11 +19,11 @@
         "ROS2"
     ],
     "compatible_engines": [
-        "o3de-sdk>=2.3.0",
-        "o3de>=2.3.0"
+        "o3de-sdk>=4.2.0",
+        "o3de>=4.2.0"
     ],
     "icon_path": "preview.png",
-    "requirements": "Requires ROS 2 installation (supported distributions: Humble). Source your workspace before building the Gem",
+    "requirements": "Requires ROS 2 installation (supported distributions: Humble and Jazzy). Source your workspace before building the Gem",
     "documentation_url": "https://docs.o3de.org/docs/user-guide/gems/reference/robotics/ros2/",
     "dependencies": [
         "Atom_RPI",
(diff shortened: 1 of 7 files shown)


Templates/Ros2ProjectTemplate/Template/Gem/CMakeLists.txt
```diff
@@ -125,7 +125,7 @@ if(PAL_TRAIT_BUILD_HOST_TOOLS)
         BUILD_DEPENDENCIES
             PUBLIC
                 ${gem_name}.Editor.Static
-                Gem::Atom_Feature_Common.Static
+                Gem::Atom_Feature_Common.Public
```
spham-amzn, line 128: Maybe we can just remove this dependency, I don't think its needed

spham-amzn, line 128: Maybe we remove this dependency, I don't think its needed explicitly here

spham-amzn (approved):

Gems/ROS2/gem.json
```diff
@@ -19,11 +19,11 @@
         "ROS2"
     ],
     "compatible_engines": [
-        "o3de-sdk>=2.3.0",
-        "o3de>=2.3.0"
+        "o3de-sdk>=4.2.0",
+        "o3de>=4.2.0"
```
ruffsl, line 23: Was this change correct? The latest stable version is only `o3de-sdk 2.3.2` with this release from last month:
- https://github.com/o3de/o3de/releases/tag/2409.2
- https://github.com/o3de/o3de/blob/2409.2/engine.json#L7
- https://github.com/o3de/o3de/blob/995e3aff8d608faaf749e261a7ea796df5417f24/engine.json#L7

This otherwise results in this template error:

```shell
$ ${O3DE_HOME}/scripts/o3de.sh create-project --project-path $PROJECT_PATH --template-path ${O3DE_EXTRAS_HOME}/Templates/Ros2RoboticManipulationTemplate
1.11.1
Project created at /home/ruffsl/ws/airsim/projects/AirSim
[ERROR] o3de.register: AirSim is not known to be compatible with the following objects/APIs and requires the --force parameter to register:
  The following dependency requirements could not be satisfied:


ROS2 is incompatible because: o3de-sdk 2.3.2 does not match any version specifiers in the list of compatible engines: ['o3de-sdk>=4.2.0', 'o3de>=4.2.0']
```

@spham-amzn @jhanca-robotecai

ruffsl: Never mind, following git blame on the upstream o3de repo on GitHub switched me to a stale GitHub fork.
- https://github.com/o3de/o3de/blob/development/engine.json#L7C17-L7C22
- https://github.com/o3de/o3de/blob/8719dc3e6f40512b954f4baa99b0b0298b7331fc/engine.json#L7C17-L7C22

PR #800, opened 2024-11-29
Add notification bus to spawner

arturkamieniecki:
## What does this PR do?

This PR adds a bus to the spawner component that notifies other components about spawns.

## How was this PR tested?

By spawning and despawning entities and listening on the bus with a custom component.

### DIFF
--- a/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -8,6 +8,7 @@
 
 #include "ROS2SpawnerComponent.h"
 #include "Spawner/ROS2SpawnerComponentController.h"
+#include <AzCore/Component/EntityId.h>
 #include <AzCore/Math/Quaternion.h>
 #include <AzCore/Serialization/EditContext.h>
 #include <AzCore/Serialization/SerializeContext.h>
@@ -18,6 +19,7 @@
 #include <ROS2/Georeference/GeoreferenceBus.h>
 #include <ROS2/ROS2Bus.h>
 #include <ROS2/ROS2GemUtilities.h>
+#include <ROS2/Spawner/SpawnerBus.h>
 #include <ROS2/Utilities/ROS2Conversions.h>
 #include <ROS2/Utilities/ROS2Names.h>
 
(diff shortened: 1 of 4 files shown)


jhanca-robotecai:
Is this PR ready for the review?

arturkamieniecki:
> Is this PR ready for the review?

Yes. This code was created to allow notifications about spawning to be sent to other components. Currently, no ROS2Gem components use this feature.

arturkamieniecki:
Drafted as new functionality is added.

arturkamieniecki:
I've added the notification handlers to the behavior context.

arturkamieniecki:
I've added the notification handlers to the behavior context.

jhanca-robotecai (approved):
Tested with ScriptCanvas and the notifications work correctly (both spawning and despawning).

A design question: what is the reasoning behind the passing three data fields (`spawnableName`, `rootEntityId`, `ticketName`) when spawning and only one (`ticketName`) when despawning? It looks somewhat inconsistent.

arturkamieniecki:
> A design question: what is the reasoning behind the passing three data fields (spawnableName, rootEntityId, ticketName) when spawning and only one (ticketName) when despawning? It looks somewhat inconsistent.

I wanted to send as much identifying information as possible. The spawnable name identifies the object type, entity id its place in O3DE, and the ticket name is the unique key that identifies it. There is no need to resend the spawnable name and entity id for the despawn notification as the ticket name uniquely identifies it.

Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -98,6 +101,11 @@ namespace ROS2
         {
             serialize->Class<ROS2SpawnerComponent, ROS2SpawnerComponentBase>()->Version(1);
         }
+
+        if (AZ::BehaviorContext* behaviorContext = azrtti_cast<AZ::BehaviorContext*>(context))
+        {
+            behaviorContext->EBus<SpawnerNotificationBus>("SpawnerNotificationBus")->Handler<SpawnerNotificationsBusHandler>();
```
jhanca-robotecai, line 107: One more comment - since we are calling the component `ROS2SpawnerComponent`, I would suggest keeping `ROS2` within the name, so we can easily differentiate it from the built in spawner notifications.
```suggestion
            behaviorContext->EBus<SpawnerNotificationBus>("ROS2SpawnerNotificationBus")->Handler<SpawnerNotificationsBusHandler>();
```

adamdbrw (approved):

jhanca-robotecai:
Rebased and force-pushed on top of the `development` due to failing tests.

Issue #792, opened 2024-11-13
Terrain is not rendered in ROS 2 Camera sensor when using headless mode


pijaro:
**Describe the bug**
A terrain is not being rendered on ROS 2 Camera when running GameLauncher in headless mode (`-console-mode`). Running in normal mode works fine.

**Steps to reproduce**
1. Create terrain on some test level
2. Place ROS 2 Sensor camera in a level
3. Run the GameLauncher with `-console-mode` argument.
4. Check the output in RVIZ2

**Expected behavior**
Terrain is rendered correctly

**Actual behavior**
Terrain is black.

**Screenshots/Video**
| Editor | GameLauncher with console mode |
| --- | --- |
| ![image](https://github.com/user-attachments/assets/d711c1d9-dfc8-465d-a942-1e4ea38e9538) | ![image](https://github.com/user-attachments/assets/a38a985f-a8b9-4fe1-8ccd-f5df1bba2708) |

**Found in Branch**
point-release/24091

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) and [o3de/o3de-extras](https://github.com/o3de/o3de-extras) repositories**
extras: dca4060c3fe1c9fc183498d3534f4de15a09c2b7
o3de: [e8570f9d635c0abbf3d733d0524a3c0ae75d726c](https://github.com/o3de/o3de/tree/e8570f9d635c0abbf3d733d0524a3c0ae75d726c)

**Desktop/Device (please complete the following information):**
 - Device: PC
 - OS: Ubuntu 22.04
 - CPU Intel i7
 - GPU RTX 4080
 - Memory 64GB

michalpelka:
We found that switching one flag cause terrain to render:
```diff
--- a/Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Lighting/StandardLighting.azsli
+++ b/Gems/Atom/Feature/Common/Assets/ShaderLib/Atom/Features/PBR/Lighting/StandardLighting.azsli
@@ -79,7 +79,7 @@ PbrLightingOutput GetPbrLightingOutput(Surface surface, LightingData lightingDat
     lightingOutput.m_albedo.rgb = surface.albedo * lightingData.diffuseResponse * lightingData.diffuseAmbientOcclusion;
     lightingOutput.m_albedo.a = lightingData.specularOcclusion;
     lightingOutput.m_normal.rgb = EncodeNormalSignedOctahedron(surface.GetDefaultNormal());
[...]

PR #801, opened 2024-12-02
Fix: terrain visibility in console mode

pijaro:
## What does this PR do?

This PR fixes terrain rendering in ROS 2 cameras when `-console-mode` is on.

## How was this PR tested?

In custom project with ROS 2 gem anbled. I've put two ROS 2 cameras and terrain in the level. Then build the project and run GameLauncher with `-console-mode` on. Also the Editor output has been tested.

Level: 
[DefaultLevel.zip](https://github.com/user-attachments/files/17979375/DefaultLevel.zip)

| Before | After |
| --- | --- |
| before | after |

Fixes https://github.com/o3de/o3de-extras/issues/792

### DIFF
--- a/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
@@ -17,8 +17,10 @@
 #include <Atom/RPI.Public/RPISystemInterface.h>
 #include <Atom/RPI.Public/RenderPipeline.h>
 #include <Atom/RPI.Public/Scene.h>
+#include <Atom/RPI.Reflect/Asset/AssetUtils.h>
 #include <AzCore/Math/MatrixUtils.h>
 #include <AzCore/Settings/SettingsRegistry.h>
+#include <AzCore/Component/ComponentApplicationBus.h>
 #include <AzFramework/Components/TransformComponent.h>
 #include <AzFramework/Scene/SceneSystemInterface.h>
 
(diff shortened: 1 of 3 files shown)


michalpelka:
There is a possibility that it was fixed in development of o3de : https://github.com/o3de/o3de/pull/18510.
I developed a change in o3de : https://github.com/RobotecAI/o3de/tree/mp/fix_console_mode_brdf on top of 2409 that fixes a logic in BoostrapComponent, but during merge I found that it was redesigned. 
I would not merge it here. I would propose https://github.com/RobotecAI/o3de/tree/mp/fix_console_mode_brdf or 18510 as fix to be at next o3de point release.

zakmat (commented):
Impressive fix!

Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
```diff
@@ -128,6 +130,39 @@ namespace ROS2
         m_view->SetViewToClipMatrix(m_cameraSensorDescription.m_viewToClipMatrix);
         m_scene = AZ::RPI::RPISystemInterface::Get()->GetSceneByName(AZ::Name("Main"));
 
+        // In console mode, since there is no default pipeline, a BRDF texture needs to be manually initialized.
+        AZ::ApplicationTypeQuery appType;
+        AZ::ComponentApplicationBus::Broadcast(&AZ::ComponentApplicationBus::Events::QueryApplicationType, appType);
```
moudgils, line 135: Could the brdf pipeline simply not be added within BootstrapSystemComponent::EnsureDefaultRenderPipelineInstalledForScene instead of adding it here?

pijaro: I would say "that depends". 

I think that, as a rule, console-mode shouldn't come with a default pipeline, because the rendering is assumed to not exist by definition. Shouldn't the components that require rendering be responsible for setting it up completely, including BRDF if needed?

On the other side, it could be useful and a good quality of life improvement if the bootstrap component would provide a minimal required pipeline (i.e. BRDF). Alternatively, I was thinking about a bootstrap bus that would handle BRDF creation, and then use it here instead of all this code.

What do you think? (cc @michalpelka)

michalpelka: @moudgils @pijaro I've prepared fix in development O3DE:
https://github.com/o3de/o3de/pull/18531

michalpelka:
Closing, since this fixed was introduced to o3de repo : https://github.com/o3de/o3de/pull/18531

PR #806, opened 2024-12-06
ROS2 Relative spawning

pijaro:
## What does this PR do?

Enables spawning entities in an arbitrary coordinate system.

Similar to spawning in WGS coordinate system, enabling the option and setting "relative" in ROS 2 service "reference_frame" spawns entity relatively to the selected coordinate system.

image
image

## How was this PR tested?

1. Spawning with feature off
2. Spawning with feature on/off and with/without reference frame set
3. Checked different positions of reference frame

### DIFF
--- a/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
+++ b/Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
@@ -114,6 +114,7 @@ namespace ROS2
     {
         AZStd::string referenceFrame(request->reference_frame.c_str());
         const bool isWGS{ referenceFrame == "wgs84" && m_controller.GetSupportWGS() };
+        const bool relativeSpawning{ referenceFrame == "relative" && m_controller.IsRelativeSpawningEnabled() };
 
         SpawnEntityResponse response;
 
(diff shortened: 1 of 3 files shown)


pijaro:
`reference_frame` is used when we want to spawn an object using (x, y, z) coordinates. The default coordinate system is at O3DE root - (0,0,0) and no rotation. The `reference_frame` changes it to any other system - useful. when we want to spawn an object relatively to another object. It is common to use georeference component that is practically our new "origin" of the level, and we want to use it instead of global O3DE (0,0,0).

Named spawn points (`ROS2SpawnPointComponent`) positions are implicitly defined (a location in a level, we don't truly care about the coordinates) and are always in O3DE root (0, 0, 0). Relative spawning has no use in that case.

To sum up, these two frames are solving different issues and serves different purposes.

jhanca-robotecai:
Why not adding an option to spawn an object using xyz coordinates with a _named spawn point_ used as a _reference frame_? I still don't get why do we need another tooling for that. The change looks like a quickfix for a particular problem within a particular project.

`gazebo_msg` package is deprecated and we will be removing it in the next full O3DE release, hence most of the spawner's code will be reimplemented. We could start the work already by preparing an RFC and start listing all features that should be included in the spawner.

pijaro:
> Why not adding an option to spawn an object using xyz coordinates with a _named spawn point_ used as a _reference frame_?

That is an alternative way of doing it. 

However, I would argue if it is more semantically correct. In my opinion, it would be a better idea to guide the spawner where its reference frame really is (as this is a level-only information, like the position of geo frame in a level), rather than passing that in the message. Your suggestion looks like adding an "offset" to spawning point, rather than changing the reference frame.

This PR adds a feature that is very useful in any project that uses WGS geo reference origin.

michalpelka:
This modifies obsolete code.

PR #808, opened 2024-12-12
Move Joint Tracjectory State to TrajectoryComponent

MateuszWasilewski:
## What does this PR do?

This PR fixes state management when calling `JointsTrajectoryRequestBus`
Without this change, method 
```
    class JointsTrajectoryRequests : public AZ::EBusTraits
...
        //! Retrieve current trajectory goal status.
        //! @return Status of trajectory goal.
        virtual TrajectoryActionStatus GetGoalStatus() = 0;
```
Always returns `TrajectoryActionStatus::Idle` because `FollowJointTrajectoryActionServer` never receives any goals. 

This change effectively allows for using `JointsTrajectoryRequestBus` because without it. No movement will occur (cause of broken status in `JointsTrajectoryComponent::FollowTrajectory`)

## How was this PR tested?

`JointsTrajectoryRequestBus` was called in two ways.
1. Directly from cpp code by `JointsTrajectoryRequestBus::EventResult`
2. From ros2 via ros2 action server and `FollowJointTrajectoryActionServer`

Without the change only the second one worked. After the change both calls work as expected.

### DIFF
--- a/Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
+++ b/Gems/ROS2/Code/Source/Manipulation/FollowJointTrajectoryActionServer.cpp
@@ -23,16 +23,6 @@ namespace ROS2
             AZStd::bind(&FollowJointTrajectoryActionServer::GoalAcceptedCallback, this, AZStd::placeholders::_1));
     }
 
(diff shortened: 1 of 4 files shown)


michalpelka (approved):

jhanca-robotecai (approved):
Works as expected (tested in the internal project), great job!

PR #811, opened 2024-12-26
Remove Unused AWS Gem from OpenXRTest

AMZN-Gene:
## What does this PR do?
Remove unused AWS core gem from OpenXRTest
Need for Jenkins to build https://github.com/o3de/o3de/pull/18571/

## How was this PR tested?
Built OpenXRTest using a o3de branch without AWS gems
Check AssetProcessor for any failed lua/scriptscanvas assets which may have been using AWS 
Unfortunately, there were a lot of shader/material asset errors and a fatal "RPI System could not initialize" error on Editor startup, but this occurred in vanilla development branch as well as the AWS removal branch

### DIFF
--- a/Projects/OpenXRTest/Gem/enabled_gems.cmake
+++ b/Projects/OpenXRTest/Gem/enabled_gems.cmake
@@ -3,7 +3,6 @@ set(ENABLED_GEMS
     OpenXRTest
     Atom
     AudioSystem
-    AWSCore
     CameraFramework
     DebugDraw
     EditorPythonBindings
--- a/Projects/OpenXRTest/project.json
+++ b/Projects/OpenXRTest/project.json
@@ -1,6 +1,6 @@
 {
     "project_name": "OpenXRTest",
-    "version": "1.0.1",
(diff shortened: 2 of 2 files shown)


forkercat (approved):

jhanca-robotecai (commented):
I never understood how things are versioned in O3DE, but shouldn't the project be bumped from `1.0.1` to `1.1.0`?

AMZN-Gene:
> I never understood how things are versioned in O3DE, but shouldn't the project be bumped from `1.0.1` to `1.1.0`?

Great point... I'm not quite sure either, but my guess is a version bump _would_ be required when there's a possibility of incompatibility. For example, if the project was updated to work with an API change in the engine. I think in this case, since AWS Core was never actually used, this change can't ever break someone depending on which engine or gems they choose

AMZN-Gene:
Bumping version patch number. Patch number is for changes that don't include any API updates.
https://docs.o3de.org/docs/contributing/release-versioning-and-terms/

Issue #816, opened 2025-01-10
ROS2OdometrySensor crashes O3DE when linked to articulation link


jhanca-robotecai:
**Describe the bug**
`ROS2OdometrySensorComponent` is realized based on `PhysX::RigidBodyComponent`. When added to the entity, it asks for a service `PhysicsDynamicRigidBodyService`, but this service is also realized by the `PhysX::ArticulationLink`. If the latter is chosen, O3DE will crash when game mode starts.

**Assets required**
ROS2 Gem

**Steps to reproduce**
1. Add entity to empty level
2. Add `ROS2OdometrySensorComponent` to the level, add required `ROS2 Frame Component`
3. Add `Articulation Link Component`
4. Start game mode
5. Observe crash

**Expected behavior**
No crash

**Actual behavior**
Crash

**Screenshots/Video**
![Screenshot from 2025-01-10 16-31-05](https://github.com/user-attachments/assets/6c7eac38-10a4-4993-aacc-51bfb7d7faaf)

**Found in Branch**
`development`
`main`

**Commit ID from [o3de/o3de](https://github.com/o3de/o3de) and [o3de/o3de-extras](https://github.com/o3de/o3de-extras) repositories**
All commits

**Desktop/Device (please complete the following information):**
PC with Ubuntu Linux

jhanca-robotecai:
The crash was removed some time ago, due to adding `ArticulationLinkService` into _incompatible_ services (#817). Hence, the sensor works correctly with `PhysX::RigidBodyComponent` and is disabled in the other case. The issue remains open, as the implementation for `ArticulationLinkService` should be added.

PR #817, opened 2025-01-10
Disable ROS2OdometrySensor for Articulation links

jhanca-robotecai:
## What does this PR do?

This PR prevents user from getting a crash when adding `ROS2OdometrySensor` and `Articulation Link` into the same entity, see #816 

This is a simple fix that adds `Articulation Link` to the incompatible service. `ROS2OdometrySensor` using the link should be implemented as a proper solution to #816

## How was this PR tested?

Tried to add the `Articulation link`. Before (left) and after (right):
image


michalpelka (approved):

michalpelka:
It would be great to add support for IMU on articulation at some point in time

spham-amzn (approved):

PR #818, opened 2025-01-17
Fix PostProcessFeatureProcessor view aliasing in ROS2 camera

arturkamieniecki:
## What does this PR do?

Currenty the ros2 cameras view was aliased to the main game camera. This enabled the PostProcessFeatureProcessor to works with the ROS2 camera. This aliasing was done during camera activation, but if the main camera was not activated the ROS2 camera aliased to a wrong view (probably the old editor view). To fix this issue I've added the Camera Notification Bus handler to the ROS2 camera. It sets the alias when the main camera view changes.

## How was this PR tested?

By adding a ROS2 camera and fog component and checking if the fog is visible in the topic. I've also verified different this behaviour with different activation orders (1st o3de camera, 2nd ros2 camera and 1st ros2 camera, 2nd o3de camera)

### DIFF
--- a/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
+++ b/Gems/ROS2/Code/Source/Camera/CameraSensor.cpp
@@ -158,15 +158,33 @@ namespace ROS2
         m_passHierarchy.push_back("CopyToSwapChain");
 
         m_pipeline->SetDefaultView(m_view);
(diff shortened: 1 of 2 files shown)


patrykantosz (approved):
Works as expected, great job!

jhanca-robotecai (approved):

PR #819, opened 2025-01-22
Added registry setting to SteadyClock that allows to disable resettin…

michalpelka:
…g the clock when level is reloaded.

This is useful when you want to keep the clock running between level reloads, for example to keep some ROS 2 nodes happy that are running in the background, and you are reload / reset the level.

## What does this PR do?

It is new feature.
It is optional, it does not change default behavior.

## How was this PR tested?

Tested in external project against 2409.1
with registry adjusted:
```
{
    "O3DE":
    {
        "InputSystem":
        {
            "Mouse":
            {
                "CaptureMouseCursor": false
            }
        },
        "ROS2":
        {
            "SteadyClock" : true,
            "SteadyClock":
            {
                "ResetTimestampOnLevelReload": false
            },
            "Camera":
            {
                "AllowPipelineModification": false
            }
        }
    }
}
```

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Clock/SimulationTimeSource.h
+++ b/Gems/ROS2/Code/Include/ROS2/Clock/SimulationTimeSource.h
@@ -36,6 +36,7 @@ namespace ROS2
 
     private:
         double m_elapsed = 0;
+        bool m_resetTimeOnRestart = true;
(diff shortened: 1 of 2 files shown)


jhanca-robotecai (approved):
The code works as expected (changing via registry works); the default setting does not change the expected behavior. The feature makes sense - approve.

I rebased the code, clang-formatted, and force-pushed to ensure AR would have less issues.

PR #831, opened 2025-02-20
Bump versions of robotic gems and templates

jhanca-robotecai:
## What does this PR do?

1. Fix ROSBot prefab: missing parent entity
2. Bump versions of robotic gems and templates
3. Update repo.json for the release

The versions are changed as follows:
- `ROS 2 Gem` version 3.2.2 -> 3.2.3
- `RosRobotSample Gem` version 2.0.1 -> 2.0.2
- `Ros2FleetRobotTemplate` version 2.0.2 -> 2.0.3
- `Ros2ProjectTemplate` version 2.0.2 -> 2.0.3
- `Ros2RoboticManipulationTemplate` version 2.0.1 -> 2.0.2

Note, that this is a final PR for this point release from the *sig-simulation* point of view. Release files were uploaded as a part of this PR.

## How was this PR tested?

Three templates and two gems were pulled from the release and verified (gem.json; template.json).

PR #833, opened 2025-02-21
Revert ROSBotSample Gem 2.0.2 release

jhanca-robotecai:
## What does this PR do?

This PR reverts the change introduced in `ROSBotSample` Gem introduced in #831 which fixed the `cover_link` entity. Besides fixing the `cover_link` entity, the change introduces other problems. After some internal `sig-simulation` conversations, the decision was to revert it. It is better to have an incorrect fixed joint than a stationary robot.

## How was this PR tested?

`git diff ` was used to check if the change in the Gem was reverted and if the change in `repo.json` was correctly reverted.

### DIFF
--- a/Gems/RosRobotSample/Assets/ROSbot.prefab
+++ b/Gems/RosRobotSample/Assets/ROSbot.prefab
@@ -1047,7 +1047,7 @@
                     "$type": "EditorFixedJointComponent",
                     "Id": 9934662879286617893,
                     "Configuration": {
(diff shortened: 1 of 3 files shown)


pijaro (approved):
Tested the template both in Editor and monolithic gamemodes in release and profile.

Works as expected.

michalpelka (approved):
Approving, but I suffered an error that was pointed in PR for point release.

spham-amzn (approved):

PR #834, opened 2025-02-23
Add superluminal, Tracy and Optick profilers integration as Gems

guillaume-haerinck:
## What does this PR do?

Users have now the choice to use : 
- [imgui built-in](https://github.com/o3de/o3de/wiki/CPU-Profiling-Tools) profiler
- [PIX profiler](https://github.com/o3de/o3de/wiki/GPU-Profiling-Tools)
- [Superluminal](https://superluminal.eu/)
- [Tracy](https://github.com/wolfpld/tracy)
- [Optick](https://github.com/bombomby/optick)

Note : Pix integration would need to be made cleaner, as is does not rely on the `AZ::Debug::Profiler` (and said interface needs improvements to support GPU events)
Note : The profile API is lacking and will need improvements to supports these profilers to their full potential

### Superluminal

change

### Tracy

image

### Optick

optick

## How was this PR tested?

On windows custom build on development branch


Gems/ExternalProfilers/SuperluminalProfiler/Code/Source/CpuProfiler.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <AzCore/Interface/Interface.h>
+#include <AzCore/Serialization/SerializeContext.h>
+#include <Superluminal/PerformanceAPI.h>
+
+namespace SuperluminalProfiler
+{
+	void CpuProfiler::Init()
+	{
+		AZ::Interface<AZ::Debug::Profiler>::Register(this);
```
guillaume-haerinck, line 19: Given that only one profiler can be active at a time, I might want to assert there, but ideally I should use a mechanism to say that when this gem is enabled none of the other profiling gem can be enabled, I don't know if we have something like that

spham-amzn:
Looks like tabbing vs spaces issues going on

Gems/ExternalProfilers/SuperluminalProfiler/Code/Source/CpuProfiler.cpp
spham-amzn: There is, kinda-of. There is a field in gems.json called `provided_unique_service` that you can set to any string, and the scripts to enable gems should check against this. We did this for PhysX4 vs PhysX5 

https://github.com/o3de/o3de/blob/f29613a5c80d8dbc53016bdaaf3849cd6d77f60f/Gems/PhysX/Core/PhysX4/gem.json#L26

https://github.com/o3de/o3de/blob/f29613a5c80d8dbc53016bdaaf3849cd6d77f60f/Gems/PhysX/Core/PhysX5/gem.json#L26

guillaume-haerinck:
> Looks like tabbing vs spaces issues going on

Ran the clang-format on all of these files :)

Gems/ExternalProfilers/OptickProfiler/Code/Source/CpuProfiler.h
```diff
@@ -0,0 +1,49 @@
...
+#include <AzCore/Debug/Profiler.h>
+#include <AzCore/Memory/SystemAllocator.h>
+#include <AzCore/Name/Name.h>
+#include <AzCore/RTTI/RTTI.h>
+#include <AzCore/std/parallel/mutex.h>
+#include <AzCore/std/parallel/shared_mutex.h>
+
+namespace OptickProfiler
+{
+    class CpuProfiler final
```
moudgils, line 21: Minor. Add a comment describing the purpose of the class

Gems/ExternalProfilers/OptickProfiler/Code/Source/ProfilerSystemComponent.h
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include "CpuProfiler.h"
+
+#include <AzCore/Component/Component.h>
+#include <AzCore/Debug/ProfilerBus.h>
+
+namespace OptickProfiler
+{
+    class ProfilerSystemComponent
```
moudgils, line 18: Add a comment explaining the purpose of the class

Gems/ExternalProfilers/SuperluminalProfiler/Code/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+#
+# Copyright (c) Contributors to the Open 3D Engine Project.
+# For complete copyright and license terms please see the LICENSE at the root of this distribution.
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+set(SUPERLUMINAL_API_PATH "C:/Program Files/Superluminal/Performance/API" CACHE PATH "Path to the Superluminal 3rd party profiler API folder.")
```
moudgils, line 9: Having hardcoded path here does not seem right. What do other gems do?

Gems/ExternalProfilers/SuperluminalProfiler/Code/Source/CpuProfiler.h
```diff
@@ -0,0 +1,48 @@
...
+#include <AzCore/Debug/Profiler.h>
+#include <AzCore/Memory/SystemAllocator.h>
+#include <AzCore/Name/Name.h>
+#include <AzCore/RTTI/RTTI.h>
+#include <AzCore/std/parallel/mutex.h>
+#include <AzCore/std/parallel/shared_mutex.h>
+
+namespace SuperluminalProfiler
+{
+    class CpuProfiler final : public AZ::Debug::Profiler
```
moudgils, line 21: Needs a class comment

moudgils (commented):
Looks good for the most part. Some minor comments. We should add also add info on how to use the profilers. If not on the main documentation web site then atleast on the Github wiki.

Gems/ExternalProfilers/SuperluminalProfiler/Code/CMakeLists.txt
guillaume-haerinck: Hey ! Its just the default value and is exposed in the cmake during configure

Gems/ExternalProfilers/OptickProfiler/Code/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0 OR MIT
+#
+#
+
+# Add Optick dependency using FetchContent
+include(FetchContent)
+FetchContent_Declare(
+    Optick
+    GIT_REPOSITORY "https://github.com/bombomby/optick"
```
nick-l-o3de, line 13: this works?  Isn't the actual clone address 
`https://github.com/bombomby/optick.git`

Gems/ExternalProfilers/OptickProfiler/Code/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+    GIT_TAG "8abd28dee1a4034c973a3d32cd1777118e72df7e" # version 1.4.0+
+)
+
+set(OPTICK_ENABLED ON)
+set(OPTICK_BUILD_GUI_APP OFF)
+set(OPTICK_INSTALL_TARGETS OFF)
+
+FetchContent_MakeAvailable(Optick)
+
```
nick-l-o3de, line 22: Users would probably appreciate if you did something like
```cmake
message(STATUS "OptickProfiler Gem uses https://github.com/bombomby/optick.git version 1.4.0 (License: MIT)")
```

Gems/ExternalProfilers/SuperluminalProfiler/Code/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+list(APPEND CMAKE_MODULE_PATH ${SUPERLUMINAL_API_PATH})
+
```
nick-l-o3de, line 11: as above it might be worth doing a message indicating that its a non free license, maybe even before it tries to find_package so that if there is an error finding package they go to the commercial site?

Gems/ExternalProfilers/TracyProfiler/Code/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+#
+
+# Add Tracy dependency using FetchContent
+include(FetchContent)
+FetchContent_Declare(
+    Tracy
+    GIT_REPOSITORY "https://github.com/wolfpld/tracy"
+    GIT_TAG "5d542dc09f3d9378d005092a4ad446bd405f819a" # version 0.11.1
+)
+
```
nick-l-o3de, line 16: as usual, I'd highly recommend
```cmake
message(STATUS "TracyProfiler gem uses https://github.com/wolfpld/tracy version 0.11.1 (License: BSD-3-Clause)")
```

Gems/ExternalProfilers/TracyProfiler/Code/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+set(TRACY_ENABLE ON)
+set(TRACY_CALLSTACK ON)
+set(TRACY_ON_DEMAND ON)
+ly_append_configurations_options(
+    DEFINES
+        TRACY_ENABLE
```
nick-l-o3de, line 22: Isn't this super dangerous, like, doesn't ly_append_configurations_options set this option on for every single compilation unit in the entire engine, which requires a full recompile of the entire engine?

If so that would make it impossible to ever have this module included with an install or prebuilt since turning it on requires a full rebuild of the engine.

But I don't see anywhere in the o3de engine code that uses this define?  where does it actually need to be  defined (what targets?)  You can always add it as target_compile_options() to specific targets, after fetchcontent_makeavailable

Gems/ExternalProfilers/OptickProfiler/Code/CMakeLists.txt
guillaume-haerinck: Yes it was working but you're right better clean it up

Gems/ExternalProfilers/TracyProfiler/Code/CMakeLists.txt
guillaume-haerinck: Yeah it was by purpose but wasn't doing anything at the end. Its the Tracy documentation which outlines to define that for every files and as sampling wasn't working on windows I thought it might be related, turns out it is not so I removed it altogether

guillaume-haerinck:
> Looks good for the most part. Some minor comments. We should add also add info on how to use the profilers. If not on the main documentation web site then atleast on the Github wiki.

I'm about to put a PR up regarding documentation on o3de.org

guillaume-haerinck:
I would also need to modify the `repo.json` to allow the external profiler gems to be downloaded individually, but I am not sure on how to fill up some fields such as the Hash for the .zip

nick-l-o3de (approved):

moudgils (approved):

PR #839, opened 2025-03-07
Add gripper's initial position setting in FingerGripperComponent

knicked:
## What does this PR do?

This PR adds an editor field allowing to change the initial position of the gripper in the FingerGripperComponent, which was previously hardcoded as `0.0f`. This would cause issues if the controller made a small error and set the position to a negative value close to zero - when trying to execute any trajectory using moveit, it would complain about it with the following error and reject to execute it:
image

image

## How was this PR tested?
In the ROS2 Robotic Manipulation template, in the Finger Gripper Component of the manipulator, various values were put into the "initial position" field and the gripper was put in those positions correctly when starting game mode.
image

image
image
image

### DIFF
--- a/Gems/ROS2/Code/Source/Gripper/FingerGripperComponent.cpp
+++ b/Gems/ROS2/Code/Source/Gripper/FingerGripperComponent.cpp
@@ -57,7 +57,8 @@ namespace ROS2
                 ->Field("VelocityEpsilon", &FingerGripperComponent::m_velocityEpsilon)
                 ->Field("DistanceEpsilon", &FingerGripperComponent::m_goalTolerance)
                 ->Field("StallTime", &FingerGripperComponent::m_stallTime)
-                ->Version(1);
+                ->Field("InitialPosition", &FingerGripperComponent::m_initialPosition)
+                ->Version(2);
 
             if (AZ::EditContext* ec = serialize->GetEditContext())
             {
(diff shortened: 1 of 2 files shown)


Gems/ROS2/Code/Source/Gripper/FingerGripperComponent.h
```diff
@@ -80,5 +80,6 @@ namespace ROS2
         float m_velocityEpsilon{ 0.01f }; //!< The epsilon value used to determine whether the gripper is moving
         float m_goalTolerance{ 0.001f }; //!< The epsilon value used to determine whether the gripper reached it's goal
         float m_stallTime{ 0.1f }; //!< The time in seconds to wait before determining the gripper is stalled
+        float m_initialPosition{ 0.0f };
```
jhanca-robotecai, line 83: Please add a comment to match the other declarations.

Gems/ROS2/Code/Source/Gripper/FingerGripperComponent.cpp
```diff
@@ -81,7 +82,12 @@ namespace ROS2
                         AZ::Edit::UIHandlers::Default,
                         &FingerGripperComponent::m_stallTime,
                         "Stall Time",
-                        "The time to wait before considering the gripper as stalled.");
+                        "The time to wait before considering the gripper as stalled.")
+                    ->DataElement(
+                        AZ::Edit::UIHandlers::Default,
+                        &FingerGripperComponent::m_initialPosition,
+                        "Initial Position",
+                        "The initial position of the gripper.");
```
jhanca-robotecai, line 90: Please add a comment about the unit. 
Is the value limited anyhow (e.g. the positive numbers only)? If so, please use the `Attribute` to secure the user from setting the value incorrectly.

knicked: I've added a comment about the units.
The value of this parameter will be passed to JointsManipulationComponent and then down to the PhysX bus, which will apply the limits of each joint of the gripper (there might be more than one joint and they might have different limits - it doesn't sound reasonable to have different limits when all the joints are controlled by the same value, but currently nothing restricts the user from doing that). Therefore in my opinion no value limiting should happen in this part of the interface.

michalpelka (approved):

jhanca-robotecai (approved):
Works as expected, thank you for the contribution

PR #847, opened 2025-03-28
Move sensors from ROS2 Gem into new ROS2Sensors

jhanca-robotecai:
## What does this PR do?

**Note:** this PR targets the feature branch, not the `development` branch.

This PR creates a new Gem `ROS2Sensors` and moves the implementation of the sensors away from the `ROS2` Gem. This is a part of the RFC: https://github.com/o3de/sig-simulation/pull/96

What was done in detail:
- creating a new Gem
- clang-formatting the code; removing the obsolete auto-generated bus
- adding O3DE license header to all files
- moving the implementation from `ROS2` to the new Gem
- moving icons from `ROS2` to the new Gem

What was done additionally:
- sensor hooks and model hooks were disabled (commented-out) as suggested in the RFC; this part will be recovered after the whole implementation is ready
- GUIDs of the components were moved to a public interface file
- `CMakeLists.txt` in the new Gem was made in the different way than previously (all components other than lidar are now structured as in other O3DE Gems, i.e. with a proper split into public API and private static implementation; the lidar requires some rework that is out of the scope of this RFC)

What is still missing for this part of the work:
- `README` was not updated - it will be done after all changes in `ROS2` Gem
- unused dependencies in `ROS2` Gem were not removed - it will be done after all changes in `ROS2` Gem
- changes in the documentation are missing - it will be done after all changes in `ROS2` Gem

## How was this PR tested?

This PR was tested using a sample project with all sensors enabled. The sensor data was checked in RViz2, the icons and the configuration of each sensor was checked in the Editor. The console was checked for warnings/errors. It was tested against engine of 2505.0 stabilization branch.

Note: this code was also tested against `RobotecGPULidar` implementation of the lidar, changes that will be required in the latter are drafted in https://github.com/RobotecAI/o3de-rgl-gem/pull/77


Gems/ROS2/Code/Source/Sensor/Events/PhysicsBasedSource.cpp
```diff
@@ -8,9 +8,8 @@
 
 #include <AzFramework/Physics/PhysicsSystem.h>
 
-#include <ROS2/Utilities/ROS2Conversions.h>
 #include <ROS2/Sensor/Events/PhysicsBasedSource.h>
-#include <ROS2/Sensor/SensorConfiguration.h>
+#include <ROS2/Utilities/ROS2Conversions.h>
```
patrykantosz, line 13: Please add 
`#include <AzCore/Serialization/SerializeContext.h>`
Without this header non-unity build fails during compilation.

Gems/ROS2/Code/Source/Sensor/Events/TickBasedSource.cpp
```diff
@@ -1,16 +1,15 @@
...
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
 
-#include <ROS2/Sensor/Events/TickBasedSource.h>
-#include <ROS2/Sensor/SensorConfiguration.h>
+#include <AzCore/std/numeric.h>
 #include <ROS2/ROS2Bus.h>
+#include <ROS2/Sensor/Events/TickBasedSource.h>
 #include <ROS2/Utilities/ROS2Conversions.h>
```
patrykantosz, line 12: Please add 
`#include <AzCore/Serialization/SerializeContext.h>`
Without this header non-unity build fails during compilation.

patrykantosz (commented):
Few minor comments. I did not have time to perform runtime tests of the sensors after split.

Gems/ROS2Sensors/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.h
```diff
@@ -10,7 +10,8 @@
 
 #include <AzCore/Component/Component.h>
 #include <AzCore/RTTI/TypeInfoSimple.h>
-#include <ROS2/Camera/CameraPostProcessingRequestBus.h>
+#include <ROS2Sensors/Camera/CameraPostProcessingRequestBus.h>
+#include <ROS2Sensors/ROS2SensorsTypeIds.h>
 
 namespace ROS2
```
arturkamieniecki, line 16: It seems that the UUIDs for sensors were moved into the ROS2Sensors namespace but the implementation is still in the ROS2 namespace. Please verify if this is intentional.

This comment applies to all other instances of ROS2 namespace in ROS2Sensors gem.

arturkamieniecki (commented):
Looks good. I have not yet built the new gem.

Gems/ROS2/Code/Source/SystemComponents/ROS2SystemComponent.cpp
```diff
@@ -58,7 +53,8 @@ namespace ROS2
             {
                 ec->Class<ROS2SystemComponent>(
                       "ROS 2 System Component",
-                      "This component is responsible for creating ROS 2 node and executor, provides ROS 2 interfaces, manages ROS 2 clock and "
+                      "This component is responsible for creating ROS 2 node and executor, provides ROS 2 interfaces, manages ROS 2 clock "
+                      "and "
                       "publishes transforms.")
```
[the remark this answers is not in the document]
jhanca-robotecai: it is, indeed, weird. This is an outcome of `clang-format`, which splits the lines that are too long, but it does not joint the short ones. Fixed.

Gems/ROS2Sensors/Code/Source/Camera/PostProcessing/ROS2ImageEncodingConversionComponent.h
jhanca-robotecai: I am still about to make my mind about the namespaces. Primarily, I wanted to keep the `ROS2` namespace whenever the code is copied. But you are right, that the current mix makes no sense. I am open to comments.

arturkamieniecki (approved):
I've compiled the code and tested it a little (only some sensors), looks good!

jhanca-robotecai:
I added the comment about the namespace into RFC: https://github.com/o3de/sig-simulation/pull/96#pullrequestreview-2759776395 , so it is not lost. I will merge this PR to enable the follow-up PR with splitting `ROS2Controllers` out of `ROS2` Gem.

PR #864, opened 2025-04-08
[Simulation Interfaces] Adjust error code type + ROS 2 frame namespace

michalpelka:
## What does this PR do?

Adjust errror codes to allow some extensatbility.

## How was this PR tested?

- run unit tests.


norbertprokopiuk (approved):

michalpelka:
It will not work. I've tested it.
Setting namespace after activation will cause artifacts in tf frame.
image

PR #871, opened 2025-04-09
Nprokopiuk/simulation interfaces set simulation state

norbertprokopiuk:
## What does this PR do?

This PR adds possibility to get and set simulation state for both SimulationInterfaces and SimulationInterfacesROS2 Gems.

## How was this PR tested?

Calls to created ros2 services via rqt


michalpelka (approved):

michalpelka:
Tested manually, works as inteded.

PR #882, opened 2025-04-11
[simulation interfaces] Changed default behavior of the SimulationEntitiesManager to registry only main entity.

michalpelka:
## What does this PR do?

Changed default behavior of the SimulationEntitiesManager to registry only main entity.

We can register part of spawned robot, but it needs special design of the prefab. Advanced user can enable this behavior by setting the registry to:
```
"SimulationInterfaces":
{
    "RegisterEverySimulatedEntity" : true
}
```
Also naming strategy was adjusted to be more predictable.

**Default behavior**
Only main robot's entity is registered:
image
**RegisterEverySimulatedEntity true behavior**
Robots parts are registered as well.
image

## How was this PR tested?

- manual tests
- run existing tests.


norbertprokopiuk (commented):
Overall it looks fine, solves one issue, but unfortunately introduces new one. It improves registration when prefab has structure where there is only one simulated body per hierarchy level. If there are simulated bodies on the same level in the hierarchy, default behaviour registers only the one which was spawned first. I causes that only this one entity can be modified and there is no way to modify other ones. I think it should somehow handled in the default case or at least notified somewhere, because it is kind of unexpected for the user.

PR #886, opened 2025-04-14
Moved sensor assets from ROS2 Gem to ROS2SampleRobots Gem

pawelbudziszewski:
## What does this PR do?

Moved meshes, materials, and textures of sensors from ROS2 gem to ROS2SampleRobots gem (which already contains assets of 2 lidars)

This PR is in line with the changes proposed in [RFC for simulation-related gems and templates](https://github.com/o3de/sig-simulation/pull/94)

## How was this PR tested?

Checked if prefabs load without warnings


jhanca-robotecai (approved):
There is some inconsistency in the naming: `LidarOS2` vs. `OrbbeckCamera` (should be `CameraOrbbeck`), but I would not care that much at this point.

PR #909, opened 2025-05-12
Fix LevelGeoreferencing Editor tests

jhanca-robotecai:
## What does this PR do?

This PR fixes the Editor tests in `LevelGeoreferencing` Gem. They used to fail due to the incorrect call `AddActiveGems` - the name of the Gem changed, but not everywhere.

Besides cleaning this one issue, a few more renames around the _define_ (`PAL_TRAIT`) names were to ensure the code can be executed against the SDK.

## How was this PR tested?

Tests were called against O3DE `stabilization/25050`.


MateuszWasilewski (approved):
Everything works correctly :+1:
I tested the change on an empty project.
```bash
cmake --build Project/build/linux --config profile --target LevelGeoreferencing.Tests LevelGeoreferencing.Editor.Tests
ctest -C profile --verbose --output-on-failure --test-dir Project/build/linux --tests-regex 'Georeferencing'
```

pawelbudziszewski (approved):

PR #910, opened 2025-05-12
Add FrameSystemComponent to AssetBuilder

jhanca-robotecai:
## What does this PR do?

Edit: The text was updated following the update of the code, as explained below.

The implementation in ROS2Gem was not available in AssetBuilder. That said, it was not able to activate the component due to the missing implementation when importing URDF/SDF files (it might want to do so, e.g., when baking cylinder mesh while importing colliders). As a result, the asset builder was not able to import the model. This PR adds a missing system component to the asset builder. 

## How was this PR tested?
- executed tests
- changed namespace types (empty/custom/entity_name) within the Editor a few times
- created an entity within another entity to verify that the parent namespace was read correctly
- moved the entity with a frame component between two different parents to verify that the parent namespace was read correctly

### DIFF
--- a/Gems/ROS2/Code/Source/Frame/ROS2FrameSystemBus.h
+++ b/Gems/ROS2/Code/Source/Frame/ROS2FrameSystemBus.h
@@ -7,13 +7,13 @@
  */
 #pragma once
 
+#include <AzCore/Component/EntityId.h>
+#include <AzCore/EBus/EBus.h>
 #include <AzCore/EBus/Policies.h>
 #include <AzCore/Interface/Interface.h>
 #include <AzCore/RTTI/RTTIMacros.h>
 #include <AzCore/std/string/string.h>
 #include <ROS2/Frame/ROS2FrameComponent.h>
-#include <AzCore/Component/EntityId.h>
-#include <AzCore/EBus/EBus.h>
 
 namespace ROS2
 {
@@ -66,7 +66,7 @@ namespace ROS2
     {
     public:
         static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
-        static constexpr AZ::EBusHandlerPolicy AddressPolicy = AZ::EBusHandlerPolicy::Single;
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
     };
 
     using ROS2FrameSystemInterface = AZ::Interface<ROS2FrameSystemRequests>;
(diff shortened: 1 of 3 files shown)


norbertprokopiuk (approved):
Code looks good. I tested the solution in the editor and with ros2 frame viewer and everything works as intended

Gems/ROS2/Code/Source/Frame/ROS2FrameSystemComponent.h
```diff
@@ -53,7 +53,7 @@ namespace ROS2
     //! Used to register, unregister, track the frame entities in the level entity tree.
     class ROS2FrameSystemComponent
         : public AZ::Component
-        , public ROS2FrameSystemInterface::Registrar
```
arturkamieniecki, line 56: There should be no need for the complete removal of the interface. (https://docs.o3de.org/docs/user-guide/programming/messaging/az-interface/#converting-from-an-ebus-implementation)

Gems/ROS2/Code/Source/Frame/ROS2FrameSystemBus.h
```diff
@@ -66,9 +66,9 @@ namespace ROS2
     {
     public:
         static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
-        static constexpr AZ::EBusHandlerPolicy AddressPolicy = AZ::EBusHandlerPolicy::Single;
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::Single;
     };
 
     using ROS2FrameSystemInterface = AZ::Interface<ROS2FrameSystemRequests>;
```
arturkamieniecki, line 72: If the interface is fully removed then there is no need for this define.

michalpelka (approved):
Made own test, did not find any issue in code. LGTM

michalpelka:
Good point @nick-l-o3de - checking for nullptr would be enough to activate those entities. The communication via this bus could be ignored in tool context (sdf builder, but for now). However, in my opinion (after second thought), we should create ROS2FrameComponent also in context of SDF builder so builder would be able to use all available API in future.

nick-l-o3de:
So we could submit this if you like, or replace it?  What would you want?

jhanca-robotecai:
Following the Discord discussion, it was decided to keep the interface instead of a bus. Additionally, the check for a `nullptr` was added just in case. The system component was added to the `AssetBuilder` to enable the `Activation` from the _AssetProcessor_.

michalpelka (approved):

PR #916, opened 2025-05-26
Merge `stabilization/25050` into `main`

jhanca-robotecai:
## What does this PR do?

This is a release PR for O3DE `2505.0` Release.

This PR was created following the procedure in https://github.com/o3de/o3de/wiki/How-to-merge-from-stabilization-branches-to-main-for-release

## How was this PR tested?

Diff between this branch and the `stabilization/25050` branch was checked.
https://github.com/o3de/o3de-extras/compare/stabilization/25050...RobotecAI:o3de-extras:jh/release_25050?expand=1

PR #919, opened 2025-05-29
Remove XR Test project from repo.json

spham-amzn:
## What does this PR do?

This removes the OpenXRTest project from repo.json so that it does not appear in the Project Manager automatically.

## How was this PR tested?

Can only validate once in canonical.o3de.org

### DIFF
--- a/repo.json
+++ b/repo.json
@@ -11580,50 +11580,5 @@
         }
     ],
     "projects_data": [
-        {
-            "project_name": "OpenXRTest",
-            "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


jhanca-robotecai (approved):
The changes look valid. There are a few more differences between the `repo.json` file in this repository and in https://github.com/o3de/canonical.o3de.org, but I planned to fix them after the release. Please correct me if I am wrong, but I understand the file in this repository is effectively ignored, and the one from https://github.com/o3de/canonical.o3de.org is used?

This change should be cherry-picked to #916 if we want to really get it into the release.

PR #932, opened 2025-07-10
[WarehouseAssets] Remove leftover material slot overrides

alek-kam-robotec-ai:
## What does this PR do?

This PR removes unnecessary, leftover material slot overrides from the Warehouse Assets gem's prefabs. This change is related to the new [RGL gem release](https://github.com/RobotecAI/o3de-rgl-gem/pull/81). Removed overrides referred to non-existent material slots.

## How was this PR tested?

Modified prefabs were tested in the editor to ensure proper material overrides. They were tested with the RGL gem active.

### DIFF
--- a/Gems/ROS2SampleRobots/Assets/RosbotXL/ROSbotXL.prefab
+++ b/Gems/ROS2SampleRobots/Assets/RosbotXL/ROSbotXL.prefab
@@ -602,16 +602,20 @@
                     "Id": 2016576711781101659,
                     "Controller": {
                         "Configuration": {
-                            "materials": {
-                                "{}": {
-                                    "MaterialAsset": {
-                                        "assetId": {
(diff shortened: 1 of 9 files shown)


jhanca-robotecai (changes requested):
The change works as expected. 

Marking it as request changes for the following reasons:
- [x] we need an approval from sig-release to target branch to the point release branch
- [x] we should verify if there are any other prefabs with this problem
- [x] we should bump the version of the Gem and prepare it for the release (including `repo.json` files)

jhanca-robotecai:
Checked all other prefabs for the same issue; bumped versions.
`repo.json` will be prepared after this is accepted and merged (the push to the cannonical is also needed)

matteogrs (approved):
`sig-release` approves to fix the assets manually for this first point release

jhanca-robotecai (approved):

PR #935, opened 2025-07-15
Bring back DAE

michalpelka:
## What does this PR do?

The DAE support was enabled in deleted registry setting.

## How was this PR tested?

Against project that has robot with DAE files in their URDF.

### DIFF
--- a/Gems/ROS2RobotImporter/Registry/sceneassetimporter.setreg
+++ b/Gems/ROS2RobotImporter/Registry/sceneassetimporter.setreg
@@ -1,0 +1,39 @@
+{
+    "O3DE":
+    {
+        "SceneAPI":
+        {
+            "AssetImporter":
+            {
+            	// This list controls the set of file types that will get processed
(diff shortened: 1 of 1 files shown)


Gems/ROS2/Registry/sceneassetimporter.setreg
```diff
@@ -0,0 +1,40 @@
...
+            	// the source asset to automatically rebuild itself.
+                "MetaDataTypes": {
+                    "dae.assetinfo": "dae",
+                    "obj.assetinfo": "obj"
+                }
+            }
+        }
+    }
+}
+}
```
jhanca-robotecai, line 40: Please double-check the openings/closings of the bracelets.

I would also consider pushing this `setreg` to `ROS2RobotImporter` Gem. I see no reason why `ROS2` Gem would change the behavior of the _SceneAPI_.

jhanca-robotecai (approved):

PR #942, opened 2025-07-23
Fix missing textures in manipulator template

michalpelka:
## What does this PR do?

Fix missing texture in template

## How was this PR tested?
I've backported fixes from generated project to template


jhanca-robotecai (approved):
The missing files were not visible on the scene, but triggered warning message about missing textures. This PR fixes the warning messages.

I additionally bumped the template version and added files to `template.json`, hence they can be copied while creating a new project out of a template.

Note, that the zip file is not available yet, it will be generated during the release.

PR #951, opened 2025-08-06
ROS2 API improvements: PublisherConfiguration, QoS, TopicConfiguration, TickBasedSource, PhysicsBasedSource

jhanca-robotecai:
## What does this PR do?

The ultimate target of `ROS2` Gem's API refactor is to get rid of the `Gem::ROS2.Static` object that is linked instead of the `Gem::ROS2.API` target.

As a next step in cleaning-up the API of `ROS2` Gem the following changes were implemented:
- converted `ROS2Names` class into a bus handled by the system component. Rationale: the handler will always be available if the system component is available (if the system component is not available, the whole stack will terminate when looking for a ROS 2 node before creating a service/action/publisher/subscriber).
- moved implementation from `PublisherConfiguration.cpp`, `QoS.cpp` and `TopicConfiguration.cpp` into header files. Rationale: static `Reflect` methods in the codebase will not slow down the build too much, while the implementation is needed in other Gems.
- ~~moved implementation from `PhysicsBasedSource.cpp` and `TickBasedSource.cpp` into header files. Rationale: the files are included only by the sensors, hence the build will not slow down too much, while the implementation is needed in other Gems.~~ Edit: this part was removed from this PR, it will be a part of another PR

Besides, some small changes took place:
- moved some objects from the static target into private static target to simplify the linkage
- removed some obsolete includes

## How was this PR tested?

The most important part, i.e., changes in linking, are verified by the successful build (+passing tests). Additionally I created an entity with the name which does not follow the naming conventions and the `ROS2FrameEditorComponent` triggered a warning message about the incorrect name. Next, the same warning was triggered in the _game mode_, proving `ROS2FrameComponent` can use the bus correctly. The _TickBasedSource_ and _PhysicsBasedSource_ changes were tested by adding some ROS 2 sensors to the scene.

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Communication/QoS.h
+++ b/Gems/ROS2/Code/Include/ROS2/Communication/QoS.h
@@ -63,7 +63,7 @@ namespace ROS2
                         ->DataElement(AZ::Edit::UIHandlers::Default, &QoS::m_depth, "History depth", "Determines DDS publisher queue size");
                 }
             }
-        };
+        }
 
         //! Convert and return a rclcpp::QoS instance based on member values (Editor combos selection).
         //! @return a <a href="https://docs.ros2.org/latest/api/rclcpp/classrclcpp_1_1QoS.html">ROS2 QoS struct</a>.
@@ -74,7 +74,8 @@ namespace ROS2
         }
 
     private:
-        AZ::Crc32 OnQoSSelected()
+        //! Helper function for the UI to refresh attributes and values when QoS is selected.
+        AZ::Crc32 OnQoSSelected() const
         {
             return AZ::Edit::PropertyRefreshLevels::AttributesAndValues;
         }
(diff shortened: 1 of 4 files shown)


Gems/ROS2/Code/Source/Tools/ROS2EditorSystemComponent.cpp
```diff
@@ -65,21 +65,26 @@ namespace ROS2
...
     void ROS2EditorSystemComponent::Deactivate()
     {
+        ROS2NamesRequestBus::Handler::BusDisconnect();
         AzToolsFramework::EditorEvents::Bus::Handler::BusDisconnect();
         AzToolsFramework::EditorEntityContextNotificationBus::Handler::BusDisconnect();
     }
 
     void ROS2EditorSystemComponent::OnStartPlayInEditorBegin()
     {
+        ROS2NamesRequestBus::Handler::BusDisconnect(); // ROS2NamesRequestBus will be handled by ROS2SystemComponent
```
michalpelka, line 80: I think this API should be available in Editor as well

michalpelka (approved):

Gems/ROS2/Code/Source/Tools/ROS2EditorSystemComponent.cpp
jhanca-robotecai: The API is available in the Editor, as the the bus is connected in `Activate()`. It is disconnected here, as the same functionality is taken over by `ROS2SystemComponent`. Next, after Game Mode is terminated, the bus is connected again in `OnStopPlayInEditor` method.

PR #952, opened 2025-08-07
Added "Force dynamic" switch to the ros2 frame component

patrykantosz:
## What does this PR do?

This PR adds an additional switch to the `ROS 2 Frame` component that forces it to be a dynamic frame. This option is disabled by default.
## How was this PR tested?

I used different prefabs with ROS 2 frames, changed their configuration (mainly the dynamic force switch), and checked:
- rviz2
- `/tf` topic
- `/tf_static` topic

### DIFF
--- a/Gems/ROS2/gem.json
+++ b/Gems/ROS2/gem.json
@@ -1,6 +1,6 @@
 {
     "gem_name": "ROS2",
-    "version": "4.0.0",
+    "version": "4.0.1",
     "display_name": "ROS2",
     "license": "Apache-2.0",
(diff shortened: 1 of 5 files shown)


Gems/ROS2/gem.json
```diff
@@ -1,6 +1,6 @@
 {
     "gem_name": "ROS2",
-    "version": "4.0.0",
+    "version": "4.0.1",
```
norbertprokopiuk, line 3: This PR adds new feature so version 4.1.0 seems more suitable 
```suggestion
    "version": "4.1.0",
```

jhanca-robotecai (approved):

PR #961, opened 2025-08-25
[Simulation Interfaces] Do not change applied state on set_entity_state request.

michalpelka:
## What does this PR do?
The scale was cleared to one, when state of entity was applied.

## How was this PR tested?

I've fixed this issue in project I worked on.

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
@@ -491,7 +491,10 @@ namespace SimulationInterfaces
             {
                 AZ::Transform parentTransform;
                 AZ::TransformBus::EventResult(parentTransform, parentEntityId, &AZ::TransformBus::Events::GetWorldTM);
+                float scale = 1.0f;
+                AZ::TransformBus::EventResult(scale, entityId, &AZ::TransformBus::Events::GetLocalUniformScale);
                 auto transformToSet = parentTransform.GetInverse() * state.m_pose;
+                transformToSet.SetUniformScale(scale);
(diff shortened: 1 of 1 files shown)


Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -491,7 +491,10 @@ namespace SimulationInterfaces
             {
                 AZ::Transform parentTransform;
                 AZ::TransformBus::EventResult(parentTransform, parentEntityId, &AZ::TransformBus::Events::GetWorldTM);
+                float scale = 1.0f;
+                AZ::TransformBus::EventResult(scale, entityId, &AZ::TransformBus::Events::GetLocalUniformScale);
```
pijaro, line 495: What about non-uniform scale? Maybe put some note in a readme (or somewhere else) that non-uniform scaled objects are not 100% supported?

jhanca-robotecai: The code is correct.

The uniform scale is a part of a transform in O3DE. It is reset when loading parent's tranform (needed to calculate the relative position and orientation). The scale of the entity has to be applied back.

The non-uniform scale is a separate component, that is added on top of the transform. It stayed unchanged when calling `AZ::TransformBus`.

jhanca-robotecai (approved):
Works as expected.

Base scene:
Screenshot from 2025-08-25 17-28-30



I moved the red object (uniform scale) and the blue object (non-uniform scale) by vector {0,2,0}. 
Original code:
Screenshot from 2025-08-25 17-29-00

This PR:
Screenshot from 2025-08-25 17-30-30

PR #968, opened 2025-08-28
[ROS 2] Rework ROS 2 frame 

michalpelka:
## What does this PR do?
This PR is major rework of ROS 2 frame system in O3DE.

After years of using existing system we propose an almost complete re-write of the system.
Old system was non-flexible, implementation was unclean and did not follow best-practices. It was close to untestable spaghetti code.

Main point of the whole PR is centralization of the logic for namespace "computation" function 'ComputeNamespace' defined in  `Code/Source/Frame/NamespaceComputation.h` and it unit-tests in `Code/Tests/Frame/ROS2FrameComponentTest.cpp`. That centralized logic is easy to use and understand.

However some new features where added (mainly for building and modifying TFs in game mode):

### Public Interface ROSFrameInterface for ROS2FrameComponent. 
This API allows to change configuration for disabled component (e.g. during pre-spawning) or creating a TF tree from scratch in game mode.

### Public API `ROS2FrameComponentBus` 
API that allows to get information from an active `ROS2FrameComponent`.

### Public API `ROS2FrameTrackingInterface`
API that allows you to find corresponding entities to ROS2Frames and vice versa.

### Registry setting `/O3DE/ROS2/GlobalFrameName`
This registry setting allows to set root frame name. This allows to choose from names like `odom`, `world`, `map` depending on system under test. You can disable root frame publication (e.g. if you system under test is robot localization package). It allows to make simulation compatible to [REP-105](https://www.ros.org/reps/rep-0105.html.)

**Robot without namespace with  /O3DE/ROS2/GlobalFrameName not set**
```
 ros2 service call /spawn_entity simulation_interfaces/srv/SpawnEntity "{ 
  name: 'foo',
  allow_renaming: false,
  uri: 'product_asset:///rosbotxl/rosbotxl.spawnable'
 }"
```
| image
|---------------------------------------------------------------------------------------------------------------|

**Robot with namespace with  /O3DE/ROS2/GlobalFrameName not set**
```
ros2 service call /spawn_entity simulation_interfaces/srv/SpawnEntity "{
  name: 'foo',
  allow_renaming: false,
  entity_namespace: 'fooo',
  uri: 'product_asset:///rosbotxl/rosbotxl.spawnable'}"
```
image

**Robot with namespace with  /O3DE/ROS2/GlobalFrameName set to empty**
```
ros2 service call /spawn_entity simulation_interfaces/srv/SpawnEntity "{
  name: 'foo',
  allow_renaming: false,
  entity_namespace: 'fooo',
  uri: 'product_asset:///rosbotxl/rosbotxl.spawnable'}"
```
image

### Scripting tools

Publishing to ROS 2 TF from LUA/Script canvas
image

ROS2 frame component bus:
image

With this scripting API you can build complex scenarios, recreating and mocking parts of system under test.
```lua

local ScriptTest = 
{
    Properties =
    {
    }
}

function ScriptTest:OnActivate()
        self.tickBusHandler = TickBus.Connect(self)
        self.namespace = ROS2FrameComponentBus.Broadcast.GetNamespace();
        
end

function ScriptTest:OnDeactivate()
     self.tickBusHandler:Disconnect()
end

function ScriptTest:OnTick(deltaTime, timePoint)
     sourceFrame  = "world"
     targetFrame = self.namespace .."/map"
     transform = Transform.CreateIdentity()
     TFInterfaceBus.Broadcast.PublishTransform(sourceFrame,targetFrame,transform, false);
     sourceFrame = self.namespace .."/map"
     targetFrame = self.namespace .."/odom"
     transform = Transform.CreateIdentity()
     TFInterfaceBus.Broadcast.PublishTransform(sourceFrame,targetFrame,transform, false);
     
      self.tickBusHandler:Disconnect()
end

return ScriptTest
```
giving you this tf tree:
image

With this scripting API, with some LUA/SC effort you can do anything you want want with TF tree. Imagination is a limit!

### Tools for disabling and enabling entities without sudden jumps in transform

The design on applying parents in AZ::TransformComponent is clever. But clever in wrong way.
It tries to be smart with keeping world transform not changed when parenting/reparenting.

But, what will happen when we disable parent entity?
Local transform is changed to parent's world times child's local causing child to be in the same place.
However, when parent is activated, it will jump to location  parent's world times parent's world.
That seems to be a bug / corner case in O3DE. Since Robotics is only place where we need to deal with that, I've added helpers to safely disable, and enable entities.
Example of usage:
```lua
local DisableEnableEntity = 
{
    Properties =
    {
    	TestEntity = {default = EntityId()}
    }
}

function DisableEnableEntity:OnActivate()
	self.counter = 0
        self.tickBusHandler = TickBus.Connect(self)
        
end

function DisableEnableEntity:OnDeactivate()
     self.tickBusHandler:Disconnect()
end

function DisableEnableEntity:OnTick(deltaTime, timePoint)
	if self.counter == 10 then
		ROS2FrameTrackingRequestBus.Broadcast.DisableEntitySafely(self.Properties.TestEntity)
		Debug.Log("Disable")
	end
	if self.counter == 20 then
	
		ROS2FrameTrackingRequestBus.Broadcast.EnableEntitySafely(self.Properties.TestEntity)
		Debug.Log("Enable")
		self.counter = 0
	end
	self.counter = self.counter  + 1
end

return DisableEnableEntity
```

## How was this PR tested?

Unit tests:
image

Run existing project.

## Compatibility
 - The default setting should give you the same experience
 - Prefabs are compatible, but spawnables needs to be rebuilt.
 - Removed tools for legacy game only TF frames (pre 2409 version).

### DIFF
--- a/Gems/ROS2/Code/Include/ROS2/Frame/ROS2FrameTrackingInterface.h
+++ b/Gems/ROS2/Code/Include/ROS2/Frame/ROS2FrameTrackingInterface.h
@@ -30,7 +30,7 @@ namespace ROS2
 
         //! Get all currently registered frame entities.
         //! @return Set of EntityIds representing all registered frames
-        virtual const AZStd::unordered_set<AZ::EntityId>& GetRegisteredFrames() const = 0;
+        virtual const AZStd::unordered_set<AZ::EntityId>& GetRegisteredFrameEntityIds() const = 0;
 
         //! Check if a specific frame entity is registered.
         //! @param frameEntityId The EntityId to check
--- a/Gems/ROS2/Code/Source/Frame/ROS2FrameGameSystemComponent.cpp
+++ b/Gems/ROS2/Code/Source/Frame/ROS2FrameGameSystemComponent.cpp
@@ -92,7 +92,7 @@ namespace ROS2
         m_entityIdToFrameId.erase(frameEntityId);
     }
 
-    const AZStd::unordered_set<AZ::EntityId>& ROS2FrameGameSystemComponent::GetRegisteredFrames() const
+    const AZStd::unordered_set<AZ::EntityId>& ROS2FrameGameSystemComponent::GetRegisteredFrameEntityIds() const
     {
         return m_registeredEntities;
     }
--- a/Gems/ROS2/Code/Tests/Frame/ROS2FrameComponentTest.cpp
+++ b/Gems/ROS2/Code/Tests/Frame/ROS2FrameComponentTest.cpp
@@ -331,7 +331,7 @@ namespace UnitTest
 
         // Initially no frames should be registered
         EXPECT_EQ(trackingInterface->GetRegisteredFrameCount(), 0);
-        EXPECT_TRUE(trackingInterface->GetRegisteredFrames().empty());
+        EXPECT_TRUE(trackingInterface->GetRegisteredFrameEntityIds().empty());
         EXPECT_TRUE(trackingInterface->GetAllNamespacedFrameIds().empty());
 
         // Create and activate a frame component
@@ -349,9 +349,9 @@ namespace UnitTest
         // Frame should now be registered
         EXPECT_EQ(trackingInterface->GetRegisteredFrameCount(), 1);
         EXPECT_TRUE(trackingInterface->IsFrameRegistered(entity.GetId()));
-        EXPECT_FALSE(trackingInterface->GetRegisteredFrames().empty());
-        EXPECT_EQ(trackingInterface->GetRegisteredFrames().size(), 1);
-        EXPECT_TRUE(trackingInterface->GetRegisteredFrames().contains(entity.GetId()));
+        EXPECT_FALSE(trackingInterface->GetRegisteredFrameEntityIds().empty());
+        EXPECT_EQ(trackingInterface->GetRegisteredFrameEntityIds().size(), 1);
+        EXPECT_TRUE(trackingInterface->GetRegisteredFrameEntityIds().contains(entity.GetId()));
 
         // Check namespaced frame ID lookup
         auto namespacedFrameId = trackingInterface->GetNamespacedFrameId(entity.GetId());
@@ -372,7 +372,7 @@ namespace UnitTest
 
         EXPECT_EQ(trackingInterface->GetRegisteredFrameCount(), 0);
         EXPECT_FALSE(trackingInterface->IsFrameRegistered(entity.GetId()));
-        EXPECT_TRUE(trackingInterface->GetRegisteredFrames().empty());
+        EXPECT_TRUE(trackingInterface->GetRegisteredFrameEntityIds().empty());
         EXPECT_TRUE(trackingInterface->GetAllNamespacedFrameIds().empty());
     }
 
@@ -406,7 +406,7 @@ namespace UnitTest
 
         // Check all frames are registered
         EXPECT_EQ(trackingInterface->GetRegisteredFrameCount(), numFrames);
-        EXPECT_EQ(trackingInterface->GetRegisteredFrames().size(), numFrames);
+        EXPECT_EQ(trackingInterface->GetRegisteredFrameEntityIds().size(), numFrames);
         EXPECT_EQ(trackingInterface->GetAllNamespacedFrameIds().size(), numFrames);
 
         // Check each frame individually
@@ -434,7 +434,7 @@ namespace UnitTest
         entities[0]->Deactivate();
         entities[2]->Deactivate();
         EXPECT_EQ(trackingInterface->GetRegisteredFrameCount(), 0);
-        EXPECT_TRUE(trackingInterface->GetRegisteredFrames().empty());
+        EXPECT_TRUE(trackingInterface->GetRegisteredFrameEntityIds().empty());
     }
 
     TEST_F(ROS2FrameComponentFixture, FrameTrackingInterfaceEdgeCases)
--- a/Gems/ROS2/Code/Source/Clients/ROS2SystemComponent.cpp
+++ b/Gems/ROS2/Code/Source/Clients/ROS2SystemComponent.cpp
@@ -9,8 +9,7 @@
 
 #include "ROS2SystemComponent.h"
 
-#include "ROS2/Clock/ROS2ClockRequestBus.h"
-
+#include <ROS2/Clock/ROS2ClockRequestBus.h>
 #include <ROS2/Communication/PublisherConfiguration.h>
 #include <ROS2/Communication/QoS.h>
 #include <ROS2/Communication/TopicConfiguration.h>
--- a/Gems/ROS2/Code/Source/Frame/ROS2FrameGameSystemComponent.h
+++ b/Gems/ROS2/Code/Source/Frame/ROS2FrameGameSystemComponent.h
@@ -41,7 +41,7 @@ namespace ROS2
         void UnregisterFrame(const AZ::EntityId& frameEntityId) override;
 
         // ROS2FrameTrackingRequestBus::Handler
-        const AZStd::unordered_set<AZ::EntityId>& GetRegisteredFrames() const override;
+        const AZStd::unordered_set<AZ::EntityId>& GetRegisteredFrameEntityIds() const override;
         bool IsFrameRegistered(const AZ::EntityId& frameEntityId) const override;
         size_t GetRegisteredFrameCount() const override;
         AZ::EntityId GetFrameEntityByNamespacedId(const AZStd::string& namespacedFrameId) const override;


Gems/ROS2/Code/Source/Spawner/ROS2SpawnerComponent.cpp
```diff
@@ -301,7 +301,8 @@ namespace ROS2
                 entity->SetName(instanceName);
                 if (!spawnableNamespace.empty())
                 {
-                    frameComponent->UpdateNamespaceConfiguration(spawnableNamespace, NamespaceConfiguration::NamespaceStrategy::Custom);
+                    // TODO Mpelka - fix
```
michalpelka, line 304: Need implementation

arturkamieniecki: Implemented in https://github.com/RobotecAI/o3de-extras/pull/69

Gems/ROS2RobotImporter/Code/Source/RobotImporter/SDFormat/ROS2SDFormatHooksUtils.cpp
```diff
@@ -203,8 +203,8 @@ namespace ROS2RobotImporter::SDFormat
         }
         else if (!remappedNamespace.empty())
         {
-            frameConfiguration.m_namespaceConfiguration.SetNamespace(
-                remappedNamespace, ROS2::NamespaceConfiguration::NamespaceStrategy::Custom);
+            // frameConfiguration.m_namespaceConfiguration.SetNamespace(
```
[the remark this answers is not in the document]
arturkamieniecki: Implemented.

arturkamieniecki:
Something is not working right when using the `JointsPositionsEditorComponent` and clicking on find all joints. This needs verification.

Gems/ROS2/Code/Source/Frame/NamespaceComputation.cpp
```diff
@@ -0,0 +1,220 @@
...
+        return name;
+    }
+
+    AZStd::optional<ROS2FrameConfiguration> GetConfigurationFromComponent(AZ::EntityId id)
+    {
+        AZ::Entity* entity = nullptr;
+        AZ::ComponentApplicationBus::BroadcastResult(entity, &AZ::ComponentApplicationRequests::FindEntity, id);
+        if (!entity)
+        {
+            return ROS2FrameConfiguration();
```
michalpelka, line 34:
```suggestion
            return AZStd::nullopt;
```

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -72,162 +76,129 @@ namespace ROS2
...
         {
-            JSR::ResultCode componentIdLoadResult = ContinueLoadingFromJsonObjectField(
-                &configInstance->m_publishTransform,
-                azrtti_typeid<decltype(configInstance->m_publishTransform)>(),
-                inputValue,
-                "Publish Transform",
-                context);
-
-            result.Combine(componentIdLoadResult);
+            frameSystemInterface->RegisterFrame(GetEntityId());
```
norbertprokopiuk, line 116: I would say that first registering frame then making it responsive for bus calls is better order. In that way when frame is responsive it is already available in all systems related to it

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -72,162 +76,129 @@ namespace ROS2
...
-
-        if (m_publishTransform)
+        if (m_sourceFrame == AZStd::nullopt)
         {
-            AZ_TracePrintf("ROS2FrameComponent", "Setting up %s", GetNamespacedFrameID().data());
-
-            // The frame will always be dynamic if it is a top entity or if its configuration forces it to be dynamic..
-            if (IsTopLevel() || m_forceDynamic)
+            // cache parent frame
+            const auto* ros2FrameComponent = Internal::GetFirstROS2FrameAncestor(GetEntity());
```
norbertprokopiuk, line 147: isn't `GetFirstROS2FrameAncestor` similar to one of the `GetLastEntityWithROS2FrameComponent` or `GetFirstEntityWithROS2FrameComponent`? Could you use it here?

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -361,23 +256,86 @@ namespace ROS2
...
+    void ROS2FrameComponent::DisablePublishingTransform()
+    {
+        m_disabled = true;
+        m_parentFrame.reset();
+        m_ros2Transform.reset();
+        m_sourceFrame.reset();
+        if (AZ::TickBus::Handler::BusIsConnected())
+        {
+            AZ::TickBus::Handler::BusDisconnect();
+        }
```
norbertprokopiuk, line 275: `BusDisconnect` already does check if bus is actually connected

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -361,23 +256,86 @@ namespace ROS2
...
+        auto* registry = AZ::SettingsRegistry::Get();
+        AZ_Error("ROS2FrameComponent", registry, "No settings registry found, using default odometry frame name");
+        if (registry)
+        {
+            if (!registry->Get(odometryFrame, DefaultGlobalFrameNameConfigurationKey))
+            {
+                odometryFrame = DefaultGlobalFrameName;
+            }
+        }
+        return GetNamespacedName(m_computedNamespace, odometryFrame);
```
norbertprokopiuk, line 338: should global frame have namespace? If it can have namespace then we can have multiple of them -> frame is not longer global. In my understanding global frame can be only one, like `world`

Gems/ROS2/Code/Source/Frame/ROS2FrameConfiguration.cpp
```diff
@@ -43,21 +43,11 @@ namespace ROS2
...
                         "Publish Transform",
                         "Publish Transform")
-                    ->DataElement(AZ::Edit::UIHandlers::Default, &ROS2FrameConfiguration::m_forceDynamic, "Force Dynamic", "Force Dynamic")
-                    ->ClassElement(AZ::Edit::ClassElements::Group, "Info")
-                    ->Attribute(AZ::Edit::Attributes::AutoExpand, true)
-                    ->UIElement(AZ::Edit::UIHandlers::Label, "Effective namespace", "")
-                    ->Attribute(AZ::Edit::Attributes::ValueText, &ROS2FrameConfiguration::m_effectiveNamespace)
-                    ->UIElement(AZ::Edit::UIHandlers::Label, "Full name", "")
-                    ->Attribute(AZ::Edit::Attributes::ValueText, &ROS2FrameConfiguration::m_fullName);
+                    ->DataElement(AZ::Edit::UIHandlers::Default, &ROS2FrameConfiguration::m_forceDynamic, "Force Dynamic", "Force Dynamic");
```
norbertprokopiuk, line 46: formatting

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
```diff
@@ -31,21 +35,26 @@ namespace ROS2
 
     void ROS2FrameEditorComponent::Init()
     {
-        m_configuration.m_namespaceConfiguration.Init();
     }
 
     void ROS2FrameEditorComponent::Activate()
     {
+        AZ_Printf("BuildGameEntity", "ROS2FrameEditorComponent::Activate for entity %s", this->GetEntityId().ToString().c_str());
```
norbertprokopiuk, line 42: maybe  AZ_Trace to reduce number of logs?
```suggestion
        AZ_Trace("BuildGameEntity", "ROS2FrameEditorComponent::Activate for entity %s", this->GetEntityId().ToString().c_str());
```

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
```diff
@@ -56,49 +65,55 @@ namespace ROS2
...
+            if (!registry->Get(odometryFrame, DefaultGlobalFrameNameConfigurationKey))
+            {
+                odometryFrame = DefaultGlobalFrameName;
+            }
+        }
+
+        const auto name_space = ComputeNamespace(m_configuration, GetEntityId());
 
-        return namespacedFrameName;
+        return GetNamespacedName(name_space, odometryFrame);
```
norbertprokopiuk, line 82: same with namespaces as for game component

Gems/ROS2/Code/Source/Frame/ROS2FrameGameSystemComponent.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    const AZStd::unordered_set<AZ::EntityId>& ROS2FrameGameSystemComponent::GetRegisteredFrames() const
+    {
+        return m_registeredEntities;
+    }
+
+    bool ROS2FrameGameSystemComponent::IsFrameRegistered(const AZ::EntityId& frameEntityId) const
+    {
+        return m_registeredEntities.find(frameEntityId) != m_registeredEntities.end();
```
norbertprokopiuk, line 89:
```suggestion
        return m_registeredEntities.contains(frameEntityId);
```

Gems/ROS2/Code/Source/Frame/ROS2FrameGameSystemComponent.cpp
```diff
@@ -0,0 +1,123 @@
...
+        AZStd::unordered_set<AZStd::string> frameIds;
+        AZStd::ranges::transform(
+            m_entityIdToFrameId | AZStd::views::values,
+            AZStd::inserter(frameIds, frameIds.end()),
+            [](const auto& frameId)
+            {
+                return frameId;
+            });
+
+        return frameIds;
```
norbertprokopiuk, line 120: Isn't it the same, but shorter?
```suggestion
        const auto valueView = AZStd::ranges::views::values(m_entityIdToFrameId);
        AZStd::unordered_set<AZStd::string> frameIds{ valueView.begin(), valueView.end() };

        return frameIds;
```

Gems/ROS2/Code/Tests/Frame/ROS2FrameComponentTest.cpp
```diff
@@ -174,8 +206,242 @@ namespace UnitTest
...
+
+    TEST_F(ROS2FrameComponentFixture, FrameTrackingInterfaceMultipleFrames)
+    {
+        auto trackingInterface = ROS2::ROS2FrameTrackingInterface::Get();
+        ASSERT_NE(trackingInterface, nullptr);
+
+        // Create multiple frame entities
+        constexpr int numFrames = 3;
+        std::vector<AZStd::unique_ptr<AZ::Entity>> entities;
+        std::vector<AZStd::string> expectedFrameIds;
```
norbertprokopiuk, line 318: Is there a reason to use std over AZStd?

norbertprokopiuk (changes requested):
I went through the code and I ran automatic tests, which passed. I will do manual tests shortly

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
arturkamieniecki: I disagree. Frame should be ready to handle EBus calls before registration. It should be responsive before any other system knows that is responsive (ie registered). This way also EBus calls can be used during registration.

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
arturkamieniecki: Anwsered in: https://github.com/o3de/o3de-extras/pull/968#discussion_r2316080341.

Gems/ROS2/Code/Source/Frame/NamespaceComputation.cpp
```diff
@@ -0,0 +1,220 @@
...
+        auto* transformInterface = entity->GetTransform();
+        if (!transformInterface)
+        {
+            return;
+        }
+        AZ::EntityId parentId = transformInterface->GetParentId();
+        if (!parentId.IsValid())
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
michalpelka: Added assert before recursion end.

Gems/ROS2/Code/Source/Frame/NamespaceComputation.cpp
```diff
@@ -0,0 +1,220 @@
...
+        {
+            if (HasROS2FrameComponent(*it))
+            {
+                return *it;
+            }
+        }
+        return AZ::EntityId();
+    }
+
+    AZStd::string GetNamespacedName(const AZStd::string& namespaceName, const AZStd::string& name)
```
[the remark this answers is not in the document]
michalpelka: No, I am using it outside, in ROS2FrameComponent.cpp

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
norbertprokopiuk: namespaced "global" frames like `odom` are completely valid vase and they should be kept. They might sit in the same origin, but if you decide to simulate SLAM or rather localisation on created map, your odom should start in (0,0,0) and later on it should be moved by localisation system. Namespaced "global" frames allow to prevent name collision if multiple robot uses the same map.

In my understanding `odom` is rather local frame, tightly related to robot. On the other hand frames like `map` or `world` are always static and they shouldn't be moved after creation. I would consider those two as actual global frames. 

You might find this document useful: https://www.ros.org/reps/rep-0105.html. It describes frame convention in ROS 2. The most important are these images: (world/earth could be treated as the same frame)
image

michalpelka:
Created a backup branch  RobotecAI:mp/rework_frame-bkp.

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
michalpelka: No, since that is a compact version that does not get full hierarchy. It is specialized function for this case.

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
michalpelka: agree with @arturkamieniecki .

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
```diff
@@ -54,51 +62,55 @@ namespace ROS2
...
     }
 
     bool ROS2FrameEditorComponent::IsTopLevel() const
     {
-        return ROS2FrameSystemInterface::Get()->IsTopLevel(GetEntityId());
+        AZ_Warning("ROS2FrameComponent", false, "Not implemented yet");
```
norbertprokopiuk, line 86:
```suggestion
        AZ_Warning("ROS2FrameEditorComponent", false, "Not implemented yet");
```

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
```diff
@@ -54,51 +62,55 @@ namespace ROS2
...
     }
 
     AZStd::string ROS2FrameEditorComponent::GetNamespacedFrameID() const
     {
-        AZStd::string namespacedFrameID;
-        ROS2NamesRequestBus::BroadcastResult(
-            namespacedFrameID, &ROS2NamesRequests::GetNamespacedName, GetNamespace(), m_configuration.m_frameName);
-        return namespacedFrameID;
+        auto name_space = ComputeNamespace(m_configuration, GetEntityId());
+        return GetNamespacedName(name_space, m_configuration.m_frameName);
```
norbertprokopiuk, line 93:
```suggestion
        auto namespace = ComputeNamespace(m_configuration, GetEntityId());
        return GetNamespacedName(namespace, m_configuration.m_frameName);
```

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
```diff
@@ -54,51 +62,55 @@ namespace ROS2
...
 
-    AZ::Name ROS2FrameEditorComponent::GetNamespacedJointName() const
+    AZStd::string ROS2FrameEditorComponent::GetNamespacedJointName() const
     {
-        AZStd::string namespacedJointName;
-        ROS2NamesRequestBus::BroadcastResult(
-            namespacedJointName, &ROS2NamesRequests::GetNamespacedName, GetNamespace(), m_configuration.m_jointName);
-        return AZ::Name(namespacedJointName.c_str());
+        auto name_space = ComputeNamespace(m_configuration, GetEntityId());
+        return GetNamespacedName(name_space, m_configuration.m_jointName);
```
norbertprokopiuk, line 113:
```suggestion
        auto namespace = ComputeNamespace(m_configuration, GetEntityId());
        return GetNamespacedName(namespace, m_configuration.m_jointName);
```

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
```diff
@@ -361,23 +262,97 @@ namespace ROS2
...
-        , m_isDynamic(configuration.m_isDynamic)
-        , m_forceDynamic(configuration.m_forceDynamic){};
+    ROS2FrameComponent::ROS2FrameComponent(const ROS2FrameConfiguration& ros2FrameConfiguration)
+        : m_configuration(ros2FrameConfiguration)
+    {
+    }
+
+    //! Disable publishing the transform. It allows to tinker with the transform without spamming /tf_static
+    //! you can disable it inactive state, and then enable it again.
+    void ROS2FrameComponent::DisablePublishingTransform()
```
norbertprokopiuk, line 272: Thinking out loud really, but it can really mess up TF tree. If we disable publishing of entity which has children with ros 2 frame, it will break TF tree since  parent frame of children stopped existing. Shouldn't we rather disable everything below this entity, starting from the end? Or maybe you just want users to take care of correct order of disabling frames?

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -803,12 +803,17 @@ namespace SimulationInterfaces
...
-                if (ros2Frame)
+                ROS2::ROS2FrameComponent* frameComponent = entity->template FindComponent<ROS2::ROS2FrameComponent>();
+                if (frameComponent)
                 {
-                    ros2Frame->UpdateNamespaceConfiguration(entityNamespace, ROS2::NamespaceConfiguration::NamespaceStrategy::Custom);
+                    const AZStd::string f = frameComponent->GetNamespacedFrameID();
+                    auto config = frameComponent->GetConfiguration();
+                    config.m_namespaceConfiguration.m_customNamespace = entityNamespace;
+                    config.m_namespaceConfiguration.m_namespaceStrategy = ROS2::NamespaceConfiguration::NamespaceStrategy::Custom;
+                    AZ_Printf("ROS2::ROS2FrameComponent", "Setting namespace to %s for entity %s\n", entityNamespace.c_str(), entity->GetName().c_str());
```
norbertprokopiuk, line 815:
```suggestion
                    AZ_Printf("SimulationInterfaces::SpawnEntity", "Setting namespace to %s for entity %s\n", entityNamespace.c_str(), entity->GetName().c_str());
```

norbertprokopiuk (commented):
I only went through the code, marked some places. Feel free to skip comment if you don't agree. I will do manual tests tomorrow morning and will add more comments if I find something

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
michalpelka: Yes, but what you are touching here is advanced use case. And with great power comes responsibility.
Of course you can split TFtree into two ones, but this damage is not permanent.
The previous implementation was trying to be clever. Here I propose dumb one, but flexible. However you can shoot you in the foot.

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
michalpelka: namespace is c++ keyword, it won't complile.

Gems/ROS2/Code/Include/ROS2/Frame/ROS2FrameTrackingInterface.h
```diff
@@ -0,0 +1,57 @@
...
+    //! Interface for querying and tracking ROS2 frame components in the system.
+    //! This interface provides access to the frame registry maintained by the ROS2FrameSystemComponent.
+    class ROS2FrameTrackingRequests
+    {
+    public:
+        AZ_RTTI(ROS2FrameTrackingRequests, ROS2FrameTrackingInterfaceTypeId);
+
+        //! Get all currently registered frame entities.
+        //! @return Set of EntityIds representing all registered frames
+        virtual const AZStd::unordered_set<AZ::EntityId>& GetRegisteredFrames() const = 0;
```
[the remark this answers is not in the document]
michalpelka: +1, @arturkamieniecki please fix.

Gems/ROS2/Code/Source/Frame/NamespaceComputation.cpp
```diff
@@ -0,0 +1,235 @@
...
+            }
+            const auto* componentGame = entity->FindComponent(AZ::Uuid(ROS2FrameComponentTypeId));
+            if (componentGame)
+            {
+                const auto* interface = dynamic_cast<const ROSFrameInterface*>(componentGame);
+                if (interface)
+                {
+                    return interface->GetConfiguration();
+                }
+            }
```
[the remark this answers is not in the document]
norbertprokopiuk: Lack of entity pointer is a potential bug, same goes with lack of interface from the component

Gems/ROS2/Code/Source/Frame/ROS2FrameEditorComponent.cpp
norbertprokopiuk: Yeah of course, my bad with suggestion. What I meant was to make it less python-style but we can leave it as it is

Gems/ROS2/Code/Source/Frame/NamespaceComputation.cpp
```diff
@@ -0,0 +1,249 @@
...
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+
+#include "NamespaceComputation.h"
+
+#include <AzCore/Component/ComponentApplicationBus.h>
+#include <AzCore/Component/Entity.h>
+#include <AzCore/Component/TransformBus.h>
+#include <ROS2/Frame/ROS2FrameComponent.h>
```
norbertprokopiuk, line 14: You missed this header and it doesn't compile in non-unity
```suggestion
#include <AzCore/Component/TransformBus.h>
#include <AzCore/Settings/SettingsRegistry.h>
#include <ROS2/Frame/ROS2FrameComponent.h>
```

norbertprokopiuk:
IMO after these changes it makes sense to forbid adding ROS 2 Frame component to `Level` entity. Adding frame to level entity makes whole tree wonky. It was added to kind of solve issues of the old code. We don't need to do it anymore. 
Tree with entity in level entity looks like this:
image
`world` was added in Level entity and `test` was set as global frame in registry.

tree without Frame Component in level entity looks better in my opinion:
image

Gems/ROS2/Code/Tests/Frame/ROS2FrameComponentTest.cpp
```diff
@@ -7,6 +7,9 @@
  *
  */
 
+#include "../../../../../../../engine/o3de/Code/Framework/AzCore/AzCore/std/typetraits/has_virtual_destructor.h"
```
norbertprokopiuk, line 10: :fire: :fire: :fire:

Gems/ROS2/Code/Source/Frame/ROS2FrameComponent.cpp
michalpelka: Added API that circle navigate:
```
        void DisableEntitySafely(const AZ::EntityId& entityToDisable) override;
        void EnableEntitySafely(const AZ::EntityId& entityToEnable) override;
```

Gems/ROS2/Code/Tests/Frame/ROS2FrameComponentTest.cpp
michalpelka: Stupid Clion. Piece of useless crap.

norbertprokopiuk:
Unfortunately `JointPositionEditorComponent` still fails 
image

norbertprokopiuk:
@michalpelka After your rework recursion was unnecessary in `JointPositionsEditorComponent`, that's why it brake.
Changing it to something like this fixes issue:
```cpp

    AZ::Crc32 JointsPositionsEditorComponent::FindAllJoints()
    {
        AzToolsFramework::ScopedUndoBatch undo("FindAllJoints");
        m_jointNames.clear();
        AZStd::function<void(const AZ::Entity* entity)> getAllJointsHierarchy = [&](const AZ::Entity* entity)
        {
            AZStd::set<AZ::EntityId> childrenEntityIds;
            ROS2::ROS2FrameEditorComponentBus::EventResult(
                childrenEntityIds, entity->GetId(), &ROS2::ROS2FrameEditorComponentBus::Events::GetFrameDescendants);
            // insert entity itself
            childrenEntityIds.insert(entity->GetId());
            for (const auto& entityId : childrenEntityIds)
            {
                AZ::Entity* childEntity = nullptr;
                AZ::ComponentApplicationBus::BroadcastResult(childEntity, &AZ::ComponentApplicationBus::Events::FindEntity, entityId);
                AZ_Assert(childEntity, "Entity not found!");
                AZ::Name jointName;
                ROS2::ROS2FrameComponentBus::EventResult(jointName, entityId, &ROS2::ROS2FrameComponentBus::Events::GetNamespacedJointName);

                const AZStd::string jointNameStr = jointName.GetCStr();
                const bool hasNonFixedJoints = JointUtils::HasNonFixedJoints(entity);
                if (!jointNameStr.empty() && hasNonFixedJoints)
                {
                    m_jointNames.emplace_back(jointNameStr);
                }
            }
        };
        getAllJointsHierarchy(GetEntity());
        undo.MarkEntityDirty(GetEntity()->GetId());
        return AZ::Edit::PropertyRefreshLevels::EntireTree;
    }
```

Gems/ROS2/Code/Tests/Frame/ROS2FrameComponentTest.cpp
```diff
@@ -348,17 +353,17 @@ namespace UnitTest
 
         // Check namespaced frame ID lookup
         auto namespacedFrameId = trackingInterface->GetNamespacedFrameId(entity.GetId());
-        ASSERT_TRUE(namespacedFrameId.has_value());
-        EXPECT_STREQ(namespacedFrameId.value().c_str(), "tracked_entity/tracked_frame");
+        ASSERT_TRUE(namespacedFrameId.size() > 0);
```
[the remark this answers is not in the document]
michalpelka: Could you refactor that?

Gems/ROS2/Code/Tests/Frame/ROS2FrameComponentTest.cpp
```diff
@@ -309,12 +312,14 @@ namespace UnitTest
         ROS2::ROS2FrameComponentBus::EventResult(
             globalFrameName_result, entity.GetId(), &ROS2::ROS2FrameComponentRequests::GetGlobalFrameID);
 
+        const auto globalNamespace = ROS2::GetGlobalFrameIDFromRegistry();
+
```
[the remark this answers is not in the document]
norbertprokopiuk: I don't want to force you to remove settings registry at all, but you need to apply changes that will cover all possibilities of the settings registry

arturkamieniecki:
I've added a simple change, requested in https://github.com/o3de/o3de-extras/pull/968#discussion_r2318827840

I've performed tests
- I've created a simple level with a few frames, moved them around and verified the effective namespace displayed in the editor.
- After each structure rearrangement, I've verified the TF2 tree.
- I've imported a robot from URDF, verified its generated TF2 tree.
- I've moved this robot using moveit.
- I've verified the TF2 data in rviz2, for the moving robot

PR #974, opened 2025-09-02
[Simulation Interfaces] Added presinsertion callback to spawnEntity

norbertprokopiuk:
## What does this PR do?
This PR adds possibility to pass preinsertion callback to spawnEntity bus
## How was this PR tested?
non unity build,
run existing test


jhanca-robotecai (approved):
Code looks, builds and runs as expected. Added some debug printing to the callback and it was printed correctly both in tests and GameLauncher.

PR #979, opened 2025-09-12
[Simulation Interfaces] Backport to work with O3DE 2505.1

norbertprokopiuk:
## What does this PR do?
This PR backports SImulationInterfaces Gem from development (future 2510 release) to work with O3DE 2505.1. It backports also required changes in ROS2 Gem without making any breaking changes. Gem which will be shipped to 2510 release requires changes made in engine which targets 2510 release as well. To make Gem compatible with engine 2505, Simulation Interfaces has added dependency to Physx Gem which is not needed in 2510 release

## How was this PR tested?
ran tests for SimulationInterfaces and ROS2 Gems, unity/non-unity build, monolithic project export.
Developer tests with ros2 service with rqt/terminal and scripts


Gems/ROS2/Code/Include/ROS2/ROS2TypeIds.h
jhanca-robotecai, line 1: Do we need this file?
Maybe setting `TFInterfaceTypeId` within `TransformInterface.h` is enough.

norbertprokopiuk:
@jhanca-robotecai 
Difference in files `Code/Source/Services/GetEntityInfoServiceHandler.cpp`, `Code/Source/Services/SetEntityInfoServiceHandler.cpp` should be fixed on stabilization. Those includes are redundant. Other include was just missed clang format

Other files you mentioned were removed during the development on `development` branch but I forgot to remove them. They existed in previous release.

 For linking ROS 2 package, I backported method you pointed out.
 
I backported `PublishTransform` method from TF interface as well

jhanca-robotecai (approved):
Thank you for your fixes and comments. I do not see any problems in the code anymore and my set of tests worked flawlessly (changing levels, spawning, despawning, moving entities around, stepping, pausing/stopping, etc.).

This backport will have a great use for the community, great work!

Issue #986, opened 2025-09-24
[ROS 2] [Code quality] Incoplete split


michalpelka:
**Describe the bug**

ROS 2 gem has assets that are used by ROS 2 sensor (custom graphic pipelines).
<img width="832" height="376" alt="Image" src="https://github.com/user-attachments/assets/1b337e01-58c9-4c13-b606-1cefac6925c5" />

**Assets required**
None - it is a problem with code structure

**Steps to reproduce**
None - it is a problem with code structure

**Expected behavior**
It works, for now.

**Actual behavior**
Those code should be moved next to gem that uses it (ROS 2 Sensors)

**Screenshots/Video**

**Found in Branch**
All

PR #991, opened 2025-10-03
Move camera passes ROS2->ROS2Sensors

jhanca-robotecai:
## What does this PR do?

Passes used in `ROS2CameraSensorComponent` were not moved from `ROS2` Gem to `ROS2Sensors` Gem when the code was moved. It worked correctly, as `ROS2Sensors` depends on `ROS2`. However, it makes no sense to keep the assets in the wrong place.

Closes https://github.com/o3de/o3de-extras/issues/986

## How was this PR tested?

The empty project was built and the camera sensor was tested.


michalpelka:
Works as expected.

PR #1014, opened 2025-12-31
Pass ROS 2 node args with registry

jhanca-robotecai:
## What does this PR do?

With this change user can pass ROS 2 node arguments via O3DE Registry. It might be very useful for remapping and other applications.

Note: I checked the option to directly use command line arguments, but it would require changes in the engine.

The PR to the documentation will follow: https://github.com/o3de/o3de.org/pull/2760

## How was this PR tested?

An option to remap topic `/clock` was used and the topic was remapped correctly.
`./build/2604/bin/profile/MyProject.GameLauncher --regset="/O3DE/ROS2/NodeArguments=--remap /clock:=/klocek" `


michalpelka (approved):

michalpelka:
Love this change!

PR #1020, opened 2026-02-04
Added missing state restoring if request failed

norbertprokopiuk:
## What does this PR do?

This PR fixes simulation interfaces LevelManager to correctly handle loading levels via ImGui. Original solution wasn't restoring internal state in case of request fail.

## How was this PR tested?

run GameLauncher with no level and loaded level via imgui. Simulation state changed as expected

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Clients/LevelManager.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Clients/LevelManager.cpp
@@ -123,6 +123,7 @@ namespace SimulationInterfaces
         {
             constexpr const char* errorMsg = "Additional Sources are not implemented yet";
             AZ_Warning("SimulationInterfaces", false, errorMsg);
+            m_actionRequestedFromSimInterfaces = false;
             return AZ::Failure(FailedResult(simulation_interfaces::msg::Result::RESULT_FEATURE_UNSUPPORTED, errorMsg));
         }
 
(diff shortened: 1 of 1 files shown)


patrykantosz (approved):

Gems/SimulationInterfaces/Code/Source/Clients/LevelManager.cpp
```diff
@@ -123,26 +123,30 @@ namespace SimulationInterfaces
         {
             constexpr const char* errorMsg = "Additional Sources are not implemented yet";
             AZ_Warning("SimulationInterfaces", false, errorMsg);
+            m_actionRequestedFromSimInterfaces = false;
```
jhanca-robotecai: Please correct me if I am wrong, but my understanding is that the flag is only used when loading levels to ensure we do send the notifications. If so, is the flag needed in `GetAvailableWorlds` and `GetCurrentWorld` methods? The flag might also be set to `true` before the call (after parsing the request)

jhanca-robotecai (commented):
I pushed the changes to your branch to answer my comment below. Please double-check if correct.

Gems/SimulationInterfaces/Code/Source/Clients/LevelManager.cpp
norbertprokopiuk: generally flag is used to distinguish whether action started via ros2 interface or some other O3DE system eg. ImGui.

jhanca-robotecai: But this information is only used for loading levels notifications, isn't it?

norbertprokopiuk: Yes, it is used only to handle cases when level is loaded via imgui or console in GameLauncher. It controls whether to respond to notification or not. So yes, changes you made look fine

jhanca-robotecai (approved):

PR #1022, opened 2026-02-27
Add optional to setpoint articulation

michalpelka:
## What does this PR do?

This PR changes behavior of Manpulation controller - it sets joint only if needed.

## How was this PR tested?

No test that can be reproduced


jhanca-robotecai:
The code is pushed to the backports branch only via releases.
I will close this PR, as it duplicates #1023. The code will be cherry-picked for the next O3DE release if necessary.

PR #1028, opened 2026-03-11
Added reverse lookup for simulated entities

norbertprokopiuk:
## What does this PR do?

PR adds reverse lookup to simulation interfaces to retrieve simulated entity name based on root entity Id.

## How was this PR tested?

built project, tests and run test. Tests fails but they fail the same way on development

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
@@ -1053,4 +1053,20 @@ namespace SimulationInterfaces
         return AZ::Success(m_simulatedEntityToPrefabRoot.at(name));
     }
 
+    AZ::Outcome<AZStd::string, FailedResult> SimulationEntitiesManager::GetSimulatedBodyNameById(const AZ::EntityId& entityId)
+    {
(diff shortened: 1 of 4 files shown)


michalpelka (approved):

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -1053,4 +1053,20 @@ namespace SimulationInterfaces
...
+    AZ::Outcome<AZStd::string, FailedResult> SimulationEntitiesManager::GetSimulatedBodyNameById(const AZ::EntityId& entityId)
+    {
+        const auto it = m_entityIdToSimulatedEntityMap.find(entityId);
+
+        if (it == m_entityIdToSimulatedEntityMap.end())
+        {
+            return AZ::Failure(SimulationInterfaces::FailedResult(
+                simulation_interfaces::msg::Result::RESULT_OPERATION_FAILED,
+                AZStd::string::format(
+                    "Entity with given ID \"%s\" doesn't exists in available cache of simulated entities names",
```
pijaro:
```suggestion
                    "Entity with given ID \"%s\" doesn't exist in the available cache of simulated entities names",
```

PR #1029, opened 2026-03-11
Remove get entities warning

norbertprokopiuk:
## What does this PR do?

PR removes warning from GetEntities. User should not be forced to select specific shape in c++ API. Compatibility with standard is checked in ROS 2 layer anyway.

## How was this PR tested?

built project and run test which fail even on development

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
@@ -392,14 +392,6 @@ namespace SimulationInterfaces
                     entities.push_back(name);
                 }
             }
-            else if (auto capsuleShape = dynamic_cast<Physics::CapsuleShapeConfiguration*>(shape.get()))
-            {
-                const auto capsule = capsuleShape->ToCapsule(shapePose);
-                if (capsule.Contains(worldTranslation))
-                {
-                    entities.push_back(name);
-                }
-            }
             else
             {
                 AZ_WarningOnce("SimulationInterfaces", false, "");
--- a/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
@@ -772,7 +772,7 @@ namespace SimulationInterfaces
             AZ_Warning("SimulationInterfaces", false, "Initial pose is not orthogonal");
             completedCb(
                 AZ::Failure(FailedResult(
-                simulation_interfaces::srv::SpawnEntity::Response::INVALID_POSE, "Initial pose is not orthogonal"))); //  INVALID_POSE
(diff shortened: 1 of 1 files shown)


zakmat (commented):
Consider using Az_WarningOnce in the else clause, but the message should be adjusted. We are able to use convex hull to filter physical entities - only the non-physical are unsupported

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -361,6 +362,8 @@ namespace SimulationInterfaces
                 }
             }
 
+            // allow check for all supported shaped. Simulation interfaces standard constraints are checked in ROS 2 layer of the node
```
pijaro:
```suggestion
            // Allow check for all supported shapes. Simulation interfaces standard constraints are checked in ROS 2 layer of the node.
```

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -381,9 +384,25 @@ namespace SimulationInterfaces
...
+                }
+            }
+            else if (auto capsuleShape = dynamic_cast<Physics::CapsuleShapeConfiguration*>(shape.get()))
+            {
+                const auto capsule = capsuleShape->ToCapsule(shapePose);
+                if (capsule.Contains(worldTranslation))
+                {
+                    entities.push_back(name);
+                }
+            }
```
pijaro: Duplicated else if? These are the same if's.

norbertprokopiuk:
```suggestion
```

norbertprokopiuk: yep, copy paste error. Removed

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -381,9 +384,17 @@ namespace SimulationInterfaces
...
             else
             {
-                AZ_Warning("SimulationInterfaces", false, "Unsupported bounds type, skipped");
+                AZ_WarningOnce("SimulationInterfaces", false, "");
```
jhanca-robotecai: Please fix the warning message.

jhanca-robotecai (changes requested):

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
norbertprokopiuk:
```suggestion
                AZ_WarningOnce("SimulationInterfaces", false, "Non-physical entities support only primitive shapes for bounds check");
```

norbertprokopiuk: done

jhanca-robotecai (approved):

PR #1031, opened 2026-03-17
Added notification for simulation state change

norbertprokopiuk:
## What does this PR do?

PR adds notification triggered on simulation state change.
It also changes all `SimulationManagerNotifications` from pure virtual to virtual with empty implementation to not force users  to define both callbacks if they don't need it

## How was this PR tested?

Built in non-unity, checked if notification is triggered when state change. I user keyboard shortcuts for this

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Clients/SimulationManager.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Clients/SimulationManager.cpp
@@ -519,6 +519,7 @@ namespace SimulationInterfaces
             }
         }
         m_simulationState = stateToSet;
+        SimulationManagerNotificationsBus::Broadcast(&SimulationManagerNotifications::OnSimulationStateChange, m_simulationState);
         return AZ::Success();
     }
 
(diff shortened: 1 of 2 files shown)


michalpelka (approved):

Gems/SimulationInterfaces/Code/Source/Clients/SimulationManager.cpp
```diff
@@ -519,6 +519,7 @@ namespace SimulationInterfaces
             }
         }
         m_simulationState = stateToSet;
+        SimulationManagerNotificationsBus::Broadcast(&SimulationManagerNotifications::OnSimulationStateChange, m_simulationState);
```
jhanca-robotecai, line 522: Shouldn't you add a similar empty call to `OnSimulationStepFinish` in https://github.com/o3de/o3de-extras/blob/f82e5cdc70541219c839c102dfe20ad2a0461fdb/Gems/SimulationInterfaces/Code/Source/Clients/SimulationManager.cpp#L366 ?

jhanca-robotecai: Found it in different place, sorry for confusion. It is all okay.

jhanca-robotecai (approved):

PR #1032, opened 2026-03-17
Add batch spawn entities

styczen:
## What does this PR do?

Add feature to batch spawn entities using `SimulationEntityManagerRequests` EBus.

## How was this PR tested?

- built in non-unity

### DIFF
--- a/Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/SimulationEntityManagerRequestBus.h
+++ b/Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/SimulationEntityManagerRequestBus.h
@@ -70,8 +70,7 @@ namespace SimulationInterfaces
     using SpawnableList = AZStd::vector<Spawnable>;
     using DeletionCompletedCb = AZStd::function<void(const AZ::Outcome<void, FailedResult>&)>;
     using SpawnCompletedCb = AZStd::function<void(const AZ::Outcome<AZStd::string, FailedResult>&)>;
-    using PreInsertionCb =
-        AZStd::function<void(const AZ::Outcome<AzFramework::SpawnableEntityContainerView, FailedResult>&)>;
+    using PreInsertionCb = AZStd::function<void(const AZ::Outcome<AzFramework::SpawnableEntityContainerView, FailedResult>&)>;
 
     class SimulationEntityManagerRequests
     {
@@ -121,6 +120,16 @@ namespace SimulationInterfaces
             const bool allowRename,
             PreInsertionCb preinsertionCb,
             SpawnCompletedCb completedCb) = 0;
+
+        //! Callback for when an entities have been spawned and registered.
+        virtual void SpawnEntities(
+            const AZStd::vector<AZStd::string>& names,
+            const AZStd::vector<AZStd::string>& uris,
+            const AZStd::vector<AZStd::string>& entityNamespaces,
+            const AZStd::vector<AZ::Transform>& initialPoses,
+            const AZStd::vector<bool>& allowRename,
+            AZStd::vector<PreInsertionCb>& preinsertionCb,
+            AZStd::vector<SpawnCompletedCb>& completedCb) = 0;
 
         //! Reset the simulation to begin.
         //! This will revert the entire simulation to the initial state.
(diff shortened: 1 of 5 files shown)


Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -206,7 +202,8 @@ namespace SimulationInterfaces
                 simulation_interfaces::msg::SimulatorFeatures::ENTITY_BOUNDS,
                 simulation_interfaces::msg::SimulatorFeatures::DELETING,
                 simulation_interfaces::msg::SimulatorFeatures::SPAWNABLES,
-                simulation_interfaces::msg::SimulatorFeatures::SPAWNING });
+                simulation_interfaces::msg::SimulatorFeatures::SPAWNING,
+                SPAWNING_BATCH });
```
norbertprokopiuk: I would avoid adding this. This is used purely to define coverage of the standard within the Gem. Batch spawning is not yet part of the standard so we shouldn't add it

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -907,6 +891,76 @@ namespace SimulationInterfaces
...
+        }
+
+        if (spawningEntities.empty())
+        {
+            if (completedCb)
+            {
+                completedCb(BatchSpawnResult{});
+            }
+            return;
+        }
```
norbertprokopiuk, line 923: Some logging/warnings would be useful here

Gems/SimulationInterfaces/Code/Source/Clients/SimulationEntitiesManager.cpp
```diff
@@ -907,6 +891,76 @@ namespace SimulationInterfaces
...
+
+                if (batchContext->m_completedCount != batchContext->m_result.m_spawnResults.size())
+                {
+                    return;
+                }
+
+                if (batchContext->m_completedCb)
+                {
+                    batchContext->m_completedCb(batchContext->m_result);
+                }
```
norbertprokopiuk: I'm not convinced to this approach. You created `batchContext` as a shared pointer, which is passed as a copy to each callback. If you would extend `BatchSpawnContext` struct with descturctor which calls `completedCb` all this logic could be omitted. By passing context by copy you extend lifetime of the `batchContext` and only after last callback, when all references are gone, callback will be called.

Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/SimulationEntityManagerRequestBus.h
```diff
@@ -70,8 +72,38 @@ namespace SimulationInterfaces
...
+                [](const AZ::Outcome<AZStd::string, FailedResult>& result)
+                {
+                    return result.IsSuccess();
+                });
+        }
+    };
+
+    using BatchSpawnCompletedCb = AZStd::function<void(const BatchSpawnResult&)>;
+
+    static constexpr SimulationFeatureType SPAWNING_BATCH = 50;
```
norbertprokopiuk: Feature id is inherited from the standard. You shouldn't add new value for it. We need to wait for standard to be extended to add this to supported feautres

norbertprokopiuk (commented):
I just went through the code, didn't test it. I marked some issues that got my attention

Gems/SimulationInterfaces/Code/Tests/Tools/Mocks/SimulationEntityManagerMock.h
```diff
@@ -26,16 +26,17 @@ namespace UnitTest
...
                 const AZStd::string& name,
                 const AZStd::string& uri,
                 const AZStd::string& entityNamespace,
                 const AZ::Transform& initialPose,
-                const bool allowRename,
-                PreInsertionCb preinsertionCb,
-                SpawnCompletedCb completedCb));
+                 const bool allowRename,
+                 PreInsertionCb preinsertionCb,
+                 SpawnCompletedCb completedCb));
```
patrykantosz: Are that additional spaces intended?

norbertprokopiuk (approved):
Looks good to me. I tested it with custom component to call the bus and works as intended.

jhanca-robotecai (approved):

PR #1033, opened 2026-03-18
[Simulation  interfaces] Public path utilities

norbertprokopiuk:
## What does this PR do?

PR moves path/uri utilities to public API target

## How was this PR tested?

built project in non-unity

### DIFF
--- a/Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
+++ b/Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
@@ -7,8 +7,6 @@
  */
 #pragma once
 
-#include <AzCore/IO/Path/Path.h>
-#include <AzCore/Settings/SettingsRegistry.h>
 #include <AzCore/std/optional.h>
 #include <AzCore/std/string/string.h>
 
@@ -43,4 +41,4 @@ namespace SimulationInterfaces::PathUtilities
         return {};
     }
 
-} // namespace SimulationInterfaces::RegistryUtilities
+} // namespace SimulationInterfaces::PathUtilities
--- a/Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
+++ b/Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include <AzCore/IO/Path/Path.h>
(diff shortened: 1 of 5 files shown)


Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
```diff
@@ -0,0 +1,46 @@
...
+    {
+        if (uri.starts_with(ProductAssetPrefix))
+        {
+            const AZStd::string_view productAssetPrefix{ ProductAssetPrefix };
+            return uri.substr(productAssetPrefix.length());
+        }
+        return {};
+    }
+
+} // namespace SimulationInterfaces::RegistryUtilities
```
jhanca-robotecai:
```suggestion
} // namespace SimulationInterfaces::PathUtilities
```

Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright (c) Contributors to the Open 3D Engine Project.
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include <AzCore/IO/Path/Path.h>
+#include <AzCore/Settings/SettingsRegistry.h>
```
jhanca-robotecai:
```suggestion
```

Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
```diff
@@ -0,0 +1,44 @@
...
+    //! URI: "product_asset:///path/to/file.txt"
+    //! relative path: "path/to/file.txt"
+    inline AZStd::string UriToRelPath(AZStd::string_view uri)
+    {
+        if (uri.starts_with(ProductAssetPrefix))
+        {
+            const AZStd::string_view productAssetPrefix{ ProductAssetPrefix };
+            return uri.substr(productAssetPrefix.length());
+        }
+        return {};
```
pijaro, line 41: Shouldn't we return `uri` if the provided path is already a relative one?

```suggestion
        return uri;
```

norbertprokopiuk: I just copied the function and honestly didn't think much about implementation. But returning empty string kind of guards from passing through non-uri path. Non-uri meaning those which don't start with the product asset prefix or start with the other prefix. I would expect non-uri paths to fail this conversion

jhanca-robotecai (approved):

Gems/SimulationInterfaces/Code/Include/SimulationInterfaces/PathUtils.h
jhanca-robotecai: I agree with the argument that returning an empty string instead of returning the input is a better approach.

PR #1042, opened 2026-04-14
Change .h to .hpp for tf2_ros includes

jhanca-robotecai:
## What does this PR do?

Development branch of ROS 2 (Rolling) marked headers with `.h` extension as deprecated, triggering warnings that stop compilation (we treat warnings as errors). This would cause problems when building ROS2 Gem in O3DE 2605 with ROS 2 Lyrical Luth (to be released in May).

This PR changes the extensions.
Note, that the headers in question exist in ROS 2 Humble already, one example:
```
cat /opt/ros/humble/include/tf2_ros/tf2_ros/static_transform_broadcaster.h
/*
 * Copyright (c) 2008, Willow Garage, Inc.
 * All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions are met:
 *
 *     * Redistributions of source code must retain the above copyright
 *       notice, this list of conditions and the following disclaimer.
 *     * Redistributions in binary form must reproduce the above copyright
 *       notice, this list of conditions and the following disclaimer in the
 *       documentation and/or other materials provided with the distribution.
 *     * Neither the name of the Willow Garage, Inc. nor the names of its
 *       contributors may be used to endorse or promote products derived from
 *       this software without specific prior written permission.
 *
 * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
 * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
 * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
 * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
 * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
 * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
 * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
[...]

PR #1044, opened 2026-04-15
Support simulation_interfaces 2.1.0 package alongside 1.4.0/1.5.0/1.6.0

jhanca-robotecai:
## What does this PR do?

This PR adds a support for the latest [simulation_interfaces](https://github.com/ros-simulation/simulation_interfaces) 2.1.0 package alongside already supported 1.4.0/1.5.0/1.6.0. 2.1.0 will be a part of ROS 2 Lyrical Luth release in late May.

The implementation follows the pattern from `ROS2` Gem, i.e., the version of the API is checked at the configure level and respective implementation is selected using precompiler options.

Note: this PR includes changes that were pushed to the `stabilization/26050` branch:
- #1042
- #1043

## How was this PR tested?

I tested with ROS 2 Humble, Jazzy and Rolling using the following docker environment: https://github.com/jhanca-robotecai/TestROSRolling

The some basic CLI commands were used to check if services work as expected.

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Services/SetEntityStateServiceHandler.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Services/SetEntityStateServiceHandler.cpp
@@ -24,10 +24,13 @@ namespace ROS2SimulationInterfaces
     AZStd::optional<SetEntityStateServiceHandler::Response> SetEntityStateServiceHandler::HandleServiceRequest(
         const std::shared_ptr<rmw_request_id_t> header, const Request& request)
     {
+        const AZStd::string entityName = request.entity.c_str();
         const auto simulatorFrameId = RegistryUtilities::GetSimulatorROS2Frame();
         const AZStd::string_view messageFrameId{ request.state.header.frame_id.c_str(), request.state.header.frame_id.length() };
+
+        // Resolve frame transform offset once (needed for both pose and twist)
         AZ::Transform transformOffset = AZ::Transform::CreateIdentity();
-        if (!messageFrameId.empty() && simulatorFrameId != messageFrameId)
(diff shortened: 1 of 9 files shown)


Gems/SimulationInterfaces/Code/Source/Services/SetEntityStateServiceHandler.cpp
```diff
@@ -55,18 +95,13 @@ namespace ROS2SimulationInterfaces
             response.result.error_message = poseValidation.GetError().c_str();
             return response;
         }
```
norbertprokopiuk, line 97: I don't know How it happened but this error (`INVALID_POSE`) is only in humble and rolling so this #if is broken

norbertprokopiuk (commented):
I tested building but unfortunately right now this change is breaking for jazzy and kilted which don't have `INVALID_POSE` for SetEntityState. All changes from this pr are correct, it build correclt if I add missing error type.

Gems/SimulationInterfaces/Code/Source/Services/SetEntityStateServiceHandler.cpp
jhanca-robotecai: So the question is what to do in O3DE with broken ROS. Should we care? Hopefully ROS will be fixed one day. Until then people can build the package from source (i.e. use 2.1.0)

norbertprokopiuk: This looks like correct approach. since I didn't have other comments I approve this PR

norbertprokopiuk (approved):

pijaro (approved):

PR #1051, opened 2026-05-06
Fix test dependencies for SimulationInterfaces

zakmat:
## What does this PR do?

Splits the `SimulationInterfaces` test targets so the runtime tests stop linking `AzToolsFramework` (and its transitive Qt5 dependency). Those targets never exercised any editor or Qt APIs — they were pulling Qt5 in by inheritance.

The motivating symptom was a `_ZN9QXmlUtils6isCharE5QChar, version Qt_5_PRIVATE_API` crash at test load time on SDK builds inside a pixi/conda environment: the loader picked up conda-forge's vanilla Qt5 5.15.x from the env's `lib` directory before O3DE's patched 5.15.1, and `libQt5Xml.so.5`'s versioned symbol references against `libQt5Core.so.5` couldn't resolve. That's a loader-ordering problem in mixed-Qt environments and it can be worked around at the consumer level (e.g. prepending `-Wl,-rpath,$ORIGIN` to `LDFLAGS` at configure time, or `LD_PRELOAD`-ing the O3DE Qt5 libs at run time). The structural fix shipped here is to stop linking Qt5 from tests that don't need it; tests that genuinely use editor APIs are now isolated to a single target.

A separate issue I hit on SDK builds - `AzTestRunner` not walking up to find `project.json`, and `libROS2.so` not resolving `libAzCore.so` via RPATH - is out of scope here; the fix belongs in the engine (`AzTestRunner` / `GemTestEnvironment`) and was recently merged into stabilization for the upcoming release (https://github.com/o3de/o3de/pull/19739).

The PR makes the following changes:

**Target restructure.** The test suite now splits into three targets, only one of which still pulls in the editor link surface:
  - `SimulationInterfaces.Tests` — entity-lifecycle tests on the new `RuntimeTestApplication`.
  - `SimulationInterfaces.Editor.Tests` — actual editor tests, renamed from `TestApp` to match convention.
  - `SimulationInterfaces.ROS2Tests` — service-handler/mock tests, also moved onto the runtime application.

**Runtime test scaffolding.** New `RuntimeTestApplication` and `RuntimeTestFixture` mirror the existing `ToolsTestApplication` fixture but without the editor/Qt link surface.

**Collateral fixes** surfaced while running `Editor.Tests` end-to-end:
  - `Gems/ROS2/.../FollowingCameraComponent.cpp`: replaced a static `AZStd::unordered_map` with `fixed_unordered_map` so the SystemAllocator no longer reports leaks at test teardown.
  - `Tests/Tools/TestFixture.cpp`: `TickApp` now also calls `app->TickSystem()` so `AssetBus` notifications and `SystemTickBus::QueueFunction` lambdas drain — without this the spawnable preload never reaches `Ready` and `SpawnAllEntitiesCommand` requeues forever.
  - `Tests/Tools/SimulationInterfaceAppTest.cpp`: block on `AssetManager::BlockUntilLoadComplete` for the spawnable so the first `SpawnEntity` call doesn't race the asset I/O job.

## How was this PR tested?

1. Installed stabilization/2605 candidate SDK - since latest deb does not yet have the upstream fix ready I have applied it locally
2. Created a project from template, attached o3de-extras gems including `SimulationInterfaces`.
3. Ran non-editor tests: `ctest -R Tests -E Editor -C profile --output-on-failure`.
4. Ran editor tests: `ctest -R Editor.Tests -C profile --output-on-failure`.


Gems/SimulationInterfaces/Code/Tests/Common/RuntimeTestFixture.cpp
```diff
@@ -0,0 +1,182 @@
...
+        auto* transformComponent = entity->CreateComponent(AZ::TransformComponentTypeId);
+        AZ_Assert(transformComponent, "Failed to create TransformComponent");
+        auto* transformInterface = azrtti_cast<AZ::TransformInterface*>(transformComponent);
+        AZ_Assert(transformInterface, "Failed to get TransformInterface");
+        transformInterface->SetWorldTM(transform);
+        entity->CreateComponent(AZ::Uuid(PhysXRigidBodyComponentTypeId));
+        entity->CreateComponent(AZ::Uuid(PhysXShapeColliderComponentTypeId));
+        entity->CreateComponent(AZ::Uuid(SphereShapeComponentTypeId));
+        entity->Init();
+        entity->Activate();
```
norbertprokopiuk: technically this is against new way of entities activation introduced in o3de https://github.com/o3de/o3de/pull/19319. IT should be entity bit activation and then effective state evaluation. Something like that
```
entity->SetEntityActive(true);
entity->ApplyEffectiveActiveState();
```

Gems/SimulationInterfaces/Code/Tests/Common/RuntimeTestFixture.cpp
```diff
@@ -0,0 +1,182 @@
...
+    void SimulationInterfaceTestFixture::ClearEntities()
+    {
+        for (auto& entity : m_entities)
+        {
+            AZ::Outcome<void, SimulationInterfaces::FailedResult> output;
+            SimulationInterfaces::SimulationEntityManagerRequestBus::BroadcastResult(
+                output, &SimulationInterfaces::SimulationEntityManagerRequests::UnregisterSimulatedBody, entity.second->GetName());
+            AZ_Assert(
+                output.IsSuccess(), "Failed to unregister entity from simulation_interfaces, %s", output.GetError().m_errorString.c_str());
+            entity.second->Deactivate();
```
norbertprokopiuk: same as before, again new activation system

Gems/SimulationInterfaces/Code/Tests/Tools/SimulationInterfaceAppTest.cpp
```diff
@@ -47,6 +49,14 @@ namespace UnitTest
             AZ::Data::s_invalidAssetType,
             false);
         AZ_Assert(assetId.IsValid(), "Failed to get asset id for %s", TestSpawnable.c_str());
+
+        // Block until the spawnable is fully loaded so the first SpawnEntity call doesn't race the
+        // async asset I/O job. SpawnableEntitiesManager requeues SpawnAllEntitiesCommand until
+        // m_spawnable.IsReady(), so a not-yet-loaded asset would silently stall the test.
+        auto preloadedAsset = AZ::Data::AssetManager::Instance().GetAsset<AzFramework::Spawnable>(
+            assetId, AZ::Data::AssetLoadBehavior::PreLoad);
+        AZ::Data::AssetManager::Instance().BlockUntilLoadComplete(preloadedAsset);
```
norbertprokopiuk, line 112: `AZ::Data::AssetManager::Instance()` can be nullptr in case something is loaded in a wrong way. You could consider adding assert for valid pointer

Gems/SimulationInterfaces/Code/Tests/Common/SimulationInterfaceTestFixture.cpp
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    void SimulationInterfaceTestFixture::StepPhysics(int numSteps)
+    {
+        for (int i = 0; i < numSteps; i++)
+        {
+            auto* physicsSystem = AZ::Interface<AzPhysics::SystemInterface>::Get();
+            physicsSystem->Simulate(1.0f / 60.0f);
+        }
+    }
```
norbertprokopiuk, line 144: What was the motivation for getting this system constantly in the loop rather than getting it before and assert for being valid?

Gems/SimulationInterfaces/Code/Tests/Common/RuntimeTestFixture.cpp
```diff
@@ -0,0 +1,182 @@
...
+    {
+        auto findIt = m_entities.find(entityId);
+        if (findIt != m_entities.end())
+        {
+            AZ::Outcome<void, SimulationInterfaces::FailedResult> output;
+            SimulationInterfaces::SimulationEntityManagerRequestBus::BroadcastResult(
+                output, &SimulationInterfaces::SimulationEntityManagerRequests::UnregisterSimulatedBody, findIt->second->GetName());
+            AZ_Assert(
+                output.IsSuccess(), "Failed to unregister entity from simulation_interfaces, %s", output.GetError().m_errorString.c_str());
+            findIt->second->Deactivate();
```
norbertprokopiuk: new activation system

norbertprokopiuk (approved):
Generally works as expected and dependency clean up is neat. 

But since you touched these test we can make them compatible with o3de 2605.0. I marked some places to change entity activation to be compatible with new o3de release

Gems/SimulationInterfaces/Code/Tests/Common/SimulationInterfaceTestFixture.cpp
zakmat: removed from loop

Gems/SimulationInterfaces/Code/Tests/Tools/SimulationInterfaceAppTest.cpp
zakmat: I think Instance() is guarding itself, but I have added explicit IsReady assert

zakmat:
thanks for the review. I have rebased on top of current development and applied your suggestions.

Additionally I have shuffled a bit a fixture boiler code to reduce duplication

norbertprokopiuk (approved):
Looks good and all tests passes as expected

PR #1056, opened 2026-05-29
[SimulationIntefaces] Fix compilation issue with older ROS 2 simulation-interfaces packages

michalpelka:
## What does this PR do?

Problem solved here is to choose compile-time which version of response message to use. In version 1.5.0 there is no simulation_interfaces::msg::Result::INVALID_POSE, so we need to use both constexpr and SFINAE to give compiler a way out to compile.

## How was this PR tested?

- Build with version Simulation Interfaces 1.5.0 on Jazzy.
- Build with version Simulation Interfaces 1.5.1 on Jazzy.

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Services/SetEntityStateServiceHandler.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Services/SetEntityStateServiceHandler.cpp
@@ -13,6 +13,22 @@
 #include <SimulationInterfaces/RegistryUtils.h>
 #include <SimulationInterfaces/SimulationEntityManagerRequestBus.h>
 
+namespace detail
+{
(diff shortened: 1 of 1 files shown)


norbertprokopiuk:
@michalpelka I understand motivation, but lack of the error code was fixed in version 1.5.1: https://github.com/ros-simulation/simulation_interfaces/tree/jazzy. For time period when this PR is in review you might want to use simulation interfaces built from source (I believe ros 2 jazzy sync with apt repo is still pending)

norbertprokopiuk (approved):
builds on jazzy-simulation-interfaces 1.5.0 and 1.5.1

It's really cool that more c++20 features enter this code base

zakmat (approved):

PR #1055, opened 2026-05-28
Update multiplayer template to fix compile issue with floats.

nick-l-o3de:
## What does this PR do?

* Updates the autocomponent XML for NetworkSimplePlayerCameraComponent to have the appropriate 'f' prefix for float literals.
* Upates the template.json with a new version number.
* Updates the repo.json to have the new zip file and hash and version.

## How was this PR tested?

Local testing of compilation using this template.

PR #1057, opened 2026-05-31
Update multiplayer template to fix compile issue with floats. (#1055)

jhanca-robotecai:
## What does this PR do?

There was a decision to push a fix to Multiplayer template right after the release of 2605.0. The fix was  in the `main` branch, it should be also available in `development`.

## How was this PR tested?

The code was tested in #1055, I only cherry-picked and compared diffs.


kberg0 (approved):
Directly mirrors release fix. Approved.

PR #1066, opened 2026-07-27
Use O3DE's 3rdParty convention for ROS 2 package dependencies to survive in Prebuilt Gems

zakmat:
## What does this PR do?

This PR makes the ROS 2 package dependencies survive the gem export

ROS 2 dependency macros used to link resolved packaged directly with `target_link_libraries`. That works well for source builds, where the macros run again at every configure but brings some issue when we want Gem to be packaged for distribution.

In the latter case we would like to reuse O3DE existing mechanisms instead of handcrafting cmake files for all possible distribution variants. But using O3DE as is, it drops the deps that were not registered using its own functions.

To make it work I have replaced the `target_link_libraries` with `ly_` equivalent and reused `3rdParty::` system to make the deps discoverable (each dependency is declared under `3rdParty::ROS2::<pkg>` and can be resolved independently by producer and consumers machines. `3rdParty` is engine's magic that invokes `FindROS2.cmake` for unresolved names on consumer's side). I have crafted the code so the existing macro signatures are unchanged, therefore no edits in existing Gems are needed.

Additionally I moved the FindROS2.cmake to 3rdParty directory where it belongs by convention.

Also noticed that the above removed the need for custom registration - hence the third commit - but please advise against if you feel we should keep those lines. Some out-of-tree consumers might rely on it.

## How was this PR tested?

Linux, ROS 2 Jazzy.
Source builds of the ROS 2 Gem and some of its dependent Gems - unaffected.
Prebuilt Gems flow tested on projects with 25.05 source engine with back-ported extras and SDK 26.05 engine.


jhanca-robotecai (approved):
The code looks correct.

I tested the build with the following set-ups:
- engine (source, 2605) + simulation Gems + binary proprietary Gem using ROS 2 as dependency
- engine (SDK, 2605) + simulation Gems + binary proprietary Gem using ROS 2 as dependency

In both setups I built the projects and exported them with mono+release and nonmono+profile configs.

michalpelka (approved):

PR #1070, opened 2026-08-06
Refactor of the RobotImporter Gem files from the Utils directory.

Sebastian306:
## What does this PR do?
Refactor of the RobotImporter gem files from the Utils directory.

- Removal of unnecessary dependencies and imports from files, as well as unused functions.
- Leaving in the Utils directory only files that are independent of the rest of the code.
- Extracting and splitting the two largest files from Utils into smaller, thematically coherent files.
    - Utils/RobotImporterUtils:
        - Queries/SdfQueries - queries on sdf objects.
        - Queries/SdfPluginUtils - queries regarding plugins and their related parameters.
        - Queries/SdfVisitors - functions for traversing sdf objects in order to run queries on them.
        - Assets/AssetPathResolver - functions for finding the paths of assets referenced in a URDF/SDF file.
        - URDF/SdfParser (added to the existing file) - moving CreateSdfParserConfigFromSettings to the parsing
      functions in which the configuration created by the moved function is used.

    - SourceAssetsStorage
        - Assets/AssetTypes.h - a header file with definitions of the types used in the Assets/ files - frequently
          included instead of the whole SourceAssetsStorage.h file.
        - Assets/AssetLookup - looking up specific types of processed assets.
        - Assets/AssetImporter - moving asset files into folders handled by the AssetProcessor.
        - Assets/SceneManifestBuilder - creating side-car files for assets (.assetinfo).

- Change namespaces in Queries and Assets folders from Utils to SDFormat and Assets.
- Change Utils::TypeConversions namespace into Utils::SDFormat::TypeConversions.
## How was this PR tested?

Build and launch project with gem. Several robots were imported into the project.


jhanca-robotecai (approved):
All methods are correctly moved and namespaces are correctly applied. The dead code is correctly removed.

The only problem I found is that `SdfAssetBuilder.h` forward-declares `SdfAssetBuilderSettings` but still holds it by value. This only compiles because  it transitively gets the full definition via the `SdfParser.h` include. Slightly fragile. Please fix it in one of the follow-up PRs.

PR #1074, opened 2026-08-11
[Simulation Interfaces] Add explicit removal of simulation entities before level unload

norbertprokopiuk:
## What does this PR do?

I noticed that on O3DE 2605.0 simulation entities are not removed during level unload. I added explicit removal of those entities with keeping the `UnloadWorld` synchronous, as it was. 

## How was this PR tested?

I spawned object, restarted level via services are spawned object was gone. Before mine change it wasn't a case. 

I tested also other world related services and they work as expected


patrykantosz (approved):
Works as expected.

michalpelka (approved):

PR #1073, opened 2026-08-10
Renaming the URDF folder to a more appropriate name and moving the markers in RobotImporter Gem

Sebastian306:
## What does this PR do?

Rename the folder holding the prefab building code (URDF -> Building).
The old name is a leftover from the times when the gem did not parse SDF and used
a different library that handled URDF only.

Move the maker files into a separate Building/Makers subfolder.

PR requires prior merger of changes from PR [Collect all files related to parsing in RobotImporter Gem](https://github.com/o3de/o3de-extras/pull/1072)

## How was this PR tested?

The project was built using the new version of the gem, and a wizard was used to load several test models.

PR #1075, opened 2026-08-12
RobotImporter Gem - passing referenced assets and import status through EBuses.

Sebastian306:
## What does this PR do?

Adds two new EBuses: one for exposing information about assets referenced by the source
file, and one for collecting prefab build status messages. Both are handled by the new
`ImportSession` class, addressed by import session id - it owns the referenced asset map,
accumulates the messages and renders the final import report.

Previously the asset map had to be passed by hand through several layers down to the
makers, many of which only forwarded it. Status messages were collected separately in
each maker and pulled back up through `SdfPrefabMaker` class method. Makers now take only
an id of import session and talk over the buses. `SdfPrefabMaker` no longer owns status.

PR requires prior merger of changes from PR [Renaming the URDF folder to a more appropriate name and moving the markers in RobotImporter Gem](https://github.com/o3de/o3de-extras/pull/1073)

## How was this PR tested?

The project was built using the new version of the gem. Several robots were loaded using the wizard and a Python file.


Gems/ROS2RobotImporter/Code/Source/Tools/ROS2RobotImporterEditorSystemComponent.cpp
```diff
@@ -235,8 +237,9 @@ namespace ROS2RobotImporter
 
         const AZ::IO::Path prefabPathRelative(AZ::IO::Path("Assets") / "Importer" / prefabName);
         const AZ::IO::Path prefabPath(AZ::IO::Path(AZ::Utils::GetProjectPath()) / prefabPathRelative);
-        AZStd::unique_ptr<SdfPrefabMaker> prefabMaker = AZStd::make_unique<SdfPrefabMaker>(
-            &parsedSdfRoot, prefabPath.String(), AZStd::make_shared<Assets::ReferencedAssetMap>(referencedAssetMap), useArticulation);
+        ImportSession session(std::move(referencedAssetMap));
```
jhanca-robotecai: Although it compiles the same, please use AZStd
```suggestion
        ImportSession session(AZStd::move(referencedAssetMap));
```

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+
+        const Assets::ReferencedAssetMap m_referencedAssetMap;
+        const ImportSessionId m_Id;
+
+        AZStd::mutex m_statusLock;
+        AZStd::multimap<ImportStatusMessageType, AZStd::string> m_status;
+        unsigned int m_articulationsCounter{ 0u };
+    };
+
+} // namespace ROS2RobotImporter
```
jhanca-robotecai: newline

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.cpp
```diff
@@ -0,0 +1,105 @@
...
+    AZStd::string ImportSession::GetStatus()
+    {
+        AZStd::string report;
+        AZStd::lock_guard<AZStd::mutex> lck(m_statusLock);
+
+        // Print warnings first
+        constexpr unsigned int articulationsLimit = 64;
+        if (m_articulationsCounter >= articulationsLimit)
+        {
+            report += "\n##⚠️ Note: the number of articulations (" + AZStd::to_string(m_articulationsCounter) +
```
jhanca-robotecai: The space after `##` is needed to ensure rendering as H2 heading.
```suggestion
            report += "\n## ⚠️ Note: the number of articulations (" + AZStd::to_string(m_articulationsCounter) +
```

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.cpp
```diff
@@ -0,0 +1,105 @@
...
+        auto it = m_referencedAssetMap.find(modelUri);
+        if (it != m_referencedAssetMap.end())
+        {
+            return it->second;
+        }
+
+        return AZStd::nullopt;
+    }
+
+} // namespace ROS2RobotImporter
```
jhanca-robotecai: newline

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+        ~ImportSession();
+
+        ImportSessionId GetId() const;
+
+        //! Get descriptive status of import.
+        //! A string with the status in Markdown format.
+        AZStd::string GetStatus();
+
+    private:
+        void OnImportStatusMessage(ImportStatusMessageType, const AZStd::string&) override;
```
jhanca-robotecai: Add variable names for consistency.

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+#include <AzCore/std/containers/map.h>
+#include <AzCore/std/parallel/mutex.h>
+#include <RobotImporter/RobotImporterAssetsBus.h>
+#include <RobotImporter/RobotImporterStatusBus.h>
+
+namespace ROS2RobotImporter
+{
+    class ImportSession
+        : public RobotImporterAssetsRequestBus::Handler
+        , public RobotImporterStatusRequestBus::Handler
```
jhanca-robotecai: No reason for _public_ inheritance.

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+        ImportSessionId GetId() const;
+
+        //! Get descriptive status of import.
+        //! A string with the status in Markdown format.
+        AZStd::string GetStatus();
+
+    private:
+        void OnImportStatusMessage(ImportStatusMessageType, const AZStd::string&) override;
+        void OnArticulatedLinkCreated() override;
+        AZStd::optional<Assets::ReferencedAsset> FindRefferencedAssets(AZ::IO::Path modelUri) const override;
```
jhanca-robotecai: Please fix the typo and pass the path as _const ref_, so there is no need to copy the full path when looking for every single asset:
```suggestion
        AZStd::optional<Assets::ReferencedAsset> FindReferencedAssets(const AZ::IO::Path& modelUri) const override;
```

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+    {
+    public:
+        explicit ImportSession(Assets::ReferencedAssetMap referencedAssetMap);
+        ~ImportSession();
+
+        ImportSessionId GetId() const;
+
+        //! Get descriptive status of import.
+        //! A string with the status in Markdown format.
+        AZStd::string GetStatus();
```
jhanca-robotecai: Add _const_ for consistency.

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+
+    private:
+        void OnImportStatusMessage(ImportStatusMessageType, const AZStd::string&) override;
+        void OnArticulatedLinkCreated() override;
+        AZStd::optional<Assets::ReferencedAsset> FindRefferencedAssets(AZ::IO::Path modelUri) const override;
+
+        const Assets::ReferencedAssetMap m_referencedAssetMap;
+        const ImportSessionId m_Id;
```
jhanca-robotecai: Use _camelCase_.
```suggestion
        const ImportSessionId m_id;
```

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.cpp
```diff
@@ -0,0 +1,105 @@
...
+    void ImportSession::OnArticulatedLinkCreated()
+    {
+        AZStd::lock_guard<AZStd::mutex> lck(m_statusLock);
+        m_articulationsCounter++;
+    }
+
+    AZStd::optional<Assets::ReferencedAsset> ImportSession::FindRefferencedAssets(AZ::IO::Path modelUri) const
+    {
+        auto it = m_referencedAssetMap.find(modelUri);
+        if (it != m_referencedAssetMap.end())
```
jhanca-robotecai: You might want to limit the scope of the variable
```suggestion
        if (auto it = m_referencedAssetMap.find(modelUri); it != m_referencedAssetMap.end())
```

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+
+        AZStd::mutex m_statusLock;
+        AZStd::multimap<ImportStatusMessageType, AZStd::string> m_status;
+        unsigned int m_articulationsCounter{ 0u };
```
jhanca-robotecai: Please comment why the variables (e.g. `m_articulationsCounter`) are needed

Gems/ROS2RobotImporter/Code/Source/RobotImporter/RobotImporterAssetsBus.h
```diff
@@ -0,0 +1,33 @@
...
+    class RobotImporterAssetsBus : public AZ::EBusTraits
+    {
+    public:
+        static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::ById;
+        using BusIdType = ImportSessionId;
+
+        //! Return referenced asset from RefferencedAssetMap
+        //! @param modelUri URI path for the refferenced asset from the source file
+        virtual AZStd::optional<Assets::ReferencedAsset> FindRefferencedAssets(AZ::IO::Path modelUri) const = 0;
```
jhanca-robotecai: Fix the signature as commented in the handler.

The comment is misleading: _modelUri_ in this call should rather be called a _modelAssetKey_ or something; it is a composition: `modelUri + "/" + assetUri`

Gems/ROS2RobotImporter/Code/Source/RobotImporter/RobotImporterStatusBus.h
```diff
@@ -0,0 +1,45 @@
...
+#pragma once
+
+#include <AzCore/EBus/EBus.h>
+#include <AzCore/EBus/Policies.h>
+#include <AzCore/std/string/string.h>
+#include <RobotImporter/Assets/AssetTypes.h>
+
+namespace ROS2RobotImporter
+{
+    using ImportSessionId = AZ::u32;
```
jhanca-robotecai: This is declared in both buses, you might easily get into troubles if one is accidentally changed while the other is not. Keep in mind, that both are in headers, hence can easily pollute the code.

Gems/ROS2RobotImporter/Code/Source/RobotImporter/RobotImporterStatusBus.h
```diff
@@ -0,0 +1,45 @@
...
+        static constexpr AZ::EBusHandlerPolicy HandlerPolicy = AZ::EBusHandlerPolicy::Single;
+        static constexpr AZ::EBusAddressPolicy AddressPolicy = AZ::EBusAddressPolicy::ById;
+        using BusIdType = ImportSessionId;
+
+        virtual void OnImportStatusMessage(ImportStatusMessageType, const AZStd::string&)
+        {
+        }
+        virtual void OnArticulatedLinkCreated()
+        {
+        }
```
jhanca-robotecai: 1. Please comment the interface.
2. Please rename the methods. The ones starting with `On` suggest this is a _NotificationBus_ by O3DE convention (one emitter, many listeners). It would be more verbose to use names such as _ReportArticulatedLinkCreated_, _IncreaseArticulationsCounter_, or similar.

Gems/ROS2RobotImporter/Code/Source/RobotImporter/RobotImporterAssetsBus.h
```diff
@@ -0,0 +1,33 @@
...
+ * For complete copyright and license terms please see the LICENSE at the root of this distribution.
+ *
+ * SPDX-License-Identifier: Apache-2.0 OR MIT
+ *
+ */
+#pragma once
+
+#include <AzCore/EBus/EBus.h>
+#include <AzCore/EBus/Policies.h>
+#include <AzCore/std/string/string.h>
```
jhanca-robotecai, line 13: You are using `AZ::IO::Path` in the codebase, but no string.
```suggestion
#include <AzCore/IO/Path/Path.h>
```

Gems/ROS2RobotImporter/Code/Source/RobotImporter/RobotImporterStatusBus.h
```diff
@@ -0,0 +1,45 @@
...
+    {
+        Model,
+        Link,
+        Joint,
+        Sensor,
+        SensorPlugin,
+        ModelPlugin
+    };
+
+    class RobotImporterStatusBus : public AZ::EBusTraits
```
jhanca-robotecai: I don't like long names and I do not like the discussions about names, but both buses are important (interfaces of the piece of the system). When connecting to bus I clearly need to know the direction of the information flow and the function of the bus. _AssetsBus_ and _StatusBus_ does not answer the question about the function. Maybe something like _ImporterStatusAggregationBus_ would be more verbose? Maybe _importer_ can be completely skipped due to the namespace and the Gem's name?

Gems/ROS2RobotImporter/Code/Source/RobotImporter/Building/PrefabMakerUtils.cpp
```diff
@@ -114,22 +114,25 @@ namespace ROS2RobotImporter::PrefabMakerUtils
     }
 
     AZStd::optional<Assets::AvailableAsset> GetAssetFromUri(
-        const Assets::ReferencedAssetMap& referencedAssetMap, const AZStd::string& modelUri, const AZStd::string& assetUri)
+        const ImportSessionId sessionId, const AZStd::string& modelUri, const AZStd::string& assetUri)
     {
         const AZStd::string modelAssetUri = (modelUri.empty()) ? assetUri : modelUri + "/" + assetUri;
```
jhanca-robotecai: I guess we can create `AZ::IO::Path` here

Gems/ROS2RobotImporter/Code/Source/RobotImporter/Building/PrefabMakerUtils.h
```diff
@@ -52,16 +53,16 @@ namespace ROS2RobotImporter::PrefabMakerUtils
...
     //! @return Asset for the mesh, if found in the mapping.
     AZStd::optional<Assets::AvailableAsset> GetAssetFromUri(
-        const Assets::ReferencedAssetMap& sdfAssetsMapping, const AZStd::string& modelUri, const AZStd::string& assetUri);
+        const ImportSessionId sessionId, const AZStd::string& modelUri, const AZStd::string& assetUri);
 
     //! Get Asset from path. Version for std::string.
     //! @see GetAssetFromUri.
     AZStd::optional<Assets::AvailableAsset> GetAssetFromUri(
-        const Assets::ReferencedAssetMap& sdfAssetsMapping, const AZStd::string& modelUri, const std::string& assetUri);
+        const ImportSessionId sessionId, const AZStd::string& modelUri, const std::string& assetUri);
```
jhanca-robotecai: Can we please refactor this part as well? I see no point in helper methods calling bus; the bus can be called directly in _CollidersMaker_ and _VisualsMaker_. Thanks!

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
```diff
@@ -0,0 +1,45 @@
...
+#include <RobotImporter/RobotImporterStatusBus.h>
+
+namespace ROS2RobotImporter
+{
+    class ImportSession
+        : public RobotImporterAssetsRequestBus::Handler
+        , public RobotImporterStatusRequestBus::Handler
+    {
+    public:
+        explicit ImportSession(Assets::ReferencedAssetMap referencedAssetMap);
```
jhanca-robotecai, line 23: You might want to disable copy/move:

```suggestion
        explicit ImportSession(Assets::ReferencedAssetMap referencedAssetMap);
        CollidersMaker(const CollidersMaker& other) = delete;
```

Gems/ROS2RobotImporter/Code/Source/Tools/ImportSession.h
jhanca-robotecai: No reason to make it _public_.
```suggestion
    class ImportSession
        : private RobotImporterAssetsRequestBus::Handler
        , private RobotImporterStatusRequestBus::Handler
```

jhanca-robotecai (commented):
The code works as expected and it is a step forward. I am not sure if the architecture is complete, though.

In particular, I am not sure if `ImportSession` is needed at all. Right now it handles two buses to solve two problems:
1) It aggregates status messages
2) It manages assets

We have three different processing paths: (i) wizard, (ii) handler in system component for scripts, (iii) asset builder. Integrating status messages is used only in the wizard. The two remaining paths are producing messages, but never printing those. We cannot run multiple wizards at once, hence we do not need different handlers per session. As for managing assets: we currently key the assets with modelUri+assetUri. If we integrate source urdf/sdf/xarcro into the key, we won't get any collisions. This way we could completely remove SessionImporter, hence removing sessionId from all makers. This way we wouldn't need to create _makers_ objects for each asset in AssetBuilder, because of _makers_ being simply static utils. On the contrary, the current implementation (RAII) let's us clean the memory after importing each robot.

For now we can follow the path suggested in this PR, but I would keep that in mind. Please clean the nitpicks listed in the review though.

jhanca-robotecai (approved):
Thank you for incorporating all fixes!

PR #1078, opened 2026-08-13
Fix ResetSimulationService handler: bits not ints

### LINKED ISSUE #1077: [Simulation Interfaces] Reset Scope should be a bitfiled
**Describe the bug**
The `/reset_simulation` service should interpret the given `scope` as a bitfield (see https://github.com/ros-simulation/simulation_interfaces/blob/main/srv/ResetSimulation.srv). The current implementation in
[Gems/SimulationInterfaces/Code/Source/Services/ResetSimulationServiceHandler.cpp](https://github.com/o3de/o3de-extras/blob/development/Gems/SimulationInterfaces/Code/Source/Services/ResetSimulationServiceHandler.cpp#L36-L133)
directly compares the int values and does not mask the individual bits.

**Steps to reproduce**
1. `ros2 service call /reset_simulation simulation_interfaces/srv/ResetSimulation 'scope: 6'`

**Expected behavior**
The set scopes should be cleared.

**Actual behavior**
The call returns an error:
`simulation_interfaces.srv.ResetSimulation_Response(result=simulation_interfaces.msg.Result(result=2, error_message='Passed unknown scope'))`

The reset of the individual scopes 2 and 4 work.

**Found in Branch**
Version: 2605.0

jhanca-robotecai:
Close #1077

## What does this PR do?

Following the issue, the input scope is used as a bitmask. This required to change the order of execution of different scopes.

Tests were added.

## How was this PR tested?

Built the code, spawned object, moved object
- checked reset `SCOPE_STATE` in different bit combinations (2, 4, 5, 6)
- checked reset `SCOPE_SPAWNED`  in different bit combinations (2, 4, 5, 6)

### DIFF
--- a/Gems/SimulationInterfaces/Code/Source/Services/ResetSimulationServiceHandler.cpp
+++ b/Gems/SimulationInterfaces/Code/Source/Services/ResetSimulationServiceHandler.cpp
@@ -33,72 +33,8 @@ namespace ROS2SimulationInterfaces
     AZStd::optional<ResetSimulationServiceHandler::Response> ResetSimulationServiceHandler::HandleServiceRequest(
         const std::shared_ptr<rmw_request_id_t> header, const Request& request)
     {
-        if (request.scope == Request::SCOPE_STATE)
-        {
-            Response response;
-            AZ::Outcome<void, SimulationInterfaces::FailedResult> resetingOutcome;
-            SimulationInterfaces::SimulationEntityManagerRequestBus::BroadcastResult(
-                resetingOutcome, &SimulationInterfaces::SimulationEntityManagerRequests::ResetAllEntitiesToInitialState);
-            if (resetingOutcome.IsSuccess())
-            {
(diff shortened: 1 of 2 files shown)


norbertprokopiuk (commented):
Generally works as expected, scopes can be now combined. I tested state + time and all 3 combined. Tests passes as well. 

But I noticed that scope =0 doesn't restart time. It must be some sort of regression/change in the engine itself, which earlier forced me to add this PR #1074. I had to explicitly remove all spawned entities before reload of level. 

But now this issue with time, suggests that system entity and its component isn't reactivated on level reload. We need to either add explicit time reset on level reload or investigate why system entity stopped reactivating. I'm okay with both approaches, but I would do it in this PR to fully fix reset simulation service

norbertprokopiuk (approved):
I accidentally have added setreg disabling clock reset on level relaod. After removing it it works just ifne

PR #1085, opened 2026-08-14
Refactor of RobotImporter asset detection and copying - deduplication across diffrent import method.

Sebastian306:
## What does this PR do?

Extraction of duplicated code parts between different import processes. 
Extraction of a duplicated function for finding an asset via the Asset Processor for a resolved path. Added a check in the wizard's asset search for whether the source asset file is not already in the project, before searching through all files to find one with the same content. 
Extraction of duplicated logic for copying assets via the widget and EBus.

PR requires prior merger of changes from PR [RobotImporter Gem - passing referenced assets and import status through EBuses.](https://github.com/o3de/o3de-extras/pull/1075)

## How was this PR tested?

A sample project featuring a new robot model. An xacro file for the "mini_pupper" robot has been loaded using Python code and the wizard.

### DIFF
--- a/Gems/ROS2RobotImporter/Code/Source/RobotImporter/Assets/AssetImporter.cpp
+++ b/Gems/ROS2RobotImporter/Code/Source/RobotImporter/Assets/AssetImporter.cpp
@@ -16,29 +16,29 @@
 
 namespace ROS2RobotImporter::Assets
 {
-    void CopyReferencedAssetsAndCreateAssetMap(
+    bool CopyReferencedAssets(
         ReferencedAssetMap& referencedAssetMap,
         const AZ::IO::Path& sourceFilePath,
-        const SdfAssetBuilderSettings& sdfBuilderSettings,
+        CopyAssetsStatusCallback statusCallback,
         AZStd::string_view outputDirSuffix,
         AZ::IO::FileIOBase* fileIO)
     {
-        ResolveAssetMap(referencedAssetMap, sourceFilePath, sdfBuilderSettings);
-        AZStd::mutex referencedAssetMapMutex;
         if (referencedAssetMap.empty())
         {
-            return;
+            return false;
         }
 
         auto destDirectory = PrepareImportedAssetsDest(sourceFilePath, outputDirSuffix, fileIO);
         if (!destDirectory.IsSuccess())
         {
-            return;
+            AZ_Error("CopyReferencedAsset", false, "Failed to create destination folder for imported assets");
+            return false;
         }
 
         AZStd::unordered_map<AZ::IO::Path, unsigned int> duplicatedFilenames;
         for (auto& [unresolvedFileName, referencedAsset] : referencedAssetMap)
         {
+            auto copyStatus = CopyStatus::Copying;
             if (duplicatedFilenames.contains(referencedAsset.m_assetUri))
             {
                 duplicatedFilenames[referencedAsset.m_assetUri]++;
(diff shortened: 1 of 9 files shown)


Gems/ROS2RobotImporter/Code/Source/RobotImporter/Assets/AssetImporter.cpp
```diff
@@ -16,29 +16,29 @@
...
+        CopyAssetsStatusCallback statusCallback,
         AZStd::string_view outputDirSuffix,
         AZ::IO::FileIOBase* fileIO)
     {
-        ResolveAssetMap(referencedAssetMap, sourceFilePath, sdfBuilderSettings);
-        AZStd::mutex referencedAssetMapMutex;
         if (referencedAssetMap.empty())
         {
-            return;
+            return false;
```
jhanca-robotecai: This way `CopyReferencedAssets()` returns false both on real failure and when there is nothing to copy. This looks incorrect to me.

Gems/ROS2RobotImporter/Code/Source/Tools/ROS2RobotImporterEditorSystemComponent.cpp
```diff
@@ -154,7 +155,12 @@ namespace ROS2RobotImporter
         auto referencedAssetMap = Assets::GetReferencedAssetFilenames(parsedSdfRoot);
         if (importAssetWithFile)
         {
-            Assets::CopyReferencedAssetsAndCreateAssetMap(referencedAssetMap, filePath, sdfBuilderSettings);
+            Assets::ResolveAssetMap(referencedAssetMap, filePath, sdfBuilderSettings);
+            bool copyResultOk = Assets::CopyReferencedAssets(referencedAssetMap, filePath);
+            if (!copyResultOk)
+            {
+                AZ_Warning("ROS2RobotImporterEditorSystemComponent", false, "Copying assets failed");
+            }
```
jhanca-robotecai: `AZ_Warning` holds a condition.
```suggestion
            AZ_Warning("ROS2RobotImporterEditorSystemComponent", copyResultOk, "Copying assets failed");
```

Gems/ROS2RobotImporter/Code/Source/RobotImporter/Assets/SceneManifestBuilder.cpp
jhanca-robotecai: This include is not needed anymore.

Gems/ROS2RobotImporter/Code/Source/RobotImporter/Assets/AssetLookup.h
```diff
@@ -31,15 +31,13 @@ namespace ROS2RobotImporter::Assets
...
     //! Steps:
-    //! - Function scans all available O3DE assets by calling `GetInterestingSourceAssetsCRC`.
-    //! - Files pointed by resolved paths have their checksum computed `GetFileCRC`.
-    //! - Suitable mapping to the O3DE asset is found by comparing the checksum of the file pointed by the resolved path and source asset.
-    //! @param unresolvedAssetMap - list of the unresolved paths from the SDF/URDF file that are to be found as assets
-    //! @param sourceFilePath - path of the SDF/URDF file, used for resolving paths of referenced assets
-    //! @param sdfBuilderSettings - the builder settings that should be used to resolve paths
-    void FindReferencedAssets(
-        ReferencedAssetMap& unresolvedAssetMap, const AZ::IO::Path& sourceFilePath, const SdfAssetBuilderSettings& sdfBuilderSettings);
+    //! - Assets resolved to a path inside an Asset Processor scan folder are looked up by that path `LookupSourceAssetByPath`.
```
jhanca-robotecai:
```suggestion
    //! - Assets resolved to a path inside an Asset Processor scan folder are looked up by that path `GetAvailableAssetInfo`.
```

Gems/ROS2RobotImporter/Code/Source/SdfAssetBuilder/SdfAssetBuilder.cpp
```diff
@@ -113,51 +107,22 @@ namespace ROS2RobotImporter
...
-            bool sourceAssetFound{ false };
-            AZ::Data::AssetInfo assetInfo;
-            AZStd::string watchFolder;
-            AssetSysReqBus::BroadcastResult(
-                sourceAssetFound, &AssetSysReqBus::Events::GetSourceInfoBySourcePath, asset.m_resolvedPath.c_str(), assetInfo, watchFolder);
-
-            if (!sourceAssetFound)
+            // If the source asset has been found by the Asset Processor, save the mapping between raw uri
+            // reference and Asset Processor source asset information.
+            auto assetInfo = Assets::GetAvailableAssetInfo(asset.m_resolvedPath);
```
jhanca-robotecai, line 112: The old code compared CRC of two files: the one the URI pointed to, and the one the Asset Processor matched by path. The outcome was accepted if the two hashes were equal.
`GetAvailableAssetInfo` doesn't do it. Could you please explain why this change is valid?

Sebastian306: That is indeed an oversight. Although I thought that `watchFolder` + `relativePath` should always point to the same file as `globalSourceAssetPath`, a closer look reveals exceptions to this rule.
For example, consider two files whose names differ only in capitalization.
The CRC comparison has been restored.

jhanca-robotecai (approved):

PR #1089, opened 2026-08-28
Add workflow to test simulation Gems for PR

jhanca-robotecai:
## What does this PR do?

This PR adds a GitHub Actions workflow (`simulation-gems-pr.yml`) that builds and tests the following Gems:
- LevelGeoreferencing
- ROS2
- ROS2Controllers
- ROS2RobotImporter
- ROS2Sensors
- SimulationInterfaces

The Action is triggered on PRs touching those gems and via manual dispatch. It resolves the engine's own toolchain/runner image from its `ar.yml` and checks out the engine at the matching branch (`development` / `stabilization/*`) so tests always run against the right engine version.

The Action uses sparse check-out to keep the clone fast and builds only necessary test targets.

The downside: the engine still uses Ubuntu 2204 for tests, which forces o3de-extras to be tested against ROS 2 Humble.
A separate Github Actions workflow running against Jazzy and Lyrical, triggered on `development` and `stabilization/*` pushes (after PR merge) will follow in a separate PR.

## How was this PR tested?

The tests were run in my fork: https://github.com/jhanca-robotecai/o3de-extras/actions/runs/33093070251/job/98591015759
They failed, which is a valid reason to implement tests for the branch.

### DIFF
--- a/.github/workflows/simulation-gems-pr.yml
+++ b/.github/workflows/simulation-gems-pr.yml
@@ -24,6 +24,9 @@ concurrency:
   group: simulation-gems-pr-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
 env:
   GIT_LFS_SKIP_SMUDGE: "1"
 
@@ -42,7 +45,7 @@ jobs:
     steps:
       # base_ref names the matching engine branch; empty on manual workflow_dispatch runs.
       - name: Checkout engine ar.yml (${{ github.base_ref || inputs.engine_ref }})
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           repository: o3de/o3de
           ref: ${{ github.base_ref || inputs.engine_ref }}
@@ -69,10 +72,34 @@ jobs:
               "ubuntu-26.04": "lyrical",
           }
 
-          with open(".github/workflows/ar.yml") as f:
-              job = yaml.safe_load(f)["jobs"]["Linux-Profile"]["with"]
+          try:
+              with open(".github/workflows/ar.yml") as f:
+                  ar_yml = yaml.safe_load(f)
+              job = ar_yml["jobs"]["Linux-Profile"]["with"]
+          except FileNotFoundError:
+              sys.exit(
+                  "::error::engine/.github/workflows/ar.yml not found. "
+                  "Unable to resolve the engine config."
+              )
+          except KeyError:
+              sys.exit(
+                  "::error::jobs.Linux-Profile.with not found in ar.yml. "
+                  "The engine's ar.yml layout may have changed - update the "
+                  "parsing logic in this workflow to match."
+              )
 
-          image = job["image"]
+          try:
+              image = job["image"]
+              clang_version = job["clang_version"]
+              gcc_version = job["gcc_version"]
+              cmake_version = job["cmake_version"]
+          except KeyError as e:
+              sys.exit(
+                  f"::error::jobs.Linux-Profile.with.{e.args[0]} not found in ar.yml. "
+                  "The engine's ar.yml layout may have changed - update the "
+                  "parsing logic in this workflow to match."
+              )
+
           ros_distro = ROS_DISTRO_BY_IMAGE.get(image)
           if ros_distro is None:
               sys.exit(
@@ -81,9 +108,9 @@ jobs:
               )
 
           print(f"image={image}")
-          print(f"clang_version={job['clang_version']}")
-          print(f"gcc_version={job['gcc_version']}")
-          print(f"cmake_version={job['cmake_version']}")
+          print(f"clang_version={clang_version}")
+          print(f"gcc_version={gcc_version}")
+          print(f"cmake_version={cmake_version}")
           print(f"ros_distro={ros_distro}")
           PY
 
@@ -98,7 +125,7 @@ jobs:
       ROS_DISTRO: ${{ needs.resolve-engine-config.outputs.ros_distro }}
     steps:
       - name: Checkout o3de-extras (PR head)
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           path: o3de-extras
 
@@ -106,7 +133,7 @@ jobs:
       # Excludes the largest unrelated gems and most of AutomatedTesting
       # (clones only about 40% of a full engine's size without LFS content).
       - name: Checkout o3de engine (${{ needs.resolve-engine-config.outputs.engine_sha }})
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           repository: o3de/o3de
           ref: ${{ needs.resolve-engine-config.outputs.engine_sha }}
@@ -151,7 +178,7 @@ jobs:
 
       # Keyed on engine commit + toolchain + image/ROS so a bump busts the cache.
       - name: Cache ccache and 3rdParty packages
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: |
             ~/.ccache
@@ -161,7 +188,7 @@ jobs:
             simulation-gems-${{ needs.resolve-engine-config.outputs.image }}-${{ env.ROS_DISTRO }}-
 
       - name: Install CMake matching the engine's ar.yml
-        uses: lukka/get-cmake@latest
+        uses: lukka/get-cmake@fffaaafeea488556c2c12dad60690008bc1caacb # v4.4.2
         with:
           cmakeVersion: ${{ needs.resolve-engine-config.outputs.cmake_version }}
 
@@ -249,6 +276,7 @@ jobs:
       - name: Build gem modules, test targets and project assets
         timeout-minutes: 330
         run: |
+          source /opt/ros/${{ env.ROS_DISTRO }}/setup.bash
           cmake --build "${{ env.PROJECT_PATH }}/build/linux" --config profile --target \
             LevelGeoreferencing LevelGeoreferencing.Editor \
             ROS2 ROS2.Editor \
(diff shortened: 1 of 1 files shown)


.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+          import yaml
+
+          ROS_DISTRO_BY_IMAGE = {
+              "ubuntu-22.04": "humble",
+              "ubuntu-24.04": "jazzy",
+              "ubuntu-26.04": "lyrical",
+          }
+
+          with open(".github/workflows/ar.yml") as f:
+              job = yaml.safe_load(f)["jobs"]["Linux-Profile"]["with"]
```
arturkamieniecki: Consider adding a better error message for the workflow the ar.yml file is not found or the safe_load fails. As of now these messages will be generic python errors, that could not give meaningful insight.

Something like ar.yml file not, found. Unable to resolve the engine config. Or Linux build profile not found. Check the ar.yml file if it exists.

This will make sure that when these files are moved, changed it is clear why the job fails.

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+
+          image = job["image"]
```
arturkamieniecki: Similarly to the previous comment, verify if this value exists as the ar.yml file is not part of this repo and can change anytime.

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+
+      - name: Run gem tests
+        run: |
+          source /opt/ros/${{ env.ROS_DISTRO }}/setup.bash
+          Xvfb :0 -screen 0 1400x1050x16 &
+          export DISPLAY=:0
+          ctest --test-dir "${{ env.PROJECT_PATH }}/build/linux" -C profile \
+            --output-on-failure \
+            --output-junit Testing/test-results.xml \
+            --tests-regex "(LevelGeoreferencing\.Tests|LevelGeoreferencing\.Editor\.Tests|ROS2\.Editor\.Tests|ROS2Controllers\.Tests|ROS2Controllers\.Editor\.Tests|ROS2RobotImporter\.Editor\.Tests|ROS2Sensors\.Tests|SimulationInterfaces\.Tests|SimulationInterfaces\.Editor\.Tests|SimulationInterfaces\.ROS2Tests)"
```
arturkamieniecki: Maybe using `xvfb-run` could be better. Something like ```xvfb-run --auto-servernum --server-args="-screen 0 1400x1050x16" \
  ctest --test-dir "${{ env.PROJECT_PATH }}/build/linux" -C profile \
    --output-on-failure \
    --output-junit Testing/test-results.xml \
    --tests-regex "...```

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+      engine_sha: ${{ steps.parse.outputs.sha }}
+      image: ${{ steps.parse.outputs.image }}
+      clang_version: ${{ steps.parse.outputs.clang_version }}
+      gcc_version: ${{ steps.parse.outputs.gcc_version }}
+      cmake_version: ${{ steps.parse.outputs.cmake_version }}
+      ros_distro: ${{ steps.parse.outputs.ros_distro }}
+    steps:
+      # base_ref names the matching engine branch; empty on manual workflow_dispatch runs.
+      - name: Checkout engine ar.yml (${{ github.base_ref || inputs.engine_ref }})
+        uses: actions/checkout@v4
```
mateuszwasilewski: Bump to v7, Node 20 used by v4 is deprecated

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+env:
+  GIT_LFS_SKIP_SMUDGE: "1"
+
+jobs:
+  # runs-on can't read a step output in its own job, so resolve it here first.
+  resolve-engine-config:
+    name: Resolve engine image, toolchain and ROS 2 distro
+    runs-on: ubuntu-latest
+    outputs:
+      engine_sha: ${{ steps.parse.outputs.sha }}
```
mateuszwasilewski: If we add `id: checkout` to checkout step
Then we can simply use its output
```suggestion
      engine_sha: ${{ steps.checkout.outputs.commit }}
```
https://github.com/actions/checkout/blob/3d3c42e5aac5ba805825da76410c181273ba90b1/action.yml#L114

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+            "Gems/ScriptCanvasTesting", "Gems/Vegetation"
+          ]' engine.json > /tmp/engine.json
+          mv /tmp/engine.json engine.json
+          jq '.external_subdirectories -= ["TechnicalArt/DccScriptingInterface"]' \
+            Gems/AtomLyIntegration/gem.json > /tmp/atomly.json
+          mv /tmp/atomly.json Gems/AtomLyIntegration/gem.json
+
+      # Keyed on engine commit + toolchain + image/ROS so a bump busts the cache.
+      - name: Cache ccache and 3rdParty packages
+        uses: actions/cache@v4
```
mateuszwasilewski:
```suggestion
        uses: actions/cache@v6
```
v4 uses Node 20 which is deprecated.

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+        with:
+          path: |
+            ~/.ccache
+            ~/.o3de/3rdParty
+          key: simulation-gems-${{ needs.resolve-engine-config.outputs.image }}-${{ env.ROS_DISTRO }}-${{ needs.resolve-engine-config.outputs.engine_sha }}-clang${{ needs.resolve-engine-config.outputs.clang_version }}
+          restore-keys: |
+            simulation-gems-${{ needs.resolve-engine-config.outputs.image }}-${{ env.ROS_DISTRO }}-
```
mateuszwasilewski: I think clang version should be used as a part of restore-keys.
When clang version is changes all of ccache will be irrelevant

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+
+      # Pinned to the SHA resolve-engine-config already resolved.
+      # Excludes the largest unrelated gems and most of AutomatedTesting
+      # (clones only about 40% of a full engine's size without LFS content).
+      - name: Checkout o3de engine (${{ needs.resolve-engine-config.outputs.engine_sha }})
+        uses: actions/checkout@v4
+        with:
+          repository: o3de/o3de
+          ref: ${{ needs.resolve-engine-config.outputs.engine_sha }}
+          path: engine
```
mateuszwasilewski:
```suggestion
          path: ${{ env.ENGINE_PATH }}
```
To be consistent with usage in the rest of the job.

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+          jq '.external_subdirectories -= ["TechnicalArt/DccScriptingInterface"]' \
+            Gems/AtomLyIntegration/gem.json > /tmp/atomly.json
+          mv /tmp/atomly.json Gems/AtomLyIntegration/gem.json
+
+      # Keyed on engine commit + toolchain + image/ROS so a bump busts the cache.
+      - name: Cache ccache and 3rdParty packages
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.ccache
```
mateuszwasilewski: This job doesn't setup ccache at all.
It should be added similarly to o3de
https://github.com/o3de/o3de/blob/6b6f2bf95fabeb4cf3cf378bb344219bfde09b0b/.github/workflows/linux-build.yml#L102

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,284 @@
...
+          Xvfb :0 -screen 0 1400x1050x16 &
+          export DISPLAY=:0
+          ctest --test-dir "${{ env.PROJECT_PATH }}/build/linux" -C profile \
+            --output-on-failure \
+            --output-junit Testing/test-results.xml \
+            --tests-regex "(LevelGeoreferencing\.Tests|LevelGeoreferencing\.Editor\.Tests|ROS2\.Editor\.Tests|ROS2Controllers\.Tests|ROS2Controllers\.Editor\.Tests|ROS2RobotImporter\.Editor\.Tests|ROS2Sensors\.Tests|SimulationInterfaces\.Tests|SimulationInterfaces\.Editor\.Tests|SimulationInterfaces\.ROS2Tests)"
+
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v4
```
mateuszwasilewski: Update to v7

jhanca-robotecai:
The tests passed: besides fixing all issues listed by the reviewers, I also fixed tests - missing ROS 2 in env: https://github.com/jhanca-robotecai/o3de-extras/actions/runs/33401555304/job/99518889613

.github/workflows/simulation-gems-pr.yml
```diff
@@ -0,0 +1,311 @@
...
+        with:
+          cmakeVersion: ${{ needs.resolve-engine-config.outputs.cmake_version }}
+
+      - name: Install clang / gcc matching the engine's ar.yml
+        run: |
+          sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+          sudo apt-get update
+          sudo apt-get install -y gcc-${{ needs.resolve-engine-config.outputs.gcc_version }} g++-${{ needs.resolve-engine-config.outputs.gcc_version }}
+          curl -fsSL https://apt.llvm.org/llvm.sh -o /tmp/llvm.sh
+          sudo bash /tmp/llvm.sh ${{ needs.resolve-engine-config.outputs.clang_version }}
```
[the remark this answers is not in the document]
jhanca-robotecai: This workflow is deliberately mirroring the engine's established build convention, not introducing a new risk pattern. apt.llvm.org's llvm.sh doesn't publish per-version checksums to pin against.

nick-l-o3de (approved):
It really looks like a good start to me.

arturkamieniecki (approved):

PR #1067, opened 2026-07-28
Introduce compression publication to ROS 2 camera

michalpelka:
## What does this PR do?

Introduce compression to ROS 2 camera.

- modified UX to ROS2ImageEncodingConversionComponent - it now allows to choose conversion
- added to ROS2ImageEncodingConversionComponent depth32F1 -> mono16/mono8
- moved conversion code out of component for reuse in other components
- added runtime fallback to ROS2ImageCompressionComponent (depth32F1 to mono16)- applied when there is no ROS2ImageEncodingConversionComponent with warning in the log.
- added automatically generated tests.

## New UX of ROS2ImageEncodingConversionComponent
In development was bug that prevented user getting back from invalid config. I've changed it to message, that user is about shoot themself in foot:
Screenshot from 2026-07-30 11-33-07

## ROS2ImageCompressionEditorComponent
image

I also do not prevent user from trying to compress depth with JPEG (what makes no sense in any application).

## How was this PR tested?
image

- added tests
- build and tested against o3de development.

PR #1095, opened 2026-09-03
Development -> Stabilization cherries: part 1

jhanca-robotecai:
## What does this PR do?

This is a set of features that were added to the `development` branch after the _stabilization_ period started, but before the _feature-lock_. The features will be added value to the release and should be integrated:
- #1067: Camera compression feature is an additional component, it does not break the current pipelines. No risk in adding a new, standalone, component.
- #1066: The new linking technique was stale in review for long time due to extensive testing, it is the most risky cherry from the group, but it is very beneficial for users who release ROS 2 O3DE projects. It will be tested again during the stabilization period.
- #1090: Bump of the 3rdParty, low risk - the previous release works perfectly fine with 2610, we will ask users to downgrade in the worst case scenario. 
- #1089: Tests running on Github, no influence at end users, no risk.

## How was this PR tested?

The code builds, the tests run. The code was reviewed in the respective PRs.


michalpelka (approved):

nickschuetz:
Approved on behalf of @o3de/sig-release

